PR #1, opened 2020-08-05
Validate MSVC Windows build 

AndreaLanfranchi:


CMakeLists.txt
```diff
@@ -53,7 +53,10 @@ target_include_directories(secp256k1 PRIVATE secp256k1 INTERFACE secp256k1/inclu
 
 # Compiler options
 if(MSVC)
-  add_compile_options(/W4 /WX)
```
yperbasis, line 56: Shall we keep /WX?

CMakeLists.txt
```diff
@@ -87,7 +90,13 @@ target_link_libraries(evmone PUBLIC evmc intx::intx PRIVATE ethash::keccak)
 add_subdirectory(silkworm)
 
 add_executable(check_changes cmd/check_changes.cpp)
-target_link_libraries(check_changes PRIVATE silkworm absl::flags_parse absl::time)
+
+if(MSVC)
+  # lmdb requires link to ntdll.lib
```
yperbasis, line 95: Let's link silkworm to ntdll rather than check_changes and tests individually

AndreaLanfranchi: Wilco

CMakeLists.txt
AndreaLanfranchi: Yes we can but have to resolve several warnings

yperbasis: I see

PR #75, opened 2020-10-28
default logging to cerr, all the rest is optional

gcolvin:
Usage:
```
#include <silkworm/common/log.hpp>
...
// log to default logger
SILKWORM_LOG(LogInfo) << "default logger" << endl;
...
// create local logger and log to it
silkworm::Logger logger(LogInfo, cerr, file_ostream);
...
SILKWORM_LOG_TO(logger, LogCritical) << "local logger" << endl;
...
}
```
Also, to reset the logging level:
```
SILKWORM_LOG_LEVEL(level);
SILKWORM_LOG_TO_LEVEL(logger, level);
```


yperbasis:
Please reformat the files with clang-format.

silkworm/common/log.hpp
```diff
@@ -27,31 +28,40 @@ enum LogLevels {
...
-                  << std::put_time(std::gmtime(&time), " %Y-%m-%d %H:%M:%S %Z ") \
-                  << file << ":" << line << "| ";
+    std::ostream& log(LogLevels level, const char* file, int line) {
+        std::time_t time = std::time(nullptr);
+        return stream << LogTags[level]
+                      << std::put_time(std::gmtime(&time), " %Y-%m-%d %H:%M:%S %Z ")
+                      << file << ":" << line << "| ";
     }
+
+    static Logger& default_logger();
```
yperbasis, line 47: Where is default_logger() defined? You apparently forgot to add log.cpp into the review.

silkworm/common/log.hpp
```diff
@@ -27,31 +28,40 @@ enum LogLevels {
     LogTrace, LogDebug, LogInfo, LogWarn, LogError, LogCrit, LogNone
```
yperbasis, line 28: What does LogNone mean?

silkworm/common/log.hpp
```diff
@@ -27,31 +28,40 @@ enum LogLevels {
...
+        "TRACE", "DEBUG", "INFO ", "WARN ", "ERROR", "CRIT ", "NONE"
+    };
 };
 
-// log to whatever logger is in scope
-#define SILKWORM_LOG(level_) \
-    if (logger.level() < (level_)) {} \
-    else logger.log(level_, __FILE__, __LINE__)
+#define SILKWORM_LOG_TO(logger_, level_) \
+    if ((logger_).level() < (level_)) {} \
```
yperbasis, line 59: This check should be the other way round: logger with ERROR verbosity should show ERROR and CRIT messages and ignore TRACE, DEBUG, INFO, WARN.

silkworm/common/log_test.cpp
```diff
@@ -54,30 +54,27 @@ namespace {
...
-    SILKWORM_LOG(LogDebug) << "LogDebug" << std::endl;
+    SILKWORM_LOG_TO_LEVEL(logger, LogTrace);
+    SILKWORM_LOG_TO(logger, LogDebug) << "LogDebug" << std::endl;
     if (!test_log("", "", "")) return false;
 
     return true;
 }
 
-TEST_CASE("Logging") {
-    CHECK(test_logging() == true);
```
yperbasis, line 80: Why did you remove the CHECK? Unit tests should assert, otherwise they are pointless.

yperbasis (changes requested):

silkworm/common/log.hpp
```diff
@@ -27,31 +28,40 @@ enum LogLevels {
...
+    else (logger_).log((level_), __FILE__, __LINE__)
+#define SILKWORM_LOG_TO_LEVEL(logger_, level_) ((logger_).level(level_))
```
yperbasis, line 61: There's no need for it to be a macro, is there? We should [avoid macros if possible](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros).

gcolvin: Yeah - turns out 'commit -a' doesn't do it

It needs to be a macro to avoid evaluating the `log(level,file,line)` call and the subsequent `<<` ops when the logging level is too low.

silkworm/common/log.hpp
gcolvin: Just that.  Useful for creating a logger that is silent until turned on, or made silent later.

silkworm/common/log.hpp
yperbasis: OK, but SILKWORM_LOG_TO_LEVEL doesn't have to be a macro, does it?

silkworm/common/log.hpp
yperbasis: Hmm. It might be confusing to the caller. I'd expect `SILKWORM_LOG(LogNone) << "hello" << endl;` not to print anything, but it's actually above LogCrit.

silkworm/common/log_test.cpp
gcolvin: Accident in testing. Cleaned it all up.

silkworm/common/log.hpp
gcolvin: Easy enough, but not my intuition.  It seems we want to see critical messages for any logging level but LogNone.  Your way means that the more critical the logging level the less verbose the logging, and the less information as to why the critical condition occurred.  Different philosophies I guess.

yperbasis: The convention I'm talking about is pretty much the de-facto standard. Boost.Log does the same: https://www.boost.org/doc/libs/1_74_0/libs/log/doc/html/log/tutorial/trivial_filtering.html and https://github.com/boostorg/log/blob/develop/include/boost/log/trivial.hpp#L42

silkworm/common/log.hpp
gcolvin: Yes. If we order verbosity in your direction it goes at the other end.

silkworm/common/log.hpp
gcolvin: OK.  Opposite of what I'm used to, but I'll change.

silkworm/common/log.hpp
gcolvin: I find it much more clear to have four consistently named logging macros than two macros and two raw C++ calls.  It also makes using and changing the code behind the macros easier -- who wants to remember that the default logger is currently accessed by reference via a static member function of a class rather than one of the other ways it might be?

silkworm/common/log.hpp
yperbasis: I'd remove LogNone altogether to avoid potential confusion.

gcolvin: LogNone is the only way to turn off logging.

yperbasis (approved):

Issue #78, opened 2020-10-29
Transaction analytics by transaction replay


AlexeyAkhunov:
This in an introductory project designed to find some very useful information and perhaps optimisation opportunities for Silkworm execution, at the same time requiring understanding of the code and some aspect of Turbo-Geth/Silkworm data model.
Carrying out this task will require the database obtained by fully syncing a Turbo-Geth node, which should be compatible with the silkworm code. This is because the task requires replaying all transactions from the mainnet history. Of course, for most debugging and testing, only a small subset of transaction should be replayed.

# Goals

Here are 3 high-level goals:
1. Find number of transactions that fail and consume the entire allotted gas. These are "Out of Gas" errors, hitting unrecognised instruction, making invalid jump, or explicitly calling `INVALID` instruction. These transactions do not make any changes in Ethereum state, except for deducting the ETH (`gasprice * tx gaslimit`) from the sender's balance of the transaction, and adding the same amount to the miner's balance.
2. Think about data structure (for example, compressed bitmap) that would allow us to mark the transactions found in `1.` so that we can use this data structure as one of the inputs to replay
3. Using the data structure from `2.`, see if there is a benefit in skipping the execution of these transactions entirely, and how much benefits that could bring.
4. Also using data structure from `2.`, see if we can simplify the configuration of EVM, by retrofitting some opcodes that did not exist from the beginning, and "pretend" that they existed from the very beginning of Ethereum in 2015. For example, bit shifting opcodes, if used prior to Byzantium hard fork where they were introduced, would produce failure of the type described in `1.`, and if we skip these transactions, we can simply pretend that the bit shifting opcodes were always around. Find all such opcodes and see how this can simplify the EVM configuration.

# Tools

Apart from having the synced database, you will need the code that is capable of re-executing any historical transactions at will. Such code exists in Silkworm, in the `silkworm/cmd/check_changes.cpp` file. This command line utility can be used as a starting point, it accept parameters that specify the database, as well as range of historical block numbers for which to re-play transactions.

[...]

PR #80, opened 2020-11-02
078 tx analysis

tjayrush:
Implements code required to do initial analysis for #78

### DIFF
--- a/cmd/scan_txs.cpp
+++ b/cmd/scan_txs.cpp
@@ -1,0 +1,136 @@
+/*
+   Copyright 2020 The Silkworm Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#include <absl/container/flat_hash_set.h>
+#include <absl/flags/flag.h>
+#include <absl/flags/parse.h>
+#include <absl/flags/usage.h>
+#include <absl/time/time.h>
+
+#include <boost/filesystem.hpp>
+#include <iostream>
+#include <silkworm/db/access_layer.hpp>
+#include <silkworm/execution/execution.hpp>
+
+using namespace evmc::literals;
+
+// Non-existing accounts only touched by zero-value internal transactions:
+// e.g. https://etherscan.io/address/0x000000000000000000636f6e736f6c652e6c6f67
+static const absl::flat_hash_set<evmc::address> kPhantomAccounts{
+    0x000000000000000000636f6e736f6c652e6c6f67_address,
+    0x2386f26fc10000b4e16d0168e52d35cacd2c6185_address,
+    0x5a719cf3e02c17c876f6d294adb5cb7c6eb47e2f_address,
+};
+
+ABSL_FLAG(std::string, datadir, silkworm::db::default_path(), "chain DB path");
+ABSL_FLAG(uint64_t, from, 1, "start from block number (inclusive)");
+ABSL_FLAG(uint64_t, to, UINT64_MAX, "check up to block number (exclusive)");
+
+int main(int argc, char* argv[]) {
+    absl::SetProgramUsageMessage("Executes Ethereum blocks and compares resulting change sets against DB.");
+    absl::ParseCommandLine(argc, argv);
+
+    namespace fs = boost::filesystem;
+
+    fs::path db_path(absl::GetFlag(FLAGS_datadir));
+    if (!fs::exists(db_path) || !fs::is_directory(db_path) || db_path.empty()) {
+        std::cerr << absl::GetFlag(FLAGS_datadir) << " does not exist.\n";
+        std::cerr << "Use --db flag to point to a Turbo-Geth populated chaindata.\n";
+        return -1;
+    }
+
+    absl::Time t1{absl::Now()};
+    std::cout << t1 << " Checking change sets in " << absl::GetFlag(FLAGS_datadir) << "\n";
+
+    using namespace silkworm;
+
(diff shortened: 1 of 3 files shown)


cmd/scan_txs.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // counters
+    uint64_t nTxs{0}, nErrors{0};
+
+    const uint64_t from{absl::GetFlag(FLAGS_from)};
+    const uint64_t to{absl::GetFlag(FLAGS_to)};
+
+    uint64_t block_num{from};
+    for (; block_num < to; ++block_num) {
+        std::unique_ptr<lmdb::Transaction> txn{env->begin_ro_transaction()};
```
AndreaLanfranchi, line 59: No need to open a new transaction for each block.
Just use one immediately after the db opening. this will save one abort and one reopen (with new id assignment) for each loop

AndreaLanfranchi (changes requested):

cmd/scan_txs.cpp
```diff
@@ -0,0 +1,94 @@
...
+        std::cerr << absl::GetFlag(FLAGS_datadir) << " does not exist.\n";
+        std::cerr << "Use --db flag to point to a Turbo-Geth populated chaindata.\n";
+        return -1;
+    }
+
+    using namespace silkworm;
+
+    lmdb::options db_opts{};
+    db_opts.read_only = true;
+    std::shared_ptr<lmdb::Environment> env{lmdb::get_env(absl::GetFlag(FLAGS_datadir).c_str(), db_opts)};
```
AndreaLanfranchi, line 49: This statement might throw. I'd suggest to enclose the whole loop in a try catch block to avoid unhandled errors

AndreaLanfranchi (changes requested):

cmd/scan_txs.cpp
tjayrush: Thanks so much for your comments.

Two quick questions: 

I read on the LMDB documentation not to create a long-lived transaction. The process I'm running runs for hours, so I thought that would be long-lived. But now that I think about it, I think I read about 'reset' as well. Should I use a 'reset' in each loop?

This code was copied from `./check_changes.cpp`. Should I change it there too?

cmd/scan_txs.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // counters
+    uint64_t nTxs{0}, nErrors{0};
+
+    const uint64_t from{absl::GetFlag(FLAGS_from)};
+    const uint64_t to{absl::GetFlag(FLAGS_to)};
```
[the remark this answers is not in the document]
tjayrush: I'll add that and repost.

cmd/scan_txs.cpp
tjayrush: I'll add that and repost.

cmd/scan_txs.cpp
AndreaLanfranchi: Yes long lived ro transactions are discouraged but only when they're interleaved with rw transactions: an opened ro transaction may prevent rw transactions from reusing free pages in free_dbi.
If, such in this case, the procedure is meant to quickly go through blocks (using a dedicated db - i.e. TG is not running) a single transaction saves a lot of processing time.

AndreaLanfranchi: WRT the `reset` keyword. Is not actually a method of transaction itself rather the release of the shared_ptr.
Basically reset on a shared ptr invokes the destructor of the pointed class (RAII).

tjayrush: TG is running on the same machine in my case. Just to be safe, I think I'll leave it the way it is. I can write a comment.

Question: if there is a long-lasting read transaction while TG is running and the database 'grows quickly' as the docs say, does the growth correct itself over time, or is the file much larger forever (ignoring migrations or compactions)?

One of the things I wanted to do with this code is make it very simple. I'd like some day to write a tutorial on scanning the chain for data.

AndreaLanfranchi: The `growth corrects itself` under these circumstances which must both be true:
* new data is inserted in "upsert" mode (i.e. not using MDB_APPEND or MDB_APPENDDUP flags)
* the ongoing transaction can reclaim free pages from a list of pages recorded in free_dbi where the transaction id (of the freed pages) is **lower** than the id of the current transaction.

AndreaLanfranchi: > TG is running on the same machine in my case. Just to be safe,

Be aware it takes a lot longer

tjayrush: I actually ran the tool with both a single long-lasting database transaction and a per-loop transaction, and I found no obvious difference. In fact, in some cases, the per-loop database transaction was faster. Here's some numbers:

https://docs.google.com/spreadsheets/d/13omV5sfpKa6dI-2nB7PgbHEqtKvymCpdcZifv1TlDlU/edit#gid=461497860

See the timings tab.

cmd/scan_txs.cpp
```diff
@@ -0,0 +1,94 @@
...
+int main(int argc, char* argv[]) {
+    absl::SetProgramUsageMessage("Executes Ethereum blocks and compares resulting change sets against DB.");
+    absl::ParseCommandLine(argc, argv);
+
+    namespace fs = boost::filesystem;
+
+    fs::path db_path(absl::GetFlag(FLAGS_datadir));
+    if (!fs::exists(db_path) || !fs::is_directory(db_path) || db_path.empty()) {
+        std::cerr << absl::GetFlag(FLAGS_datadir) << " does not exist.\n";
+        std::cerr << "Use --db flag to point to a Turbo-Geth populated chaindata.\n";
```
AndreaLanfranchi: The output warning should be for --datadir flag (not -db)

cmd/scan_txs.cpp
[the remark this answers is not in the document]
AndreaLanfranchi: Seems to me the check is still missing.
If user enters a value for from >= value of to then program concludes without any error yet not producing any error message. I'd personally add something like :
```
if(from >= to) {
  std::cerr >> "Invalid scan range [" << from << " .. " << to ")" << std::endl;
  return -1;
}

AndreaLanfranchi (approved):
See the two tiny comments.
For the rest looks good to me.

Edit ... sorry saw older commits.
All ok for me.

PR #82, opened 2020-11-03
Lmdb3

AndreaLanfranchi:
* New simpler interfaces for database opening
* Custom comparators for tables
* Protection for mdb_mapsize to avoid data file truncation
* secp256k1 settings for GMP on MSVC (yet slower cause it supports only inverse fields)


silkworm/db/chaindb.hpp
```diff
@@ -24,11 +24,15 @@
 
 #include <lmdb/lmdb.h>
 
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+#include <boost/interprocess/mapped_region.hpp>
```
yperbasis, line 29: It looks like the includes are only used in the .cpp file, pls move them there.

yperbasis (approved):

silkworm/db/chaindb.hpp
AndreaLanfranchi: True for algorithm and interprocess
Filesystem is instead imported at higher level by /cmd utilities (through this .hpp)

PR #84, opened 2020-11-05
Silksnake integration: extract StateBuffer interface from db::Buffer

canepat:
This PR introduces the changes needed to call `ExecutionProcessor::execute_transaction` from [Silksnake](https://github.com/torquem-ch/silksnake).

Basically this means being able to replace `db::Buffer` in `IntraBlockState` constructor with a pure virtual base class, which then Silksnake implements in a `RemoteBuffer` class accessing the data through KV gRPC.

Moreover there's a couple of minor changes:
- add pass-by-ref in `db::Buffer::insert_header` method
- add Goerli chain configuration

### DIFF
--- a/silkworm/db/buffer.cpp
+++ b/silkworm/db/buffer.cpp
@@ -228,7 +228,7 @@ void Buffer::insert_receipts(uint64_t block_number, const std::vector<Receipt>& 
     }
 }
 
-void Buffer::insert_header(BlockHeader block_header) {
+void Buffer::insert_header(BlockHeader& block_header) {
     Bytes rlp{};
     rlp::encode(rlp, block_header);
     ethash::hash256 hash{keccak256(rlp)};
--- a/silkworm/db/buffer.cpp
+++ b/silkworm/db/buffer.cpp
@@ -228,12 +228,12 @@ void Buffer::insert_receipts(uint64_t block_number, const std::vector<Receipt>& 
     }
 }
 
-void Buffer::insert_header(BlockHeader& block_header) {
+void Buffer::insert_header(const BlockHeader& block_header) {
     Bytes rlp{};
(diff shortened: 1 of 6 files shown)


silkworm/db/buffer.cpp
```diff
@@ -228,7 +228,7 @@ void Buffer::insert_receipts(uint64_t block_number, const std::vector<Receipt>&
     }
 }
 
-void Buffer::insert_header(BlockHeader block_header) {
+void Buffer::insert_header(BlockHeader& block_header) {
```
yperbasis, line 231: OK, but pls then pass by const ref and remove std::move in the body

yperbasis (changes requested):

silkworm/state/intra_block_state.hpp
```diff
@@ -51,9 +51,9 @@ class IntraBlockState {
     IntraBlockState(const IntraBlockState&) = delete;
     IntraBlockState& operator=(const IntraBlockState&) = delete;
 
-    explicit IntraBlockState(db::Buffer& db) noexcept : db_{db} {}
+    explicit IntraBlockState(db::StateBuffer& db) noexcept : db_{db} {}
 
-    db::Buffer& db() { return db_; }
+    db::StateBuffer& db() const { return db_; }
```
yperbasis, line 56: I wouldn't mark this method as const since it returns a non-const reference.

silkworm/db/buffer.cpp
canepat: Sure, sorry for missing that

silkworm/state/intra_block_state.hpp
canepat: Oh yes, my bad

yperbasis (approved):

PR #88, opened 2020-11-06
Logger: print time in the local timezone with subseconds to match TG

yperbasis:


gcolvin:
I'm not finding the info in the docs. (But not taking time to read them all either.)  What does the final time string look like? @AlexeyAkhunov and I were wanting UTC.

Also, I measured the formatted C+ IO as taking long enough that subsecond resolution was not too useful.  Maybe this library is better that way?  Otherwise I'd rather stick with the standard library.

yperbasis:
@gcolvin The output looks something like
`INFO [11-06|17:51:31.151] Flushed buffer file`
I'm using local time zone because that's what TG does at the moment.
And printing subseconds up to 3 decimal points again because that's what TG does. Abseil is heavily used at Google, so it's most likely optimized well.

silkworm/common/log.hpp
```diff
@@ -34,8 +35,8 @@ class Logger {
     int level() { return verbosity; }
 
     std::ostream& log(LogLevels level) {
-        std::time_t time = std::time(nullptr);
-        return stream << LogTags[level] << "[" << std::put_time(std::gmtime(&time), "%m-%d|%H:%M:%S") << "] ";
+        return stream << LogTags[level] << "["
+                      << absl::FormatTime("%m-%d|%H:%M:%E3S", absl::Now(), absl::LocalTimeZone()) << "] ";
```
AndreaLanfranchi, line 39: Per abseil documentation is preferrable to call FormatTime with an explicitly declared time zone.
Maybe setting `absl::TimeZone timezone;` and init on logger constructor ?

AndreaLanfranchi (approved):

silkworm/common/log.hpp
gcolvin: OK.  @AlexeyAkhunov and I preferred UTC, Turbogeth notwithstanding. And my measures of C++ I/O showed it to be too slow for millisecond resolution to be meaningful.  Maybe I mismeasured.

gcolvin:
```
$ cat chrono.cpp
#include <iostream>
#include <chrono>

int main() 
{
    typedef std::chrono::system_clock clock_type;

    auto now = clock_type::now();
    auto seconds = std::chrono::time_point_cast<std::chrono::seconds>(now);
    auto fraction = now - seconds;
    time_t cnow = clock_type::to_time_t(now);

    auto milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(fraction);
    std::cout << "Milliseconds: " << milliseconds.count() << '\n';
    std::cout << "Milliseconds: " << milliseconds.count() << '\n';
    std::cout << "Milliseconds: " << milliseconds.count() << '\n';
    std::cout << "Milliseconds: " << milliseconds.count() << '\n';
    std::cout << "Milliseconds: " << milliseconds.count() << '\n';
}
Silver:build gcolvin$ g++ -std=c++11 -O3 chrono.cpp
Silver:build gcolvin$ ./a.out
Milliseconds: 56
Milliseconds: 56
Milliseconds: 56
Milliseconds: 56
Milliseconds: 56
```

gcolvin:
Measurement seems fast enough, so that's good.  I'd prefer using the C++11 chrono facility rather than a third party.
```
#include <iostream>
#include <chrono>

void print_millis()
{
  using namespace std::chrono;
  auto now = std::chrono::system_clock::now();
  auto duration = now.time_since_epoch();
  auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
  std::cout << "millis: " << millis << std::endl;
}

int main() {
  print_millis();
  print_millis();
  print_millis();
  return 0;
}
```
```
$ g++ -std=c++11 -O3 chrono.cpp
$ ./a.out
millis: 1604691476459
millis: 1604691476459
millis: 1604691476459

AndreaLanfranchi:
My two cents.
I'd prefer UTC time too to have a unified view of log timings we may collect from testers across different time zones.
In the end the chain is unique worldwide

yperbasis:
> My two cents.
> I'd prefer UTC time too to have a unified view of log timings we may collect from testers across different time zones.
> In the end the chain is unique worldwide

OK, I've made UTC the default.

PR #89, opened 2020-11-09
[Optimisation] Don't buffer writes of receipts and change sets

yperbasis:


silkworm/db/access_layer.cpp
```diff
@@ -275,4 +275,16 @@ bool read_storage_mode_receipts(lmdb::Transaction& txn) {
     return val && val->length() == 1 && (*val)[0] == 1;
 }
 
+void write_receipts(lmdb::Transaction& txn, uint64_t block_number, const std::vector<Receipt>& receipts) {
+    auto log_table{txn.open(table::kLogs)};
+    for (uint32_t i{0}; i < receipts.size(); ++i) {
+        if (!receipts[i].logs.empty()) {
+            log_table->put(log_key(block_number, i), cbor_encode(receipts[i].logs));
```
AndreaLanfranchi, line 282: Why don't use put_append here ? Should be faster as the cursor simply moves to last record

AndreaLanfranchi (approved):

AskAlexSharov:
using Append for changesets also?

yperbasis:
> using Append for changesets also?

No, not for change sets. Perhaps we should do it for change sets as well. Let me measure if it brings any performance gains.

PR #91, opened 2020-11-10
Lmdb4

AndreaLanfranchi:
Make execute use chaindb without duplicating table names
Also leverage checks about data protection on Windows (don't accept map sizes below data file size)

### DIFF
--- a/cmd/dbtool.cpp
+++ b/cmd/dbtool.cpp
@@ -342,9 +342,9 @@ dbTablesInfo get_tablesInfo(std::shared_ptr<lmdb::Environment>& env) {
 
     int rc{unnamed->get_first(&key, &data)};
     while (rc == MDB_SUCCESS) {
-        // if (data.mv_size < sizeof(size_t)) {
-        //    lmdb::err_handler(MDB_INVALID);
-        //}
+
+        //auto dataview{ db::from_mdb_val(data) };
+        //std::cout << std::setw(24) << std::left << (const char*)key.mv_data << to_hex(dataview) << std::endl;
 
         auto named = tx->open({(const char*)key.mv_data});
         table = new dbTableEntry{named->get_dbi(), named->get_name()};
--- a/cmd/dbtool.cpp
+++ b/cmd/dbtool.cpp
@@ -166,38 +166,6 @@ void sig_handler(int signum) {
     shouldStop = true;
 }
 
-std::optional<uint64_t> parse_size(const std::string& strsize) {
-    if (strsize.empty()) return {0};
-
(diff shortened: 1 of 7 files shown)


cmd/dbtool.cpp
```diff
@@ -122,7 +122,7 @@ struct dbFreeInfo {
 struct db_options_t {
     std::string datadir{silkworm::db::default_path()};  // Where data file is located
     std::string mapsize_str{};                          // Provided map_size literal
-    size_t mapsize{0};                                  // Computed map size
+    uint64_t mapsize{0};                                // Computed map size
```
yperbasis, line 125: Please leave it as size_t because otherwise it fails to compile for me on macOS.

cmd/execute.cpp
```diff
@@ -16,96 +16,142 @@
...
 
-static constexpr const char* kMigrationsTable{"migrations"};
-static constexpr const char* kDatabaseInfoTable{"DBINFO"};
-static constexpr const char* kSyncStageProgressTable{"SSP2"};
-static constexpr const char* kExecutionStage{"Execution"};
-static constexpr const char* kStorageModeReceipts{"smReceipts"};
+static constexpr const char* KExecutionStage_key{"Execution"};
+static constexpr const char* KStorageModeReceipts_key{"smReceipts"};
+
+std::optional<uint64_t> parse_size(const std::string& strsize) {
```
yperbasis, line 33: Let's move parse_size into core silkworm to avoid code duplication between check_senders, dbtool, and execute.

cmd/execute.cpp
```diff
@@ -115,59 +161,84 @@ int main(int argc, char* argv[]) {
...
+    std::unique_ptr<silkworm::lmdb::Transaction> txn{nullptr};
+
+    try
+    {
+        env = silkworm::lmdb::get_env(db_config);
+        txn = env->begin_rw_transaction();
+        bool write_receipts{storage_mode_has_write_receipts(txn)};
+        if (write_receipts && (!migration_happened(txn, "receipts_cbor_encode") ||
+                               !migration_happened(txn, "receipts_store_logs_separately"))) {
+            throw std::logic_error("Legacy stored receipts are not supported");
```
yperbasis, line 192: Why throw exception and then catch it later? What was wrong with the immediate return?

yperbasis (changes requested):

cmd/execute.cpp
AndreaLanfranchi: Nothing wrong particularly.
As a mindset I always try to keep returns from main to the very bottom

yperbasis (approved):

PR #93, opened 2020-11-11
Lookups

AndreaLanfranchi:
Quick single record data retrieve and upsert

### DIFF
--- a/cmd/execute.cpp
+++ b/cmd/execute.cpp
@@ -18,7 +18,6 @@
 #include <boost/endian/conversion.hpp>
 #include <iostream>
 #include <limits>
-#include <regex>
 #include <stdexcept>
 
 #include <silkworm/db/chaindb.hpp>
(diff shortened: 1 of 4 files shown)


yperbasis:
There are already convenience methods get and put in db::Table that handle ByteViews and check errors. For example, already_executed_block can be written as
```
    auto progress_table{txn->open(db::table::kSyncStageProgress)};
    ByteView stage_key{byte_view_of_c_str(KExecutionStage_key)};
    std::optional<ByteView> data{progress_table->get(stage_key)};
    if (data) {
        return boost::endian::load_big_u64(data->data());
    } else {
        return 0;
    }
```

AndreaLanfranchi:
You missed the point of a generic method.
Maybe a discussion before closing abruptly ?

yperbasis:
OK, so what's the point?

AndreaLanfranchi:
Nevermind ... I see my support (if any) not well perceived.
You can remove my rights from this repo in any moment.

PR #95, opened 2020-11-12
Simplify cmd/execute. Set default LMDB map size to dynamic on Windows and to 2TiB on non-Windows

yperbasis:


cmd/execute.cpp
```diff
@@ -128,82 +69,70 @@ int main(int argc, char* argv[]) {
...
-        std::clog << "Invalid --lmdb.mapSize value provided : " << map_size_str << std::endl;
+        SILKWORM_LOG(LogError) << "Invalid --lmdb.mapSize value provided : " << map_size_str << std::endl;
         return -2;
     }
 
-    std::clog << "Starting block execution. DB: " << db_file << std::endl;
+    SILKWORM_LOG(LogInfo) << "Starting block execution. DB: " << db_file << std::endl;
 
-    silkworm::lmdb::DatabaseConfig db_config{db_path, *map_size};
+    lmdb::DatabaseConfig db_config{db_path, *map_size};
```
AndreaLanfranchi, line 90: You don't need to set (few lines above) this
`std::string map_size_str{"1TB"};`

And simply change this to
```
lmdb::DatabaseConfig db_config{db_path};
if (*map_size) db_config.map_size = map_size;
```

In this way we're consistent with a single place where default map_size is defined (i.e. ind chaindb.hpp) and we avoid multiple defaults here and there.

cmd/execute.cpp
```diff
@@ -128,82 +69,70 @@ int main(int argc, char* argv[]) {
...
     db_config.set_readonly(false);
-    std::shared_ptr<silkworm::lmdb::Environment> env{nullptr};
-    std::unique_ptr<silkworm::lmdb::Transaction> txn{nullptr};
+    std::shared_ptr<lmdb::Environment> env{lmdb::get_env(db_config)};
```
AndreaLanfranchi, line 92: This may throw.
Without a try catch you may have program dump

AndreaLanfranchi (changes requested):

cmd/execute.cpp
yperbasis: Good idea, done

cmd/execute.cpp
```diff
@@ -77,11 +77,14 @@ int main(int argc, char* argv[]) {
         return -1;
     }
 
-    // Check provided map size is valid
-    auto map_size{parse_size(map_size_str)};
-    if (!map_size.has_value()) {
-        SILKWORM_LOG(LogError) << "Invalid --lmdb.mapSize value provided : " << map_size_str << std::endl;
-        return -2;
+    std::optional<size_t> map_size{std::nullopt};
```
AndreaLanfranchi, line 80: The previous version was correct and shorter.
parse_size already returns 0 if string is empty

cmd/execute.cpp
```diff
@@ -92,7 +95,10 @@ int main(int argc, char* argv[]) {
 
     SILKWORM_LOG(LogInfo) << "Starting block execution. DB: " << db_file << std::endl;
 
-    lmdb::DatabaseConfig db_config{db_path, *map_size};
+    lmdb::DatabaseConfig db_config{db_path};
+    if (map_size) {
```
AndreaLanfranchi, line 99: This imho is not correct.
if(map_size) tells if map_size has value or not.
What you might want is whether the value is zero or not (which is *map_size)

cmd/execute.cpp
yperbasis: On macOS the system displays the uncaught exception. What happens on Windows?

AndreaLanfranchi: Same on Windows. A pop-up appears for un-catched exception.
But we both work on graphical UI with terminal open.
On a headless linux program dumps memory without detail

yperbasis: Right. I've restored the try/catch to be on the safe side.

cmd/execute.cpp
yperbasis: Well, we want to leave the default mapSize value in lmdb::DatabaseConfig as is by default, not set it to 0, right? So the previous version wasn't quite correct when --lmdb.mapSize was not provided.
Also, I don't think that parse_size should return 0 if the input is empty. It should probably return std::nullopt since parse_size should only concern itself with parsing sizes, not setting defaults.

cmd/execute.cpp
yperbasis: My intention is to leave the default map_size in lmdb::DatabaseConfig as is (1TB by default) if no --lmdb.mapSize option is provided.

yperbasis: If, on the other hand, the user explicitly passes "--lmbd.mapSize 0", that should be honoured and not overwritten.

yperbasis: Though the user has no idea that 0 means something special w/o looking into the code. Perhaps we can support a special input token, e.g. "--lmbd.mapSize dynamic", instead of arguably abusing 0.

AndreaLanfranchi: > If, on the other hand, the user explicitly passes "--lmbd.mapSize 0", that should be honoured and not overwritten.

Zero mapsize won't be honored in anyway.

- If the datafile is new minimum allocation size is size of metapages for 2 system tables
- If the datafile already exists any mapSize **below** data-file-size can't be honored or you end up with a mapped size shorter than data with meta pages pointing to un-mapped region (and on Windows truncates the file)
- If the data file already exists any mapSize **above** data-file-size behaves differently on *nix vs windows: in first case has no effect on data-file size, on the latter it widens the file.

An lmdb.mapSize == 0 means : "Open the file with a map_size exactly matching data-file-size" which is actually what chaindb.cpp does for any value < data-file-size

AndreaLanfranchi: I understand that dealing with defaults is a PIA especially if we want to portability.

Correct me if I'm wrong : you see the omission of --lmdb.mapSize as "assume the default size hardcoded". This has a drawback on windows where we'll end up with min 1TB file sizes.

I see the omission of --lmdb.mapSize as "Open the datafile with a map-size == data-file-size" which produces a consistent result amonf Win and *nix i.e. the file will likely report MDB_MAPFULL if try to insert data when no space left.

AndreaLanfranchi: As an alternative we can set a directive conditional on chaindb.cpp to set the default to 0 for win and to 1TB for *nix

AndreaLanfranchi: I reiterate.

On Windows always `map_size == data-file-size`. You cannot set a map_size greater than available storage space on the volume.

On *nix you can set any map_size (eventually greater than disk or volume) and data-file-size will grow accordingly to effective data put in.

yperbasis: Yes, that's right. IMHO we should have sensible defaults. 1TB is not perfect, but for the time being it works well for both Windows and Linux/macOS.
Another possibility is to make the dynamic map size to work better as the default. For instance, what of we set the map_size to 120%, not 100% the file size to avoid MDB_MAPFULL. Perhaps we can take some inspiration from MDBX. But that'll require some work.
All in all, I'd set the default to 1TB until the dynamic one is more robust if we want to invest time in the dynamic mode.

yperbasis: OK, let's set the default to dynamic on Windows and to 1TB on Linux/macOS.

AndreaLanfranchi (approved):
Ok.
Would have done the setting of dynamic default in chaindb though

yperbasis:
> Ok.
> Would have done the setting of dynamic default in chaindb though

Yes, good idea. Done.

AndreaLanfranchi (approved):

PR #98, opened 2020-11-13
Guard against DupSort change sets

yperbasis:
Silkworm does not yet support https://github.com/ledgerwatch/turbo-geth/pull/1342

### DIFF
--- a/tg_api/silkworm_tg_api.cpp
+++ b/tg_api/silkworm_tg_api.cpp
@@ -47,14 +47,14 @@ SILKWORM_EXPORT SilkwormStatusCode silkworm_execute_blocks(MDB_txn* mdb_txn, uin
         if (write_receipts && (!db::migration_happened(txn, "receipts_cbor_encode") ||
                                !db::migration_happened(txn, "receipts_store_logs_separately"))) {
             SILKWORM_LOG(LogError) << "Legacy stored receipts are not supported" << std::endl;
-            return kIncompatibleDbFormat;
+            return kSilkwormIncompatibleDbFormat;
         }
 
         // https://github.com/ledgerwatch/turbo-geth/pull/1342
         if (db::migration_happened(txn, "acc_change_set_dup_sort_18") ||
             db::migration_happened(txn, "storage_change_set_dup_sort_22")) {
             SILKWORM_LOG(LogError) << "DupSort change sets are not supported yet" << std::endl;
-            return kIncompatibleDbFormat;
+            return kSilkwormIncompatibleDbFormat;
         }
 
         db::Buffer buffer{&txn};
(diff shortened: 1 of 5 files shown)


tg_api/silkworm_tg_api.cpp
```diff
@@ -34,17 +34,33 @@ SILKWORM_EXPORT SilkwormStatusCode silkworm_execute_blocks(MDB_txn* mdb_txn, uin
...
         auto cleanup{gsl::finally([&txn] { *txn.handle() = nullptr; })};  // avoid aborting mdb_txn
 
+        if (write_receipts && (!db::migration_happened(txn, "receipts_cbor_encode") ||
+                               !db::migration_happened(txn, "receipts_store_logs_separately"))) {
+            SILKWORM_LOG(LogError) << "Legacy stored receipts are not supported" << std::endl;
+            return kIncompatibleDbFormat;
+        }
+
+        // https://github.com/ledgerwatch/turbo-geth/pull/1342
+        if (db::migration_happened(txn, "acc_change_set_dup_sort_18") ||
```
AndreaLanfranchi, line 54: I believe we should sync with TG team to avoid this kind of constants to check migrations.
We should have some sort of values to set the "SchemaVersion" of the db. Every time a migration occurs it should increase schema version. So we can know at which level we're compatible on db opening).
All this constants require to change code again and again.

(I know this is not related to this specific PR but is a suggestion to prevent further re-dos)

AndreaLanfranchi (approved):

tg_api/silkworm_tg_api.cpp
AskAlexSharov: But shema version will only increase confusion in your code. Your code will look like: 
```
if shcemaVersion > 15 { // means changesets are dupsorted
   // do one thing
} 
// do another thing
```

Which has no advantage over `if (db::migration_happened(txn, "acc_change_set_dup_sort_18")` - because if you forget write comment then you will stay with meaningless `if schemaVersion > 15`. 

----
ufff... maybe some day I will have energy to implement things Silkworm also.

AndreaLanfranchi: My point is slightly different ... I don't mean to adapt Silkworm to any possible schema version.
Instead I mean to assert silkworm is compatible **to a discrete schema version**.
For instance at db opening an immediate check over schema version should throw if detected schema != hardcoded schema into silkworm. This should save a lot of 
```
if (this_condition) { 
   // do something
} 
if (this_other_condition) { 
   // do something else
} 
[...]
```

Bottom line ... we can check development "height" using git commits hashes. Why can't we check db schema version to check both envs are on the same page ? The "if migration_happened" seems pretty expensive to me as it may imply some migration has happened and some other not (maybe a crash during migration ?) forcing the code to adapt to dirty situations.

PR #103, opened 2020-11-18
Switch to DupSort changes

yperbasis:
Port https://github.com/ledgerwatch/turbo-geth/pull/1342

For the StorageChangeSets now, the DupSort key is blockNum-bigEndian-64bit | address | incarnation, and the DupSort value is location | value.

And for AccountChangeSets, the DupSort key is blockNum-bigEndian-64bit and the DupSort value is address | value.


silkworm/db/tables.hpp
```diff
@@ -54,26 +54,20 @@ constexpr lmdb::TableConfig kLogAddressIndex{"log_address_index"};
 constexpr lmdb::TableConfig kLogTopicIndex{"log_topic_index"};
 constexpr lmdb::TableConfig kLogs{"log"};
 constexpr lmdb::TableConfig kMigrations{"migrations"};
-constexpr lmdb::TableConfig kPlainAccountChangeSet{"PLAIN-ACS"};
+constexpr lmdb::TableConfig kPlainAccountChangeSet{"PLAIN-ACS", MDB_DUPSORT};
```
AndreaLanfranchi, line 57: Just to clarify ... are those tables changed in place or renamed due to a migration ?
I.e. shouldn't they be something like "PLAIN-ACS2" ?

The rest looks good to me.

AndreaLanfranchi (approved):

silkworm/db/tables.hpp
yperbasis: The table names stayed the same: https://github.com/ledgerwatch/turbo-geth/pull/1342/files#diff-b2285bb6007ca9e5c4056d5365a657cdf0515d19bff46268d845c5f665799cc3

PR #106, opened 2020-11-19
Support including Silkworm as part of another cmake project

yperbasis:


canepat (approved):
Looks good to me, thanks

PR #107, opened 2020-11-30
Store transactions individually

yperbasis:
Port TG DB migration https://github.com/ledgerwatch/turbo-geth/pull/1358

### DIFF
--- a/silkworm/db/access_layer.cpp
+++ b/silkworm/db/access_layer.cpp
@@ -38,12 +38,12 @@ std::optional<BlockHeader> read_header(lmdb::Transaction& txn, uint64_t block_nu
 }
 
 // TG ReadTransactions
-static std::vector<Transaction> read_transactions(lmdb::Transaction& txn, uint64_t base_id, uint64_t amount) {
+static std::vector<Transaction> read_transactions(lmdb::Transaction& txn, uint64_t base_id, uint64_t count) {
     std::vector<Transaction> v;
-    if (amount == 0) {
+    if (count == 0) {
         return v;
     }
-    v.reserve(amount);
+    v.reserve(count);
 
     auto table{txn.open(table::kEthTx)};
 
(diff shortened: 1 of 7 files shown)


AndreaLanfranchi (approved):

silkworm/db/access_layer.cpp
```diff
@@ -37,6 +37,35 @@ std::optional<BlockHeader> read_header(lmdb::Transaction& txn, uint64_t block_nu
     return header;
 }
 
+// TG ReadTransactions
+static std::vector<Transaction> read_transactions(lmdb::Transaction& txn, uint64_t base_id, uint64_t amount) {
```
AndreaLanfranchi, line 41: As `amount` refers to the number of records to load I'd change the name to `count`

silkworm/db/access_layer_test.cpp
```diff
@@ -110,11 +111,23 @@ namespace db {
             CHECK(!read_block(*txn, block_num, read_senders));
 
             BlockBody body{sample_block_body()};
-            rlp.clear();
-            rlp::encode(rlp, body);
+
+            detail::BlockBodyForStorage storage_body;
+            storage_body.base_txn_id = 1687896;
+            storage_body.txn_amount = body.transactions.size();
```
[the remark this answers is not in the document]
yperbasis: > To avoid confusion with effective transactions amount I'd prefer to rename `txn_amount` with `txn_count`

done

silkworm/db/access_layer.cpp
yperbasis: > As `amount` refers to the number of records to load I'd change the name to `count`

done

silkworm/db/access_layer.cpp
```diff
@@ -37,6 +37,35 @@ std::optional<BlockHeader> read_header(lmdb::Transaction& txn, uint64_t block_nu
...
+    MDB_val key_mdb{to_mdb_val(txn_key)};
+    MDB_val data_mdb;
+
+    uint64_t i{0};
+    for (int rc{table->seek_exact(&key_mdb, &data_mdb)}; rc != MDB_NOTFOUND && i < count;
+         rc = table->get_next(&key_mdb, &data_mdb), ++i) {
+        lmdb::err_handler(rc);
+        ByteView data{from_mdb_val(data_mdb)};
+
+        Transaction eth_txn;
```
AndreaLanfranchi, line 61: Can we have a Transaction constructor with ByteView so we can `emplace_back` directly into vector ?

yperbasis: > Can we have a Transaction constructor with ByteView so we can `emplace_back` directly into vector ?

I suspect the resulting performance gain will be insignificant. Also, I'd like to avoid coupling Transaction too tightly to its RLP serialization.

PR #114, opened 2020-12-13
added etl library

Giulio2002:


db/silkworm/etl/dataprovider/fileprovider.cpp
```diff
@@ -0,0 +1,59 @@
...
+FileProvider::FileProvider(Buffer * b, int i) {
+    file.open("tmp" + std::to_string(i), std::ios_base::in | std::ios_base::out | std::ios_base::trunc);
+    filename = "tmp" + std::to_string(i);
+    b->sort();
+    auto entries{b->getEntries()};
+    for(auto e: *entries) {
+        auto k{e.k};
+        auto v{e.v};
+        auto flow{std::string()};
+        flow.push_back((char) k.size());
```
yperbasis, line 29: What happens if k.size() or v.size() is greater than 255?

yperbasis (changes requested):

db/silkworm/etl/dataprovider/fileProvider.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef ETL_FILE_PROVIDER_H
+#define ETL_FILE_PROVIDER_H
+
+namespace silkworm::etl{
+
+class FileProvider {
+    public:
+
+        FileProvider(Buffer *, int);
```
yperbasis, line 17: What is this int argument? Please give it a name.

db/silkworm/etl/buffers/buffer.cpp
```diff
@@ -0,0 +1,42 @@
+#include <silkworm/etl/buffers/buffer.hpp>
+#include <algorithm>
+
+namespace silkworm::etl{
+
+Buffer::Buffer(size_t _optimalSize) {
+    optimalSize = _optimalSize;
```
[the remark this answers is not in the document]
AndreaLanfranchi: If I may add my two cents
Member fields should be better initialized in constructor as
```
Buffer::Buffer(size_t _optimalSize) : optimal_size_{optimal_size) {
    [...];
```

db/silkworm/etl/collector/collector.hpp
```diff
@@ -0,0 +1,30 @@
+#include <silkworm/etl/dataprovider/fileProvider.hpp>
+#include <silkworm/etl/buffers/buffer.hpp>
+#include <silkworm/db/chaindb.hpp>
+#include <dirent.h>
+
+#define IDEAL_SIZE 268435456
```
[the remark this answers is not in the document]
AndreaLanfranchi: Imho better use `constexpr` as they're always evaluated at compile time.
`const` can be either.

db/silkworm/etl/collector.hpp
```diff
@@ -0,0 +1,48 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#include <silkworm/etl/fileProvider.hpp>
+#include <silkworm/etl/buffer.hpp>
+#include <silkworm/db/chaindb.hpp>
+#include <dirent.h>
```
yperbasis, line 20: Don't use non-standard headers as we also support compilation on Windows; use Boost filesystem instead.

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,64 @@
...
+    auto value_length = boost::endian::load_big_u32((unsigned char *)buffer_value_length);
+    auto key{new unsigned char[key_length]};
+    auto value{new unsigned char[value_length]};
+    file_.read((char *)key, key_length);
+    file_.read((char *)value, value_length);
+    return {ByteView(key, key_length), ByteView(value, value_length)};
+}
+
+void FileProvider::reset() {
+    std::remove(filename_.c_str());
```
AndreaLanfranchi, line 61: Definitely this does not delete the file.
Also remember to close it before deleting.

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <silkworm/etl/file_provider.hpp>
+
+namespace silkworm::etl{
+
+FileProvider::FileProvider(Buffer * buffer, int id) {
+    filename_ = "tmp" + std::to_string(id);
+    file_.open(filename_, std::ios_base::in | std::ios_base::out | std::ios_base::trunc);
+    auto entries{buffer->get_entries()};
+    for(auto entry: entries) {
```
AndreaLanfranchi, line 25: I'd avoid making complex operations in class constructor as it gets difficult to determine if the initialization fails or any of the operations fail.
In this case however your making multiple copies of the same data:
- you're making a copy of each entry in entries
- you're copying data into a string
- you're multiple casting string into bytes

Also I don't understand the point of saving data lengths in BE. Native endianness should be enough as file contents are supposedly read from within the same architecture ... this saves the overhead of byte swapping.
A suggestion follows

Please also take note that this code does not account possible errors while writing files (eg. disk full)

```suggestion
    filename_ = "tmp" + std::to_string(id);
    file_.open(filename_, std::ios_base::in | std::ios_base::out | std::ios_base::trunc);

    union head_t {
        uint32_t lengths[2];
        uint8_t bytes[8];
    };

    head_t head{};

    for(const auto& entry: buffer->get_entries()) {
        head.lengths[0] = (uint32_t)entry.key.size();
        head.lengths[1] = (uint32_t)entry.value.size();
        file_.write((const char *)head.bytes, 8);
        file_.write((const char *)entry.key.data(), entry.key.size());
        file_.write((const char *)entry.value.data(), entry.value.size());
    }
    file_.seekp(0);
```

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,64 @@
...
+        unsigned char length_value[4];
+        boost::endian::store_big_u32(length_key, entry.key.size());
+        boost::endian::store_big_u32(length_value, entry.value.size());
+        writes.append(std::string((const char *) length_key, 4));
+        writes.append(std::string((const char *) length_value, 4));
+        writes.append(std::string((const char *) entry.key.data(), entry.key.size()));
+        writes.append(std::string((const char *) entry.value.data(), entry.value.size()));
+        file_ << writes;
+    }
+    file_.seekp(0);
```
AndreaLanfranchi, line 37: For file stream it doesn't really matter but I'd prefer to use seekg instead of seekp to clarify you're setting the pointer for subsequent reads.

AndreaLanfranchi (changes requested):

db/silkworm/etl/file_provider.cpp
Giulio2002: https://en.cppreference.com/w/cpp/io/c/remove It does, but it's just being confusing since there are many `remove`, like in `algorithm`. I will replace
 it with the `boost::filesystem` one

PR #115, opened 2020-12-14
Copy Keccak part of ethash and remove the dependency

yperbasis:
I'm trying to reduce silkworm_core's dependencies so that it could be compiled into WebAssembly.


AndreaLanfranchi (commented):
The keccakf800 function is not needed unless we want to also implement ProgPoW sealing algo.

ethash/hash_types.h
```diff
@@ -0,0 +1,50 @@
+/* ethash: C/C++ implementation of Ethash, the Ethereum Proof of Work algorithm.
+ * Copyright 2018-2019 Pawel Bylica.
+ * Licensed under the Apache License, Version 2.0.
+ */
+
+#pragma once
+
+#include <stdint.h>
+
+#ifdef __cplusplus
```
AndreaLanfranchi, line 10: Don't think this extern C block is needed any more

yperbasis:
> The keccakf800 function is not needed unless we want to also implement ProgPoW sealing algo.

Good point – I've removed it.

ethash/hash_types.h
yperbasis: > Don't think this extern C block is needed any more

You're probably right since there are no functions nor variables with external linkage in there. On the other side, there's no harm in having extern "C", we might as well leave it as is. (I took the files verbatim from https://github.com/chfast/ethash)

ethash/keccak.c
```diff
@@ -0,0 +1,382 @@
...
+    keccakf1600_implementation(state);
+}
+
+/// The pointer to the best Keccak-f[1600] function implementation,
+/// selected during runtime initialization.
+static void (*keccakf1600_best)(uint64_t[25]) = keccakf1600_generic;
+
+
+#if defined(__x86_64__) && __has_attribute(target)
+__attribute__((target("bmi,bmi2"))) static void keccakf1600_bmi(uint64_t state[25])
```
AndreaLanfranchi, line 282: Not sure I undersand this ... they both refer to same implementation.
Am I missing something ?

ethash/hash_types.h
AndreaLanfranchi: Ok

ethash/keccak.c
yperbasis: > Not sure I undersand this ... they both refer to same implementation.
> Am I missing something ?

They both use the same code, but keccakf1600_bmi is faster because it's compiled for CPUs that support [BMI1/BMI2 instructions](https://en.wikipedia.org/wiki/Bit_manipulation_instruction_set) - see `__attribute__((target("bmi,bmi2")))`

`select_keccakf1600_implementation` selects an implementation at runtime, depending on whether the CPU supports BMI1 & BMI2.

AndreaLanfranchi (approved):

Issue #111, opened 2020-12-09
Compile core into WebAssembly


yperbasis:
Strategic goal for the Internet Computer project by https://dfinity.org

PR #116, opened 2020-12-15
Move worker from silkworm_core to silkworm_db

yperbasis:
Rationale: we want to compile silkworm_core into WebAssembly (Issue #111), which doesn't support multi-threading yet.


AndreaLanfranchi (approved):
Just for my understanding... Does not support means doesn't build ? I mean a thread initialization instruction is an error despite not being used ?

yperbasis:
> Just for my understanding... Does not support means doesn't build ? I mean a thread initialization instruction is an error despite not being used ?

Currently WebAssembly doesn't support threads, so there're no thread-related instructions. I'm using https://github.com/WebAssembly/wasi-sdk to compile into WebAssembly, and it does not provide facilities like `<atomic>`, so the compilation fails.

PR #117, opened 2020-12-15
Don't use boost/algorithm/hex.hpp since it doesn't play well with Web…

yperbasis:
…Assembly


AndreaLanfranchi (approved):
A lot of limits this webassembly :)

PR #118, opened 2020-12-16
Avoid mutex in ExecutionStatePool for Wasm compatibility

yperbasis:
Issue #111

### DIFF
--- a/core/silkworm/execution/state_pool.cpp
+++ b/core/silkworm/execution/state_pool.cpp
@@ -22,13 +22,11 @@
 
 namespace silkworm {
 
-ExecutionStatePool& ExecutionStatePool::instance() noexcept {
-    static ExecutionStatePool x;
-    return x;
-}
+ExecutionStatePool::ExecutionStatePool() {}
+
+ExecutionStatePool::~ExecutionStatePool() {}
 
 std::unique_ptr<evmone::execution_state> ExecutionStatePool::acquire() noexcept {
-    std::lock_guard lock{mutex_};
     if (pool_.empty()) {
         return std::make_unique<evmone::execution_state>();
     }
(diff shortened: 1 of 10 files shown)


core/silkworm/execution/state_pool.cpp
```diff
@@ -22,13 +22,11 @@
...
-}
+ExecutionStatePool::ExecutionStatePool() {}
+
+ExecutionStatePool::~ExecutionStatePool() {}
 
 std::unique_ptr<evmone::execution_state> ExecutionStatePool::acquire() noexcept {
-    std::lock_guard lock{mutex_};
     if (pool_.empty()) {
         return std::make_unique<evmone::execution_state>();
     }
```
AndreaLanfranchi, line 32: Not in this PR but why acquire from pool_ doesnt' use move semantic (like as in release method) ?

I mean this 
```
    std::unique_ptr<evmone::execution_state> obj{pool_.top().release()};
    pool_.pop();
    return obj;
```
could be changed to this
```
    auto obj{std::move(pool_.top())};
    pool_.pop();
    return obj;
```

AndreaLanfranchi (approved):

core/silkworm/execution/state_pool.cpp
yperbasis: > Not in this PR but why acquire from pool_ doesnt' use move semantic (like as in release method) ?
> 
> I mean this
> 
> ```
>     std::unique_ptr<evmone::execution_state> obj{pool_.top().release()};
>     pool_.pop();
>     return obj;
> ```
> 
> could be changed to this
> 
> ```
>     auto obj{std::move(pool_.top())};
>     pool_.pop();
>     return obj;
> ```

This should also probably work, I'm just not that familiar with move semantics.

PR #120, opened 2020-12-17
added ETL

Giulio2002:


db/silkworm/etl/buffer.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+void Buffer::put(ByteView key, ByteView value) {
+    size_ += value.size() + key.size();
+    entries_.push_back({key, value});
+}
+
+void Buffer::sort() {
+    std::sort(entries_.begin(), entries_.end(), compareEntries);
+    std::reverse(entries_.begin(), entries_.end());
```
AndreaLanfranchi, line 31: This adds a bit of oeverhead as the container is firtsly sorted ascending and then reversed.
You can easily transform this in a single `sort` pass by modding the compareEntries above to return the **inverse** of the key.comparator result

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,66 @@
...
+   limitations under the License.
+*/
+
+#include "file_provider.hpp"
+
+namespace silkworm::etl{
+
+FileProvider::FileProvider(int id) {
+    filename_ = "tmp" + std::to_string(id);
+    file_.open(filename_, std::ios_base::in | std::ios_base::out | std::ios_base::trunc);
```
AndreaLanfranchi, line 23: There is no check whether or not the file is good.
Do we have space on disk ? The opening of the file was succesfull ? (Maybe we don't have the right to write in this directory=

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+void FileProvider::write_buffer_to_disk(Buffer *buffer) {
+    union head_t {
+        uint32_t lengths[2];
+        uint8_t bytes[8];
+    };
```
AndreaLanfranchi, line 30: You can move the type definition into the class definition (in .hpp file) to reuse it.

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,66 @@
...
+    head_t head{};
+    for(const auto& entry: buffer->get_entries()) {
+        head.lengths[0] = entry.key.size();
+        head.lengths[1] = entry.value.size();
+        file_.write((const char *) head.bytes, 8);
+        file_.write((const char *) entry.key.data(), entry.key.size());
+        file_.write((const char *) entry.value.data(), entry.value.size());
+    }
+    file_.seekg(0);
```
AndreaLanfranchi, line 39: Are all the writes correctly flushed to disk ? (check for any error like for exampe disk full)
You actually do not know if this instance of FileProvider is valid

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+Entry FileProvider::next() {
```
AndreaLanfranchi, line 42: Personal taste comment: I'd rename this method as `read_entry`. Gives a better idea of what it does (next alone seems to me this simply advances some pointer)

db/silkworm/etl/file_provider.hpp
```diff
@@ -0,0 +1,39 @@
...
+        Entry next();                               // Read next element from file starting from position 0
+        void reset();                               // Remove the file when eof is met
+
+    private:
+        std::fstream file_;
+        std::string filename_;
+};
+
+}
+#endif
```
AndreaLanfranchi, line 39: Always add a blank line to the end of files

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    if (file_.eof()) {
+        return {silkworm::ByteView(), silkworm::ByteView()};
+    }
+
+    auto key{new unsigned char[head.lengths[0]]};
+    auto value{new unsigned char[head.lengths[1]]};
+    file_.read((char *)key, head.lengths[0]);
+    file_.read((char *)value, head.lengths[1]);
+    return {silkworm::ByteView(key, head.lengths[0]), silkworm::ByteView(value, head.lengths[1])};
```
AndreaLanfranchi, line 58: Are you sure here you have correctly read any data ?
You should check file is good

AndreaLanfranchi (changes requested):

db/silkworm/etl/file_provider.hpp
```diff
@@ -0,0 +1,39 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+#ifndef ETL_SILKWORM_FILE_PROVIDER_H
+#define ETL_SILKWORM_FILE_PROVIDER_H
+
+#include <fstream>
+#include <boost/filesystem/operations.hpp>
```
AndreaLanfranchi, line 20: Move this include into `file_provider.cpp`. No boost:filesystem's is used here and there's no need to include this in headers which include this one.

AndreaLanfranchi (changes requested):

db/silkworm/etl/buffer.cpp
```diff
@@ -0,0 +1,47 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#include "buffer.hpp"
+
+namespace silkworm::etl{
+bool compareEntries(const Entry lhs, const Entry rhs) {
```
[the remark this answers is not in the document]
Giulio2002: My bad, that was not supposed to be a comparer, i added < and > operators to Entry to make it better and replaced it.

AndreaLanfranchi:
I am struggling to understand what `collector` does particularly in its `load` method

db/silkworm/etl/collector.cpp
```diff
@@ -0,0 +1,78 @@
...
+        std::push_heap(heap.begin(), heap.end(), std::greater<Entry>());
+    }
+
+    while (heap.size() != 0) {
+
+        std::pop_heap(heap.begin(), heap.end(), std::greater<Entry>());
+        auto entry{heap.back()};
+        heap.pop_back();
+        auto pairs{load(entry.key, entry.value)};
+        for (auto pair: pairs) table->put(pair.key, pair.value);
```
AndreaLanfranchi, line 62:
```suggestion
        for (auto& pair : pairs) {
            table->put(pair.key, pair.value);
        }
```

db/silkworm/etl/collector.cpp
```diff
@@ -0,0 +1,78 @@
...
+    {
+        auto entry = data_providers_.at(i).read_entry();
+        heap.push_back({entry.key, entry.value, (int)i});
+        std::push_heap(heap.begin(), heap.end(), std::greater<Entry>());
+    }
+
+    while (heap.size() != 0) {
+
+        std::pop_heap(heap.begin(), heap.end(), std::greater<Entry>());
+        auto entry{heap.back()};
```
AndreaLanfranchi, line 59: Do we really need a copy of the heap data ?

AndreaLanfranchi (changes requested):

db/silkworm/etl/collector.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    auto heap = std::vector<Entry>();
+    std::make_heap(heap.begin(), heap.end(), std::greater<Entry>());
+
+    for (unsigned int i = 0; i < data_providers_.size(); i++)
+    {
+        auto entry = data_providers_.at(i).read_entry();
+        heap.push_back({entry.key, entry.value, (int)i});
+        std::push_heap(heap.begin(), heap.end(), std::greater<Entry>());
+    }
```
AndreaLanfranchi, line 54: Why force the reordering of the heap at each insertion ?
You can do this only once.
Also if your index of iteration is required to be an int ... use it as an int to avoid casting from uint
```suggestion
    auto heap = std::vector<Entry>();
    for (int i{0}; i < data_providers_.size(); i++) {
        auto entry{data_providers_.at(i).read_entry()};
        entry.i = i;
        heap.push_back(entry);
    }
    std::make_heap(heap.begin(), heap.end(), std::greater<Entry>());
```

AndreaLanfranchi (changes requested):

db/silkworm/etl/collector.cpp
```diff
@@ -0,0 +1,78 @@
...
+   limitations under the License.
+*/
+
+#include "collector.hpp"
+
+namespace silkworm::etl{
+void Collector::flush_buffer() {
+    buffer_.sort();
+    data_providers_.push_back(FileProvider(data_providers_.size()));
+    data_providers_.at(data_providers_.size()-1).write_buffer_to_disk(buffer_.get_entries());
```
AndreaLanfranchi, line 23: When inserting new class elements into a vector use `emplace_back` instead of creating new instance of the class which gets then copied into the vector.
Also ... in the second line you clearly want to access the last inserted element of the vector. Use `back()` to get the reference to it.

```suggestion
    data_providers_.emplace_back(data_providers_.size());
    data_providers_.back().write_buffer_to_disk(buffer_.get_entries());
```

AndreaLanfranchi (changes requested):

AndreaLanfranchi:
@Giulio2002 
About this 
```
    while (heap.size() != 0) {

        std::pop_heap(heap.begin(), heap.end(), std::greater<Entry>());
        auto entry{heap.back()};
        heap.pop_back();
        auto pairs{load(entry.key, entry.value)};
        for (auto pair: pairs) table->put(pair.key, pair.value);
		auto next{data_providers_.at(entry.i).read_entry()};
        next.i = entry.i;
        if (next.key.size() == 0) {
            data_providers_.at(entry.i).reset();
            continue;
        }
        heap.push_back(next);
        std::push_heap(heap.begin(), heap.end(), std::greater<Entry>());
    }
```

Can you explain what it does ? My gut feeling is you're mixing a heap and a stack or a queue.
Correct me if I am wrong but my understanding is you're reading data from files and inserting results of reads sequentially into db tables ... when first file is completely read pass to next one and so on. Right ?
So why you're processing data backwards ? From last file to first one ? 
I'm puzzled

Giulio2002:
> @Giulio2002
> About this
> 
> ```
>     while (heap.size() != 0) {
> 
>         std::pop_heap(heap.begin(), heap.end(), std::greater<Entry>());
>         auto entry{heap.back()};
>         heap.pop_back();
>         auto pairs{load(entry.key, entry.value)};
>         for (auto pair: pairs) table->put(pair.key, pair.value);
> 		auto next{data_providers_.at(entry.i).read_entry()};
>         next.i = entry.i;
>         if (next.key.size() == 0) {
>             data_providers_.at(entry.i).reset();
>             continue;
>         }
>         heap.push_back(next);
>         std::push_heap(heap.begin(), heap.end(), std::greater<Entry>());
>     }
> ```
> 
> Can you explain what it does ? My gut feeling is you're mixing a heap and a stack or a queue.
> Correct me if I am wrong but my understanding is you're reading data from files and inserting results of reads sequentially into db tables ... when first file is completely read pass to next one and so on. Right ?
> So why you're processing data backwards ? From last file to first one ?
> I'm puzzled

ETL uses external merge sorting in order to put all the entries in sorted order by key. The file providers class just helps reading each entry. what it's doing is keeping in the heap (which i now replaced with a priority queue) at least one entry from each file at all times, and gradually put them in the database as they are sorted. The `i` member in the struct `Entry` just indicates from which file each entry is from when they are passed to the heap, so that when it's their turn  for being processed, we can read another entry from that file(given we are not at EOF)

Giulio2002:
> I am struggling to understand what `collector` does particularly in its `load` method

The load method is the equivalent of the LoadFunc of the golang implementation. It's just a component of the original implementation. In ETL, this function is called just before an entry is put in the db and can modify each entry RIGHT before they are put in the DB, or just do something additional.

AndreaLanfranchi:
Thank you so much for explanation but, besides forgiving me being so dumb, after your last commits I'm even more puzzled (if possible) :D

Let me recap and please correct me where I'm wrong:

- `collector` is a hybrid component : it stores data into memory (`buffer`) and whenever memory overflows suggested capacity it flushes data to file on disk. Right ? This operation is carried by ...
- `flush_buffer` which sorts buffer data **and** inserts (into data_providers_) a new instance of `FileProvider` which eventually writes down data on disk. `buffer_` is emptied and ready to begin to accept new data til max capacity reached.

Am I right so far ?

Now come my problems in understanding `Collector::load`: my understanding is that while `Collector::collect` behaves as data producer, `Collector::load` should be the data consumer. Right ? My points are :

- method `load` accepts as a parameter `*table` and `load` ... this should be avoided to have a parameter label with the same name of the called method (but this is negligible now). I assume the load parameter is a pointer to function wich returns a vector of data.
- if I enter load and I have an empty `data_providers_` vector I see `buffer_` data are sorted and written to table. Eventually buffer_ is emptied and function returns.
- if, instead, some data on disk already exists then buffer_ is flushed to disk too (regardless it has/hasn't reached capacity) and the vector of data_providers_ has incremented its size by one.

The first problem I see is you `flush_buffer` in any case which could mean you might end up writing an empty file if the buffer_ is empty which means every time I call Collector::load, if there are previously flushed files, for sure I will create a new empty one. In fact `flush_buffer` does not check buffer_ has data.

Now you create a prority queue populating it from **one read** obtained by each data_provider expecting the queue to have the smallest `Entry` on top. Here I observe you insert in the queue also empty values.
I'd rather change the range loop from this

```
    auto queue{std::priority_queue<Entry, std::vector<Entry>, std::greater<Entry>>()};
    for (auto& data_provider: data_providers_)
    {
        queue.push(data_provider.read_entry());
    }
```
to this
```
    auto queue{std::priority_queue<Entry, std::vector<Entry>, std::greater<Entry>>()};
    for (auto& data_provider : data_providers_) {
        auto item{data_provider.read_entry()};
        if (item.key.size()) {
            queue.push(item);
        }
    }
```

Even in this case however we do not know if data_provider has returned an empty item due to EOF reached (so effectively no data to process) or cause some other error has occurred. But besides this you will notice that `Entry` is defined as 
```
struct Entry {
    ByteView key;
    ByteView value;
    int i; // Used only for heap operations
};
```
but `data_provider.read_entry()` does not value that `i` so all of your inserted records now have `i==0` so by consequence all subsequent code where you refer to `data_providers_.at(entry.i)` will always point to first file.

At this point my understanding is you want to keep processing the smallest possible `Entry` obtained from any file. But this to me is a bit misleading as you don't want the smallest Entry ... but the smallest `Entry.key` So you may want to create a custom comparator for `queue` such as :
```
    auto key_comparer = [](Entry left, Entry right) { return left.key > right.key; };
    std::priority_queue<Entry, std::vector<Entry>, decltype(key_comparer) > queue(key_comparer);

```

Then in bottom loop you pick the smallest key **by copy** and immediately remove from the queue. There is no need. Simply defer the removal when you do not longer need the reference.

Other caveats :
- your `FileProvider` constructor accepts an `int` as a parameter but you create instances passing data_providers_.size() which is an unsigned long long: this causes an auto narrow casting. 
- change FileProvider so it stores in `id_` the ordinal number it has been created with: in such way you can mod the `read_entry` method to create an etl::Entry with `i` valued
- change Buffer to expose its size() so you can decide whether or not has data
- dont' use `auto` when the declared type is not from a function return: use an explicit declaration instead
- collector does not detect or handle whether or not one or more data_provider is in error
- be verbose in comments: comments help others understand

AndreaLanfranchi (approved):

db/silkworm/etl/file_provider.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+union head_t {
+    uint32_t lengths[2];
+    uint8_t bytes[8];
+};
+
+// FileProvider saves buffers to disk and reads from them
+class FileProvider {
+  public:
+    FileProvider(std::string& working_path, int id);        // Sort and write buffer to file
```
yperbasis, line 34: 1. Pass working_path by **const** reference since you're not changing the argument
2. Constructor's comment is misleading: the constructor only opens a file, it doesn't sort nor writes anything.

db/silkworm/etl/file_provider.cpp
```diff
@@ -0,0 +1,83 @@
...
+std::optional<std::pair<db::Entry, int>> FileProvider::read_entry() {
+    head_t head{};
+    file_.read((char *)head.bytes, 8);
+    if (file_.eof()) {
+        return std::nullopt;
+    }
+    if (file_.fail()) {
+        throw etl_error(strerror(errno));
+    }
+    auto key{new unsigned char[head.lengths[0]]};
```
yperbasis, line 66: There is a memory leak here: memory is allocated with `new`, but is nowhere released with `delete`.
One possible solution to avoid the leak: don't use `ByteView`, which doesn't own the memory, use `Bytes` instead. Something like

In a header file:
```
namespace silkworm::etl {

struct Entry {
    Bytes key;
    Bytes value;
    size_t size() const noexcept { return key.size() + value.size(); }
};
```

In `file_provider.cpp`:
```
// Entry from etl, not db namespace
std::optional<std::pair<Entry, int>> FileProvider::read_entry() {
  ...
  Entry entry;
  entry.key.resize(head.lengths[0]);
  entry.value.resize(head.lengths[1]);
  file_.read((char *)entry.key.data(), entry.key.size());
  file_.read((char *)entry.value.data(), entry.value.size());
  ...
```

db/silkworm/etl/collector.hpp
```diff
@@ -0,0 +1,54 @@
...
+// Collector collects entries that needs to be sorted and load them in the table in sorted order
+class Collector {
+  public:
+    Collector(const char* work_path = nullptr, size_t optimal_size = kOptimalBufferSize)
+        : work_path_{set_work_path(work_path)}, buffer_(Buffer(optimal_size)){};
+    ~Collector();
+
+    void flush_buffer();                       // Write buffer to file
+    void collect(db::Entry entry);             // Store key-value pair in memory or on disk
+    void load(lmdb::Table* table, Load load);  // Load collected entries in destination table
```
yperbasis, line 41: I would name call the function "transform" rather than "load". That would fit better with [ETL (Extract, transform, load)](https://en.wikipedia.org/wiki/Extract,_transform,_load)
```
void load(lmdb::Table* table, Transform transform);
```

db/silkworm/etl/collector.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef SILKWORM_ETL_COLLECTOR_H
+#define SILKWORM_ETL_COLLECTOR_H
+
+#include <silkworm/db/chaindb.hpp>
+#include <silkworm/etl/error.hpp>
+#include <silkworm/etl/buffer.hpp>
+#include <silkworm/etl/file_provider.hpp>
+
+namespace silkworm::etl {
```
yperbasis, line 25: Let's add a comment somewhere that ETL stands for Extract, Transform, Load:
https://en.wikipedia.org/wiki/Extract,_transform,_load

db/silkworm/etl/collector.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  private:
+    std::string set_work_path(const char* provided_work_path);
+
+    std::string work_path_;
+    std::vector<FileProvider> file_providers_;
+    Buffer buffer_;
+};
+// Load function for no processing
+std::vector<db::Entry> default_load(db::Entry entry);
```
yperbasis, line 51: I'd call this `identity_transform`

db/silkworm/etl/collector.cpp
```diff
@@ -0,0 +1,138 @@
...
+        if (fs::exists(path)) {
+            fs::remove_all(path);
+        }
+    }
+
+
+    void Collector::flush_buffer() {
+    if (buffer_.size()) {
+        buffer_.sort();
+        file_providers_.emplace_back(work_path_, (int)file_providers_.size());
```
yperbasis, line 42: I'd change the type of FileProvider's id to `size_t` instead of `int` to avoid this type cast.

db/silkworm/etl/buffer.hpp
```diff
@@ -0,0 +1,45 @@
...
+namespace silkworm::etl {
+
+// In ETL, a buffer must be used stores entries, sort them and write them to file
+class Buffer {
+  public:
+    Buffer(size_t optimal_size) : optimal_size_(optimal_size){};
+
+    void put(db::Entry& entry);    // Add a new entry to the buffer
+    void clear();                  // Free buffer's contents
+    bool overflows();              // Whether or not accounted size overflows optimal_size_ (i.e. time to flush)
```
yperbasis, line 34: Please mark as `const`

yperbasis (changes requested):

db/silkworm/etl/file_provider.hpp
AndreaLanfranchi: Right

db/silkworm/etl/file_provider.cpp
AndreaLanfranchi: Name overlapping (even though in different namespaces) is a bit frustrating.
I think I'll make it a public struct within Fileprovider class

yperbasis (approved):

PR #121, opened 2020-12-18
Optimize IntraBlockState::finalize_transaction

yperbasis:


core/silkworm/state/intra_block_state.cpp
```diff
@@ -255,7 +258,7 @@ void IntraBlockState::set_storage(const evmc::address& address, const evmc::byte
     if (prev == value) {
         return;
     }
-    storage_[address][key].current = value;
+    storage_[address].current[key] = value;
     journal_.push_back(std::make_unique<state::StorageChangeDelta>(address, key, prev));
```
AndreaLanfranchi, line 262: This causes a contructor of `unique_ptr` and then a move into vector.
Why not simply ...
```suggestion
   journal_.emplace_back(new state::StorageChangeDelta(address, key, prev));
```

core/silkworm/state/intra_block_state.cpp
```diff
@@ -312,10 +315,11 @@ void IntraBlockState::revert_to_snapshot(const IntraBlockState::Snapshot& snapsh
 
 void IntraBlockState::finalize_transaction() {
     for (auto& x : storage_) {
```
AndreaLanfranchi, line 317: More CXX17 style and makes clear we're iterating a map.
Also helpful in write_to_db etc.

```suggestion
    for (auto& [address, storage] : storage_) {
        for (const auto& [a, b] : storage.current) {
            storage.committed[a].original = b;
        }
        storage.current.clear();
    }
```

yperbasis: Cool, I didn't know about this feature

core/silkworm/state/intra_block_state.cpp
yperbasis: Done

core/silkworm/state/intra_block_state.cpp
```diff
@@ -307,7 +307,8 @@ void IntraBlockState::revert_to_snapshot(const IntraBlockState::Snapshot& snapsh
 }
 
 void IntraBlockState::finalize_transaction() {
-    for (auto& [_, storage] : storage_) {
+    for (auto& x : storage_) {
```
AndreaLanfranchi, line 310: You could have left the structured binding and simply add `(void)address` into the loop.
Nevertheless the cost is the same as in actual implementation `x` holds two references (first and second).

Just for reference. When GCC8 this also prevents the triggering of warning error
```
for ([[maybe_unused]] auto& [address, storage] : storage_) { 
 // Whatever
}
```

AndreaLanfranchi (approved):

core/silkworm/state/intra_block_state.cpp
```diff
@@ -307,8 +307,7 @@ void IntraBlockState::revert_to_snapshot(const IntraBlockState::Snapshot& snapsh
 }
 
 void IntraBlockState::finalize_transaction() {
-    for (auto& x : storage_) {
-        state::Storage& storage{x.second};
+    for ([[maybe_unused]] auto& [address, storage] : storage_) {
```
AndreaLanfranchi, line 310: Is only for GCC 8 ... we're building with GCC 7

PR #123, opened 2020-12-22
Split unit tests into core & db tests

yperbasis:


cmd/CMakeLists.txt
```diff
@@ -52,13 +52,13 @@ enable_testing()
...
+file(GLOB_RECURSE SILKWORM_CORE_TESTS CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/core/silkworm/*_test.cpp")
+add_executable(core_test unit_test.cpp ${SILKWORM_CORE_TESTS})
+target_link_libraries(core_test silkworm_core Catch2::Catch2)
 
-include(CTest)
-include(Catch)
-catch_discover_tests(unit_test)
+file(GLOB_RECURSE SILKWORM_DB_TESTS CONFIGURE_DEPENDS "${CMAKE_SOURCE_DIR}/db/silkworm/*_test.cpp")
+add_executable(db_test unit_test.cpp ${SILKWORM_DB_TESTS})
+target_link_libraries(db_test silkworm_db Catch2::Catch2)
```
AndreaLanfranchi, line 61: Just for clarity ... shouldn't we define also dependency qualifier as PUBLIC ?

AndreaLanfranchi (approved):

cmd/CMakeLists.txt
yperbasis: > Just for clarity ... shouldn't we define also dependency qualifier as PRIVATE ?

Well, db_test is not a library, so it probably doesn't matter whether its dependencies are public or private.

PR #128, opened 2021-01-04
Let transaction commit do fsync

AndreaLanfranchi:
See https://github.com/ledgerwatch/turbo-geth/pull/1433/files


yperbasis:
Let me run block execution with this change to check its performance impact.

yperbasis (approved):
I observed a small performance degradation of executing blocks up to 11.36M (17.3h -> 17.5h), but that's within the margin of measuring error. So it's probably worth it to get rid of MDB_NOSYNC for robustness' sake.

AndreaLanfranchi:
Ok. Thank you for measuring.

PR #129, opened 2021-01-04
Wasm compilation of core_test

yperbasis:
Part of work on Issue #111

### DIFF
--- a/core/silkworm/common/util.cpp
+++ b/core/silkworm/common/util.cpp
@@ -1,5 +1,5 @@
 /*
-   Copyright 2020 The Silkworm Authors
+   Copyright 2020-2021 The Silkworm Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/core/silkworm/common/util.cpp
+++ b/core/silkworm/common/util.cpp
@@ -24,12 +24,10 @@
 
 namespace silkworm {
 
-ByteView left_pad(ByteView view, size_t min_size) {
+ByteView left_pad(ByteView view, size_t min_size, Bytes& padded) {
     if (view.size() >= min_size) {
         return view;
     }
-
-    thread_local Bytes padded;
 
     if (padded.size() < min_size) {
         padded.resize(min_size);
@@ -46,12 +44,10 @@ ByteView left_pad(ByteView view, size_t min_size) {
     return padded;
 }
 
-ByteView right_pad(ByteView view, size_t min_size) {
+ByteView right_pad(ByteView view, size_t min_size, Bytes& padded) {
     if (view.size() >= min_size) {
         return view;
     }
-
-    thread_local Bytes padded;
 
     if (padded.size() < view.size()) {
         padded.resize(view.size());
--- a/core/silkworm/common/util.cpp
+++ b/core/silkworm/common/util.cpp
@@ -24,42 +24,52 @@
 
 namespace silkworm {
 
-ByteView left_pad(ByteView view, size_t min_size, Bytes& padded) {
+ByteView left_pad(ByteView view, size_t min_size, Bytes& buffer) {
     if (view.size() >= min_size) {
         return view;
     }
 
-    if (padded.size() < min_size) {
-        padded.resize(min_size);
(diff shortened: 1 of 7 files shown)


core/silkworm/common/util.cpp
```diff
@@ -24,13 +24,11 @@
 
 namespace silkworm {
 
-ByteView left_pad(ByteView view, size_t min_size) {
+ByteView left_pad(ByteView view, size_t min_size, Bytes& padded) {
```
AndreaLanfranchi, line 27: Maybe change name padded -> buffer (to be consistent with signature definition) ?

core/silkworm/common/util.cpp
```diff
@@ -46,13 +44,11 @@ ByteView left_pad(ByteView view, size_t min_size) {
     return padded;
 }
 
-ByteView right_pad(ByteView view, size_t min_size) {
+ByteView right_pad(ByteView view, size_t min_size, Bytes& padded) {
```
AndreaLanfranchi, line 47: Same as above

AndreaLanfranchi (changes requested):

core/silkworm/common/util.cpp
```diff
@@ -24,13 +24,11 @@
...
+ByteView left_pad(ByteView view, size_t min_size, Bytes& padded) {
     if (view.size() >= min_size) {
         return view;
     }
 
-    thread_local Bytes padded;
-
     if (padded.size() < min_size) {
         padded.resize(min_size);
     }
```
AndreaLanfranchi, line 34: Don't understand this conditional resize 
I would go for 
```suggestion
    padded.resize(view.size());                             // Resize buffer to size of input view
    size_t left_chars{min_size - view.size()};              // Compute number of chars to the left
    std::memmove(padded.data(), view.data(), view.size());  // Copy data
    padded.insert(0, left_chars, '\0');                     // Insert zeroes
    return padded;
```

AndreaLanfranchi (changes requested):

core/silkworm/common/util.cpp
AndreaLanfranchi: right_pad has also (imho) unnecessary double resize

core/silkworm/common/util.cpp
yperbasis: Renamed

core/silkworm/common/util.cpp
yperbasis: Renamed

core/silkworm/common/util.cpp
yperbasis: Since repeated padding is allowed, the input view might actually be a view of the buffer (or its substring), so I take care not to destroy any content of the buffer prior to the memmove. I've added comments.

AndreaLanfranchi: Got it now but I have yet a couple of questions

1st Looking at left_pad ...
```
    if (view.size() >= min_size) {
        return view;
    }

    if (buffer.size() < min_size) {
        buffer.resize(min_size);
    } else {
        // view & buffer might overlap in memory,
        // so we avoid shrinking the buffer prior to the memmove
    }

    assert(view.size() < min_size); // <--- This assertion can never fail due to the return on top
    size_t prefix_len{min_size - view.size()};
    [ ... ]
```

2nd. Why don't keep the original implementation keeping `buffer` as `static` (instead of thread_static) ? This way the implementation would require always a `memcpy` (impossible to have view as view-of-buffer) and original buffer gets never modified.

Sorry for pedancy

yperbasis: 1. Yep, that assert can never fail, it's just there for extra clarity
2. `static` would be unsafe in a multi-threaded environment (e.g. simultaneous calls to left_pad from different threads).

AndreaLanfranchi: 1. Ok ... I find "confusing" an assertion not needed. Anyway not a big deal
2. So I understand `core` does not support natively multi-threading but **can** be used as library in higher level multi-threaded code.

Goctha now.

AndreaLanfranchi (approved):

PR #132, opened 2021-01-09
Senders

AndreaLanfranchi:
- Implement db::stages for methods to get & set progresses (like in TG)
- Added some strength to ETL and some code clean-up
- ETL Load method overload to go through NO transformation
- Sender's recovery now leverages ETL


CMakeLists.txt
```diff
@@ -146,6 +146,7 @@ if(MSVC)
   add_compile_definitions(_SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING)
   add_compile_definitions(_SILENCE_CXX17_RESULT_OF_DEPRECATION_WARNING)
 
+  add_compile_options(/Ot)     # Optimize for speed
```
yperbasis, line 149: Is this for debug builds? /Ot is already implied for optimized builds

cmd/check_senders.cpp
```diff
@@ -243,147 +206,158 @@ void process_txs_for_signing(ChainConfig& config, uint64_t block_num, BlockBody&
...
     }
 }
 
-std::optional<uint64_t> get_highest_canonical_header(std::unique_ptr<lmdb::Table>& headers) {
-    size_t count{0};
-    lmdb::err_handler(headers->get_rcount(&count));
-    if (!count) {
-        return std::nullopt;
-    }
+uint64_t load_canonical_headers(std::unique_ptr<lmdb::Transaction>& txn, uint64_t from, uint64_t to, evmc::bytes32* out) {
```
yperbasis, line 239: Since this method simply uses txn and doesn't assume ownership, it's faster and simpler just to pass
`lmdb::Transaction*` or `lmdb::Transaction&` instead of `std::unique_ptr<lmdb::Transaction>&`.

Also, I'd change the output parameter to `std::vector<evmc::bytes32>& out` in order to avoid C-style memory management.

db/silkworm/db/access_layer.cpp
```diff
@@ -45,22 +45,29 @@ std::optional<BlockHeader> read_header(lmdb::Transaction& txn, uint64_t block_nu
...
 static std::vector<Transaction> read_transactions(lmdb::Transaction& txn, uint64_t base_id, uint64_t count) {
+    if (!count) {
+        return {};
+    }
+
+    auto table{txn.open(table::kEthTx)};
+    return read_transactions(table, base_id, count);
+}
+
+std::vector<Transaction> read_transactions(std::unique_ptr<lmdb::Table>& txn_table, uint64_t base_id, uint64_t count) {
```
yperbasis, line 56: Since this method doesn't assume ownership of txn_table, it's simpler and faster to just pass `lmdb::Table& txn_table ` instead of `std::unique_ptr<lmdb::Table>& txn_table`

db/silkworm/db/chaindb.cpp
```diff
@@ -270,6 +271,58 @@ size_t Transaction::get_id(void) { return mdb_txn_id(handle_); }
...
+    try {
+        std::unique_ptr<Table> tbl{nullptr};
+        if (!domain.name) {
+            tbl = this->open(MAIN_DBI);
+        } else {
+            tbl = this->open(domain, MDB_CREATE);
+        }
+        return tbl->put(mdb_key, mdb_data, 0);
+    } catch (const exception& ex) {
+        // We may fail opening the table
```
yperbasis, line 321: IMO the mixture of exceptions and error codes in chaindb.cpp is quite messy. My preference would be to consistently use exceptions. That's a lot a work though and should be done in a separate PR, if we decide to do it.

db/silkworm/db/chaindb.hpp
```diff
@@ -211,6 +211,12 @@ class Transaction {
 
     bool is_ro(void);  // Whether this transaction is readonly
 
+    // Quick lookup of data value provided a table domain and a key
+    std::optional<Bytes> get(const TableConfig& domain, MDB_val* mdb_key);
+
+    // Quick upsert of data value provided a table domain and a key
+    int put(const TableConfig& domain, MDB_val* mdb_key, MDB_val* mdb_data);
```
yperbasis, line 218: Where are these two methods used now?

db/silkworm/etl/collector.cpp
```diff
@@ -50,19 +49,23 @@ void Collector::collect(Entry& entry) {
...
-        for (const auto& entry : buffer_.get_entries()) {
-            auto entries{load_func(entry)};
-            for (const auto& entry2 : entries) {
-                table->put(entry2.key, entry2.value);
+        for (const auto& etl_entry : buffer_.get_entries()) {
+            auto trasformed_etl_entries{load_func(etl_entry)};
+            for (const auto& transformed_etl_entry : trasformed_etl_entries) {
+                MDB_val mdb_key{db::to_mdb_val(transformed_etl_entry.key)};
+                MDB_val mdb_data{db::to_mdb_val(transformed_etl_entry.value)};
+                lmdb::err_handler(table->put(&mdb_key, &mdb_data, flags));
```
yperbasis, line 61: How are these 3 lines better than `table->put(entry2.key, entry2.value);` ?

yperbasis (changes requested):
Please remove code duplication in Collector::load

db/silkworm/etl/collector.cpp
AndreaLanfranchi: They're not better but here I can handle flags.
The original signature for
```
void Table::put(ByteView key, ByteView data) {
    MDB_val key_val{db::to_mdb_val(key)};
    MDB_val data_val{db::to_mdb_val(data)};
    err_handler(put(&key_val, &data_val, 0));
}
```
does not implement flags which are hardcoded.
If you wish we could set the flags parameter there as defaulted to zero

db/silkworm/db/chaindb.hpp
AndreaLanfranchi: See db::stages::get/set_stage_progress

db/silkworm/db/chaindb.cpp
AndreaLanfranchi: In this specific case exceptions are forced by the fact they throw on constructor.

CMakeLists.txt
AndreaLanfranchi: A leftover from my tests.
Will remove

AndreaLanfranchi:
Not sure what you mean by code duplication in collector::load.
The second method is provided do avoid if branches when there is no need for transformation

yperbasis:
> Not sure what you mean by code duplication in collector::load.
> The second method is provided do avoid if branches when there is no need for transformation

What's the performance gain of avoiding branches? I suspect the gain is not worth code duplications.

db/silkworm/etl/collector.cpp
yperbasis: Right, let's have the flags parameter - it's probably be a net gain in code simplicity.

AndreaLanfranchi:
Sender's recovery full batch of 800M transactions +22 minutes with branch check for null transform function.

yperbasis:
> Sender's recovery full batch of 800M transactions +22 minutes with branch check for null transform function.

What's the total time? What the error in measurement? 

If the performance gain is real, I'd make the function into a template with `if constexpr` in order to avoid code duplication and preserve the gain in speed.

yperbasis:
@AndreaLanfranchi I really don't like the code duplication in Collector::load. I suggest to remove the code duplication for now, finish and merge the rest, and then revisit performance impact of Collector::load in isolation.

yperbasis (approved):

Giulio2002 (approved):

PR #133, opened 2021-01-09
use of env in ETL

Giulio2002:


AndreaLanfranchi (changes requested):
We need to keep usage transaction too for load. Actually we might use etl without an environment cause in tg integration we don't get passed an env rather only a handle to transaction

PR #134, opened 2021-01-13
Use StateBuffer instead of IntraBlockState for consensus tests

yperbasis:
(since IntraBlockState is supposed to be used inside a single block, not across blocks).

Also fix a corner case by introducing `StorageCreateDelta`.

### DIFF
--- a/cmd/consensus.cpp
+++ b/cmd/consensus.cpp
@@ -215,32 +215,42 @@ static void check_rlp_err(rlp::DecodingError err) {
 }
 
 // https://ethereum-tests.readthedocs.io/en/latest/test_types/blockchain_tests.html#pre-prestate-section
-void init_pre_state(const nlohmann::json& pre, IntraBlockState& state) {
+void init_pre_state(const nlohmann::json& pre, StateBuffer& state) {
     for (const auto& entry : pre.items()) {
         evmc::address address{to_address(*from_hex(entry.key()))};
-        const nlohmann::json& account{entry.value()};
-        Bytes balance_str{*from_hex(account["balance"].get<std::string>())};
+        const nlohmann::json& j{entry.value()};
+
+        Account account;
+        Bytes balance_str{*from_hex(j["balance"].get<std::string>())};
         auto [balance, err1]{rlp::read_uint256(balance_str, /*allow_leading_zeros=*/true)};
         check_rlp_err(err1);
-        state.set_balance(address, balance);
-        state.set_code(address, *from_hex(account["code"].get<std::string>()));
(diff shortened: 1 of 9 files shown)


cmd/consensus.cpp
```diff
@@ -215,32 +215,42 @@ static void check_rlp_err(rlp::DecodingError err) {
...
-void init_pre_state(const nlohmann::json& pre, IntraBlockState& state) {
+void init_pre_state(const nlohmann::json& pre, StateBuffer& state) {
     for (const auto& entry : pre.items()) {
         evmc::address address{to_address(*from_hex(entry.key()))};
-        const nlohmann::json& account{entry.value()};
-        Bytes balance_str{*from_hex(account["balance"].get<std::string>())};
+        const nlohmann::json& j{entry.value()};
+
+        Account account;
+        Bytes balance_str{*from_hex(j["balance"].get<std::string>())};
```
AndreaLanfranchi, line 224: Question ... from_hex returns std::optional<Bytes> ... are we sure of the sanity of the Json payload ?

cmd/consensus.cpp
```diff
@@ -284,56 +293,58 @@ Status run_block(const nlohmann::json& b, const ChainConfig& config, IntraBlockS
...
-        Bytes balance_str{*from_hex(account["balance"].get<std::string>())};
+        Bytes balance_str{*from_hex(j["balance"].get<std::string>())};
         auto [expected_balance, err1]{rlp::read_uint256(balance_str, /*allow_leading_zeros=*/true)};
         check_rlp_err(err1);
-        intx::uint256 actual_balance{state.get_balance(address)};
-        if (actual_balance != expected_balance) {
+        if (account->balance != expected_balance) {
             std::cout << "Balance mismatch for " << entry.key() << ":\n";
-            std::cout << to_string(actual_balance, 16) << " ≠ " << account["balance"] << "\n";
+            std::cout << to_string(account->balance, 16) << " ≠ " << j["balance"] << "\n";
```
AndreaLanfranchi, line 323: Please use ASCII chars as we get different display output on windows.
I'd go for `!=` instead of `≠`

AndreaLanfranchi (changes requested):

cmd/consensus.cpp
yperbasis: Done.

cmd/consensus.cpp
yperbasis: Right, I've replaced it with checked access to be on the safe side.

AndreaLanfranchi (approved):

PR #136, opened 2021-01-14
added blockhashes stage

Giulio2002:


cmd/blockhashes.cpp
```diff
@@ -0,0 +1,100 @@
+/*
+   Copyright 2020 The Silkworm Authors
```
yperbasis, line 2: Copyright year should be 2021 :)

yperbasis (approved):

cmd/blockhashes.cpp
Giulio2002: Done

PR #137, opened 2021-01-17
added stage tx lookups

Giulio2002:


cmd/check_tx_lookup.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <silkworm/common/util.hpp>
+#include <silkworm/chain/config.hpp>
+#include <iostream>
+
+using namespace silkworm;
+ 
+int main(int argc, char* argv[]) { 
+    namespace fs = boost::filesystem;
+    
+    CLI::App app{"Generates Blockhashes => BlockNumber mapping in database"};
```
AndreaLanfranchi, line 38: This line is the comment of BlockHashes stage

cmd/tx_lookup.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include <boost/endian/conversion.hpp>
+#include <silkworm/chain/config.hpp>
+#include <iostream>
+
+using namespace silkworm;
+ 
+int main(int argc, char* argv[]) { 
+    namespace fs = boost::filesystem;
+    
+    CLI::App app{"Generates Blockhashes => BlockNumber mapping in database"};
```
AndreaLanfranchi, line 37: This line is the comment of BlockHashes stage

cmd/check_tx_lookup.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+    fs::path datadir(db_path);
+    fs::path etl_path(datadir.parent_path() / fs::path("etl-temp"));
+    fs::create_directories(etl_path);
+    etl::Collector collector(etl_path.string().c_str(), /* flush size */ 512 * kMebi);
+
+    lmdb::DatabaseConfig db_config{db_path};
+    db_config.set_readonly(false); 
+    std::shared_ptr<lmdb::Environment> env{lmdb::get_env(db_config)};
+    std::unique_ptr<lmdb::Transaction> txn{env->begin_rw_transaction()};
```
AndreaLanfranchi, line 61: As you're only checking no need to set db in rw mode and also you can simply open a ro transaction

cmd/tx_lookup.cpp
```diff
@@ -0,0 +1,120 @@
...
+        SILKWORM_LOG(LogInfo) << "Entries to be inserted << " << total_entries << std::endl;
+        SILKWORM_LOG(LogInfo) << "Started Transaction Lookup Loading" << std::endl;
+        // If it was not empty before appending cannot happen
+        if (initial_block_number == 0) {
+            // Clear lookups just to be sure there is no garbage when append is enabled
+            txn->open(db::table::kTxLookup)->clear();
+            collector.load(tx_lookup_table.get(), nullptr, MDB_APPEND);
+        } else {
+            collector.load(tx_lookup_table.get(), nullptr, 0);
+        }
```
AndreaLanfranchi, line 110: MDB_APPEND is way faster when bulk loading.
I'd suggest the "else" case to delete all records from initial_block_number and eventually load collector data wiyh MDB_APPEND

AndreaLanfranchi (changes requested):
I'd rather have one executable only with CLI argument which specifies whether to do the write or the check

cmd/tx_lookup.cpp
Giulio2002: We are talking about hashes in key, we are not sure whether they are already sorted accordingly to the previous ones

cmd/blockhashes.cpp
```diff
@@ -89,7 +89,7 @@ int main(int argc, char* argv[]) {
             collector.load(blockhashes_table.get(), nullptr, 0);
         }
         // Update progress
-        db::stages::set_stage_progress(*txn, db::stages::kBlockHashesKey, current_block_number);
+        db::stages::set_stage_progress(*txn, db::stages::kBlockHashesKey, current_block_number-1);
```
AndreaLanfranchi, line 92: This is still not logically correct.
Picture this example where stage height is 10
```
        auto current_block_number{db::stages::get_stage_progress(*txn, db::stages::kBlockHashesKey)};
        auto initial_block_number = current_block_number;
```
After this lines current_block_number == initial_block_number == 10.
Here we have the first problem as the stage has already processed block 10 while the loop should begin from block 11
Then we jump to the scan loop
```
        for (int rc{header_table->seek(&key_mdb, &data_mdb)}; rc != MDB_NOTFOUND;
             rc = header_table->get_next(&key_mdb, &data_mdb)) {
            // Check if it's an header entry
            if (key_mdb.mv_size != 40) continue;
            Bytes key(static_cast<unsigned char*>(key_mdb.mv_data), key_mdb.mv_size);
            // We set the key to the hash of the header and the value to the block number
            etl::Entry entry{key.substr(8, 40), key.substr(0, 8)};
            collector.collect(entry);
            current_block_number++;
        }
```
If you look at the loop it will exit only if `rc == MDB_NOT_FOUND` which means any IO or txn error will make you loop indefinitely.
But imagine on first loop we get an MDB_NOT_FOUND: you still have current_block == 10 ... and you will end up
setting stage progress to
```
db::stages::set_stage_progress(*txn, db::stages::kBlockHashesKey, current_block_number - 1);
```
which will end up with stage set to 9.

cmd/check_tx_lookup.cpp
```diff
@@ -0,0 +1,108 @@
...
+                expected_block_number++;
+                continue;
+            }
+            Bytes transaction_key(8, '\0');
+            boost::endian::store_big_u64(transaction_key.data(), body.base_txn_id);
+            MDB_val tx_key_mdb{db::to_mdb_val(transaction_key)};
+            MDB_val tx_data_mdb;
+
+            uint64_t i{0};
+            for (int rc{transactions_table->seek_exact(&tx_key_mdb, &tx_data_mdb)}; rc != MDB_NOTFOUND && i < body.txn_count; rc = transactions_table->get_next(&tx_key_mdb, &tx_data_mdb), ++i) {
```
AndreaLanfranchi, line 90: This loop can finish only if you reach the limit of `body.txn_count` **or** `rc == MDB_NOT_FOUND`
You're ignoring every possible other db error which may cause you to insert wrong data into collector.

cmd/tx_lookup.cpp
```diff
@@ -0,0 +1,129 @@
...
+        // Extract
+        Bytes start(8, '\0');
+        boost::endian::store_big_u64(&start[0], initial_block_number);
+        uint64_t current_block_number{initial_block_number};
+        MDB_val key_mdb{db::to_mdb_val(start)};
+        MDB_val data_mdb;
+        SILKWORM_LOG(LogInfo) << "Started Transaction Lookup Extraction" << std::endl;
+        uint64_t total_entries = 0;
+        for (int rc{bodies_table->seek(&key_mdb, &data_mdb)}; rc != MDB_NOTFOUND; rc = bodies_table->get_next(&key_mdb, &data_mdb)) {
```
AndreaLanfranchi, line 83: Handle every possible db error

cmd/tx_lookup.cpp
```diff
@@ -0,0 +1,129 @@
...
+                current_block_number++;
+                continue;
+            }
+            Bytes transaction_key(8, '\0');
+            boost::endian::store_big_u64(transaction_key.data(), body.base_txn_id);
+            MDB_val tx_key_mdb{db::to_mdb_val(transaction_key)};
+            MDB_val tx_data_mdb{};
+
+            uint64_t i{0};
+            for (int rc{transactions_table->seek_exact(&tx_key_mdb, &tx_data_mdb)}; rc != MDB_NOTFOUND && i < body.txn_count; rc = transactions_table->get_next(&tx_key_mdb, &tx_data_mdb), ++i) {
```
AndreaLanfranchi, line 98: Handle every possible db error

AndreaLanfranchi (changes requested):

AndreaLanfranchi:
Also remember CLang formatting

cmd/blockhashes.cpp
Giulio2002: so, just setting it to current_block_number as i did before was correct, because it would end up in process the correct amount of blocks and resume from so that the correct blocks were processed, no?

AndreaLanfranchi: Not really ... as I showed in the discord channel with your previous revision if you had processed last block as 10 you'd end up setting stage height at 11

Giulio2002: you process block 10 previously and now you are about to process 11, i don't see whats wrong with it. you finished 10 in the previous cycle and you do not repeat it again.

AndreaLanfranchi: Stage height is not the block you're about to process: is the last block processed.

Giulio2002: ok now it should as yo said

cmd/blockhashes.cpp
```diff
@@ -60,7 +60,7 @@ int main(int argc, char* argv[]) {
     auto blockhashes_table{txn->open(db::table::kHeaderNumbers)};
     
     try {
-        auto current_block_number{db::stages::get_stage_progress(*txn, db::stages::kBlockHashesKey)};
+        auto current_block_number{db::stages::get_stage_progress(*txn, db::stages::kBlockHashesKey)-1};
```
AndreaLanfranchi, line 63: Why are you repeating the execution on the last already processed block ?
If last block processed for stage is 10 then you need to begin with 11 not with 9

AndreaLanfranchi (changes requested):

cmd/tx_lookup.cpp
```diff
@@ -96,6 +97,7 @@ int main(int argc, char* argv[]) {
 
             uint64_t i{0};
             for (int rc{transactions_table->seek_exact(&tx_key_mdb, &tx_data_mdb)}; rc != MDB_NOTFOUND && i < body.txn_count; rc = transactions_table->get_next(&tx_key_mdb, &tx_data_mdb), ++i) {
+                lmdb::err_handler(rc);
```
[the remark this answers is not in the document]
Giulio2002: when MDB_NOT_FOUND is found it break out already

cmd/tx_lookup.cpp
```diff
@@ -75,12 +75,13 @@ int main(int argc, char* argv[]) {
...
         Bytes start(8, '\0');
         boost::endian::store_big_u64(&start[0], initial_block_number);
-        uint64_t current_block_number{initial_block_number};
+        uint64_t current_block_number{initial_block_number-1};
         MDB_val key_mdb{db::to_mdb_val(start)};
         MDB_val data_mdb;
         SILKWORM_LOG(LogInfo) << "Started Transaction Lookup Extraction" << std::endl;
         uint64_t total_entries = 0;
         for (int rc{bodies_table->seek(&key_mdb, &data_mdb)}; rc != MDB_NOTFOUND; rc = bodies_table->get_next(&key_mdb, &data_mdb)) {
+            lmdb::err_handler(rc);
```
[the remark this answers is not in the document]
Giulio2002: when MDB_NOT_FOUND is found it break out already

cmd/blockhashes.cpp
```diff
@@ -70,6 +70,7 @@ int main(int argc, char* argv[]) {
         SILKWORM_LOG(LogInfo) << "Started BlockHashes Extraction" << std::endl;
 
         for (int rc{header_table->seek(&key_mdb, &data_mdb)}; rc != MDB_NOTFOUND; rc = header_table->get_next(&key_mdb, &data_mdb)) {
+            lmdb::err_handler(rc);
```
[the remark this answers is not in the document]
Giulio2002: when MDB_NOT_FOUND is found it break out already

cmd/blockhashes.cpp
Giulio2002: oops, i inverted it

cmd/tx_lookup.cpp
```diff
@@ -26,6 +26,13 @@
 
 using namespace silkworm;
 
+static Bytes compact(Bytes &b) {
+    for(size_t current_index = 0; current_index < b.size(); current_index++) {
+        if (b[current_index] != 0) return b.substr(current_index);
+    }
+    return Bytes(1, '\0');
+}
```
AndreaLanfranchi, line 34: I think this might have a problem if no leading zeroes are found: in such case you're returning a string of 1 char set to zero. (talking block height as is now is unlikely we fill a uint64 completely but ... the logic is wrong).
I would suggest to go for a more concise way as :

```
static Bytes compact(Bytes &b) {
    std::string::size_type offset{b.find_first_not_of((uint8_t)0)};
    if (offset != std::string::npos) {
        return b.substr(offset);
    }
    return b;
}
```

AndreaLanfranchi (changes requested):

AndreaLanfranchi (approved):
Ready to go for me.

PR #138, opened 2021-01-20
Wasm API

yperbasis:
Enables https://torquem.ch/eth_tests.html

### DIFF
--- a/core/silkworm/types/block.cpp
+++ b/core/silkworm/types/block.cpp
@@ -1,5 +1,5 @@
 /*
-   Copyright 2020 The Silkworm Authors
+   Copyright 2020-2021 The Silkworm Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


core/silkworm/types/block.cpp
```diff
@@ -34,6 +34,20 @@ bool operator==(const BlockBody& a, const BlockBody& b) {
...
 }
 
+void Block::recover_senders(const ChainConfig& config) {
+    uint64_t block_number{header.number};
+    bool homestead{config.has_homestead(block_number)};
+    bool spurious_dragon{config.has_spurious_dragon(block_number)};
+
+    for (Transaction& txn : transactions) {
+        if (spurious_dragon) {
+            txn.recover_sender(homestead, config.chain_id);
```
AndreaLanfranchi, line 44: There's no trapping for errors in txn.recover_senders so we don't know if it succeeds.

Question ... as we already have senders recovered and persisted on db why don't we get them from there ?

AndreaLanfranchi: Nevermind. Disregard it.
This is wasm ... no database

yperbasis: If senders are not recovered, block execution will fail with `kMissingSender`.

AndreaLanfranchi (approved):

PR #139, opened 2021-01-21
added more logging to ETL

Giulio2002:


db/silkworm/etl/collector.cpp
```diff
@@ -69,6 +70,8 @@ void Collector::load(silkworm::lmdb::Table* table, LoadFunc load_func, unsigned
         return;
     }
 
+    Logger::default_logger().set_local_timezone(true);  // for compatibility with TG logging
```
AndreaLanfranchi, line 73: Dont' override settings.
Let default logging date format be set by higher app

db/silkworm/etl/collector.cpp
```diff
@@ -96,9 +100,16 @@ void Collector::load(silkworm::lmdb::Table* table, LoadFunc load_func, unsigned
...
             for (const auto& transformed_etl_entry : load_func(etl_entry)) {
                 table->put(transformed_etl_entry.key, transformed_etl_entry.value, flags);
+                log_tracker += transformed_etl_entry.size();
             }
         } else {
             table->put(etl_entry.key, etl_entry.value, flags);
+            log_tracker += etl_entry.size();
+        }
+        // Oncde in a while updates the user on current progress
+        if (!op_name.empty() && log_tracker >= kLogInterval) {
```
AndreaLanfranchi, line 110: This is nice but actually does not give the perception of the progress we're at.
Maybe tracking **overall** size of collector and report a percentage of the load is better ?

AndreaLanfranchi (changes requested):

cmd/check_senders.cpp
```diff
@@ -646,7 +646,7 @@ int do_recover(app_options_t& options) {
         try {
             // Load collected data into Senders' table
             auto senders_table{lmdb_txn->open(db::table::kSenders)};
-            collector.load(senders_table.get(), nullptr, MDB_APPEND);
+            collector.load(senders_table.get(), nullptr, MDB_APPEND, "Senders");
```
AndreaLanfranchi, line 649: Is it necessary to define a label for connector ?
I believe a simple "Load progress %" is enough.
Nevertheless we might log an info line before invoking collector.load().

Giulio2002: We already do, in tx_lookup and blockhashes

AndreaLanfranchi (approved):

Issue #23, opened 2020-09-03
Ethereum Consensus Tests


yperbasis:
Run and pass [Ethereum Consensus Tests](https://github.com/ethereum/tests).

Current status:
- BasicTests (difficulty tests) - pass
- BlockchainTests / GeneralStateTests - pass
- BlockchainTests / InvalidBlocks - fail
- BlockchainTests / TransitionTests - pass
- BlockchainTests / ValidBlocks - pass
- TransactionTests - pass
- RLPTests - not integrated

PR #141, opened 2021-01-22
Support postStateHash consensus tests

yperbasis:
Part of Issue #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ cmd/execute -d <path-to-chaindata>
 
 
 ## Windows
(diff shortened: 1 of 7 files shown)


AndreaLanfranchi (approved):
Just a quick question: do you think is useful to define inherited classes as "final" if they cant' be further inherited from ?

yperbasis:
> Just a quick question: do you think is useful to define inherited classes as "final" if they cant' be further inherited from ?

I don't have a strong opinion on this.

PR #143, opened 2021-01-24
added unique ids to files

Giulio2002:


Giulio2002:
This one is ready, @AndreaLanfranchi

db/silkworm/etl/file_provider.cpp
```diff
@@ -23,7 +23,7 @@ namespace silkworm::etl {
 namespace fs = boost::filesystem;
 
 FileProvider::FileProvider(const std::string &working_path, size_t id) : id_{id} {
-    fs::path path{fs::path(working_path) / fs::path("tmp-" + std::to_string(id))};
+    fs::path path{fs::path(working_path) / fs::path("tmp-" + std::to_string(rand()))};
```
AndreaLanfranchi, line 26: I don't think `rand()` provides any guarantee of uniqueness.
See for instance here where you discard values already provided by rand
https://github.com/torquem-ch/silkworm/blob/master/db/silkworm/etl/collector_test.cpp#L30-L48

As we may use Collector in several instances (might happen also amongst different threads) I'd suggest to go for something which provides a more robust self-containing uniqueness. Maybe explore Boost::Uuid ... create a unique value on Collector contructor and replace literal `tmp` with that value.

AndreaLanfranchi (changes requested):

AndreaLanfranchi (approved):

PR #145, opened 2021-01-27
Partially implement block validation

yperbasis:
Most "InvalidBlocks" blockchain consensus tests pass now. Part of work on Issue #23.

Also fixes an RLP corner case.


core/silkworm/chain/validity.cpp
```diff
@@ -0,0 +1,127 @@
...
+                                                  parent->timestamp, parent_has_uncles, config)};
+    if (difficulty != header.difficulty) {
+        return ValidationError::kWrongDifficulty;
+    }
+
+    if (header.gas_used > header.gas_limit) {
+        return ValidationError::kGasAboveLimit;
+    }
+
+    uint64_t diff{header.gas_limit > parent->gas_limit ? header.gas_limit - parent->gas_limit
```
AndreaLanfranchi, line 48: To avoid confusion amongs diff and difficulty I'd rename this as `gas_delta` or `difference`

core/silkworm/chain/validity.cpp
```diff
@@ -0,0 +1,127 @@
...
+        return ValidationError::kInvalidGasLimit;
+    }
+
+    if (header.gas_limit < 5000) {
+        return ValidationError::kInvalidGasLimit;
+    }
+
+    if (header.timestamp <= parent->timestamp) {
+        return ValidationError::kInvalidTimestamp;
+    }
```
AndreaLanfranchi, line 60: I'd move these two checks for current header before verification of gas difference (might save a computation if gas below limit.

AndreaLanfranchi (approved):

PR #149, opened 2021-01-31
Logging, simplified

gcolvin:
Simplified by removing unused - and likely unneeded - functionality.  So just one logger, and easier to reset streams logged to.

Also wrapped logging in a mutex so that the logs of multiple threads interleave cleanly.

### DIFF
--- a/db/silkworm/common/log.hpp
+++ b/db/silkworm/common/log.hpp
@@ -18,6 +18,7 @@
 #define SILKWORM_COMMON_LOG_H_
 
 #include <silkworm/common/tee.hpp>
+#include <mutex>
 
 namespace silkworm {
 
--- a/db/silkworm/common/log.hpp
+++ b/db/silkworm/common/log.hpp
@@ -13,11 +13,27 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 */
+/*
+
+Usage:
+
+// log to default logger - cerr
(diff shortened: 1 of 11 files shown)


db/silkworm/common/log.hpp
```diff
@@ -17,48 +17,49 @@
...
+//
+#define SILKWORM_LOG_STREAMS(stream1_, stream2_) log_set_streams_((stream1_), (stream2_));
+
+// silence
+std::ostream& null_stream();
+
+//
+// Below are for access via macros ONLY :(
+//
+LogLevels log_verbosity_();
```
yperbasis, line 47: I'd wrap stuff that's not part of the API into `detail` namespace (like they do in Boost).

yperbasis (approved):

db/silkworm/common/log.hpp
gcolvin: Good idea.  Looks like this will our first use of that namespace.

AndreaLanfranchi (commented):
Sorry for being pedantic.
When we touch a file we should update copyright year (don't kill me)

db/silkworm/common/log.hpp
yperbasis: It's used in `db/silkworm/db/util.hpp` already :)

PR #151, opened 2021-02-04
Welcome to 2021!!!

gcolvin:
Previous merge of this branch left date changes behind.

### DIFF
--- a/db/silkworm/common/log.cpp
+++ b/db/silkworm/common/log.cpp
@@ -1,5 +1,5 @@
 /*
-   Copyright 2021 The Silkworm Authors
+   Copyright 2020 - 2021 The Silkworm Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/db/silkworm/common/log.cpp
+++ b/db/silkworm/common/log.cpp
@@ -1,5 +1,5 @@
 /*
-   Copyright 2020 The Silkworm Authors
+   Copyright 2021 The Silkworm Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


gcolvin (commented):
It's 2021.

yperbasis:
As far as I understand it, the copyright year should be the year of the first edition: [Copyright Notice Wiki](https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice_for_visually_perceptible_copies). I.e. we should leave it as 2020. Perhaps we can also add 2021, but 2020 should stay there. See also https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

I suggest that we simply have one copyright year per file, namely the year the file was created, and not bother updating it.

yperbasis (changes requested):
As far as I understand it, the copyright year should be the year of the first edition: [Copyright Notice Wiki](https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice_for_visually_perceptible_copies). I.e. we should leave it as 2020. Perhaps we can also add 2021, but 2020 should stay there. See also https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

I suggest that we simply have one copyright year per file, namely the year the file was created, and not bother updating it.

AndreaLanfranchi:
@yperbasis I believe having first (eventually only) year as the date of creation while adding as second year the one file was last modified makes no harm and gives an immediate perception of "stale" vs "active" files.
So having 2020 - 2021 should be ok ... imho

yperbasis:
Yeah, I guess 2020-2021 or 2020, 2021 is fine.

AndreaLanfranchi (approved):

yperbasis (approved):

PR #156, opened 2021-02-10
Blockchain reorgs

yperbasis:
This finishes Issue #23


AndreaLanfranchi (approved):
For better readability I'd rename `enum class ValidationError` in validity.hpp as `ValidationResult`
kOk in fact is not an error.

PR #159, opened 2021-02-11
Added History Index Stages (6 and 7)

Giulio2002:


cmd/history_index.cpp
```diff
@@ -165,9 +164,13 @@ int main(int argc, char* argv[]) {
             * cannot guarantee keys are sorted amongst different calls
             * of this stage
             */
+            auto target_table{txn->open(index_config, MDB_CREATE)};
             size_t target_table_rcount{0};
             lmdb::err_handler(target_table->get_rcount(&target_table_rcount));
             unsigned int db_flags{target_table_rcount ? 0u : (uint32_t)MDB_APPEND};
+            if (db_flags == MDB_APPEND) {
+                txn->open(index_config)->clear();
+            }
```
AndreaLanfranchi, line 173: What is the point of this ?
When db_flags == MDB_APPEND it means target_table is already empty.
Why re-empty it ?

AndreaLanfranchi (changes requested):

cmd/history_index.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        size_t allocated_space{0};
+        uint64_t block_number{0};
+        int rc{changeset_table->seek(&mdb_key, &mdb_data)};  // Sets cursor to nearest key greater equal than this
+        while (!rc) {                                     /* Loop as long as we have no errors*/
+            Bytes mdb_key_as_bytes{static_cast<uint8_t*>(mdb_key.mv_data), mdb_key.mv_size};
+            auto composite_key{get_history_key(mdb_key_as_bytes)};
+            if (bitmaps.find(composite_key) == bitmaps.end()) {
+                bitmaps.insert({composite_key, roaring::Roaring64Map()});
```
AndreaLanfranchi, line 122: Use bitmaps.emplace instead.
Your syntax create an entry and a new one by copy contructor.

AndreaLanfranchi (changes requested):

db/silkworm/etl/collector_test.cpp
```diff
@@ -71,26 +71,16 @@ void run_collector_test(LoadFunc load_func) {
     // Load data
     auto to{txn->open(db::table::kHeaderNumbers)};
     collector.load(to.get(), load_func);
-
-    // Check wheter load was performed as intended
```
yperbasis, line 75: Why remove this check?

Giulio2002: Instead of re-iterate over and over again now we do everything in load. so the check cannot be made.

db/silkworm/db/access_layer.cpp
```diff
@@ -204,17 +203,11 @@ static std::optional<ByteView> historical_storage(lmdb::Transaction& txn, const
                                                   uint64_t incarnation, const evmc::bytes32& location,
                                                   uint64_t block_number) {
     auto history_table{txn.open(table::kStorageHistory)};
-    std::optional<Entry> entry{history_table->seek(storage_history_key(address, location, block_number))};
+    std::optional<Entry> entry{history_table->seek(account_history_key(address, block_number))};
```
yperbasis, line 206: The changes to `historical_storage` don't look right. Storage history key is different from that of account history.

yperbasis (changes requested):

yperbasis:
`tests` submodule was changed while it shouldn't have – probably some merge fallout.

db/silkworm/db/bitmap.cpp
```diff
@@ -30,4 +30,38 @@ std::optional<uint64_t> seek(const roaring::Roaring64Map &bitmap, uint64_t n) {
     return std::nullopt;
 }
 
+roaring::Roaring64Map cut_left(roaring::Roaring64Map *bm, uint64_t size_limit) {
```
yperbasis, line 33: Please add a test for this function.

db/silkworm/db/bitmap.hpp
```diff
@@ -30,6 +30,9 @@ roaring::Roaring64Map read(ByteView serialized);
 // See TG SeekInBitmap64.
 std::optional<uint64_t> seek(const roaring::Roaring64Map &bitmap, uint64_t n);
 
+// Return cut bitmap of given size limit
+roaring::Roaring64Map cut_left(roaring::Roaring64Map *bitmap, uint64_t len);
```
yperbasis, line 34: If input bitmap doesn't change, then it should be marked as `const`. Also, it may not be `nullptr`, so it should be a reference rather than pointer. See [Google Style on Inputs and Outputs](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs).
In other words, the function signature should probably be
```
roaring::Roaring64Map cut_left(const roaring::Roaring64Map& bitmap, uint64_t len);
```

db/silkworm/db/access_layer.cpp
```diff
@@ -204,17 +203,11 @@ static std::optional<ByteView> historical_storage(lmdb::Transaction& txn, const
...
     if (!entry) {
         return std::nullopt;
     }
 
-    ByteView k{entry->key};
```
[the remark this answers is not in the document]
Giulio2002: ops, it was because of a failed merge

db/silkworm/db/bitmap.hpp
Giulio2002: it changes

yperbasis: OK, then the function doc should say how it changes.
A.t.m it's not clear from the doc what happens to the input and what is returned.
Also, it should be passed by reference, not by pointer, to highlight that it can't be `nullptr`.

cmd/history_index.cpp
```diff
@@ -0,0 +1,199 @@
...
+                    auto chunk_index{Bytes(entry.key.size() + 8, '\0')};
+
+                    std::memcpy(&chunk_index[0], &entry.key[0], entry.key.size());
+                    uint64_t suffix{bm.cardinality() == 0 ? (uint64_t) -1 : current_chunk.maximum()};
+                    boost::endian::store_big_u64(&chunk_index[entry.key.size()], suffix);
+                    Bytes current_chunk_bytes(current_chunk.getSizeInBytes(), '\0');
+                    current_chunk.write((char *) &current_chunk_bytes[0]);
+                    history_index_table->put(chunk_index, current_chunk_bytes, db_flags);
+                }
+            }, db_flags, 20);
```
yperbasis, line 183: `20` - magic constant again

yperbasis (changes requested):

yperbasis:
@AndreaLanfranchi are you fine with this PR? Have your concerns been addressed?

cmd/history_index.cpp
```diff
@@ -0,0 +1,189 @@
...
+                        // make chunk index
+                        Bytes chunk_index(entry.key.size() + 8, '\0');
+                        std::memcpy(&chunk_index[0], &entry.key[0], entry.key.size());
+                        uint64_t suffix{bm.cardinality() == 0 ? UINT64_MAX : current_chunk.maximum()};
+                        boost::endian::store_big_u64(&chunk_index[entry.key.size()], suffix);
+                        Bytes current_chunk_bytes(current_chunk.getSizeInBytes(), '\0');
+                        current_chunk.write((char *)&current_chunk_bytes[0]);
+                        history_index_table->put(chunk_index, current_chunk_bytes, db_flags);
+                    }
+                },
```
AndreaLanfranchi, line 172: If I am reading this correctly the transform lambda function is pretty redundant.
Specifically `chunk_index` is the copy of above created `last_chunk_index` (i.e. they both have data from `entry`) with the only difference that in every loop the last 8 bytes of chunk_index are either `UINT32_MAX` or `current_chunk.maximum()`.
Maybe refactoring a bit this lambda can save a lot of `memcpy`

cmd/history_index.cpp
```diff
@@ -0,0 +1,189 @@
...
+            std::string composite_key;
+            if (storage) {
+                char composite_key_array[kHashLength + kAddressLength];
+                std::memcpy(&composite_key_array[0], &static_cast<uint8_t *>(mdb_key.mv_data)[8], kAddressLength);
+                std::memcpy(&composite_key_array[kAddressLength], static_cast<uint8_t *>(mdb_data.mv_data),
+                            kHashLength);
+                composite_key = std::string(composite_key_array);
+            } else {
+                composite_key = std::string(static_cast<char *>(mdb_data.mv_data), kAddressLength);
+            }
```
AndreaLanfranchi, line 103: I'd evaluate if there are benefits in performance creating two variants of a lambda function in charge of bulding the key on behalf of `storage`. It should help removing an `if` branch for every block loop.
I agree this is not the heavy load bit.

AndreaLanfranchi (approved):

PR #162, opened 2021-02-18
Ability to use an exogenous EVM to run consensus tests

yperbasis:


core/silkworm/execution/evm.cpp
```diff
@@ -214,8 +214,23 @@ evmc::result EVM::call(const evmc_message& message) noexcept {
...
+    }
+
+    address_stack_.pop();
+
+    return evmc::result{res};
+}
+
+evmc_result EVM::execute_endogenously(const evmc_message& msg, ByteView code,
+                                      std::optional<evmc::bytes32> code_hash) noexcept {
+    const evmc_revision rev{revision()};
```
axic, line 233: Why not pass `rev` as parameter so that it is clear in `EVM::execute` that both use the same revision?

yperbasis: OK, done

AndreaLanfranchi (approved):

PR #165, opened 2021-02-19
Upgrade tests submodule to last rev before Berlin

chfast:
In evmone I'm using a newer consensus tests revision (the last commit before Berlin is introduced).
I noticed it causes Silkworm some issues probably in test loading. Not investigated yet.


yperbasis:
> In evmone I'm using a newer consensus tests revision (the last commit before Berlin is introduced).
> I noticed it causes Silkworm some issues probably in test loading. Not investigated yet.

The problem is that in this revision `BlockchainTests/ValidBlocks/bcStateTests/testOpcodes.json` already uses not yet supported Berlin network.

yperbasis:
Closing since PR #161 updates consensus tests to Berlin ([v7.0.1](https://github.com/ethereum/tests/releases/tag/v7.0.1))

PR #166, opened 2021-02-19
Count excluded tests as skipped

yperbasis:


AndreaLanfranchi (approved):

cmd/consensus.cpp
```diff
@@ -612,6 +612,12 @@ int main(int argc, char* argv[]) {
         }
     }
 
+    const RunResults kSkippedTest{
```
chfast, line 615: Should be `constexpr`.

PR #171, opened 2021-03-02
Refactor Sender's recovery

AndreaLanfranchi:
Adjust thread number on behalf of work load (i.e. don't spawn full number of threads if only few blocks are being processed)


db/silkworm/common/worker.hpp
```diff
@@ -30,30 +30,32 @@ class Worker {
...
     virtual ~Worker();
 
-    void start();                  // Start worker thread
+    void start(bool wait = true);  // Start worker thread (by default waits for status)
     void stop(bool wait = false);  // Stops worker thread (optionally wait for complete stop)
     void kick();                   // Kicks worker thread if waiting
 
     // Whether or not this worker/thread should stop
-    bool should_stop() { return state_.load() == WorkerState::kStopping; }
+    bool should_stop() { return state_.load(std::memory_order_relaxed) == WorkerState::kStopping; }
```
yperbasis, line 44: I'd avoid non-default (i.e. non `memory_order_seq_cst`) memory order since getting it right might be tricky, while performance gain is probably miniscule.

AndreaLanfranchi: Yes ... was a leftover. Removed everywhere else.

yperbasis (approved):
I skimmed over this change and on the surface it looks fine. However, in general I would prefer us to rely more on third-party libraries for multi-threading. There might already exist a good library with the `Worker` functionality. Perhaps @canepat can opine.

My reasoning is that, though writing multi-threaded code might be fun, it can be very tricky and require a great deal of time. Getting into nitty-gritty details of multi-threading might be not the best return on our time. Just IMHO.

AndreaLanfranchi:
Ok. Will leave open till @canepat drops his note.

canepat:
@yperbasis I agree, evaluating open-source multi-threading libraries for `Worker` replacement seems totally reasonable to me. AFAIK memory ordering programming in particular can hide subtle issues

@AndreaLanfranchi may I ask you what are the specific needs in implementing the `Worker` class? Please consider I've just taken a look at its usage without going deep

AndreaLanfranchi:
@canepat 
Worker class is a very simple wrapper around a threaded task. What it does is to provide handy methods to control the start/stop of the thread and, most important, handles the non-busy-wait for thread while in fact waiting for data to process (which saves the costy operation of spawning a new thread for each "atomic" work).

Basically every class which should do a work in a separate thread can easily derive from `Worker` and override the method `work()` where the task logic has to be placed. e.g 
Once the Worker is started the thread stays active and running for the whole life of the object or until the method `stop()` is called. The object can execute multiple tasks and even wait for data to process by the means of a condition variable.

In this implementation I have a `Recoverer` which waits for batch of transactions to process from the main thread, grinds them, and eventually signals have finished and the stays in idle wait for a new batch to process.

I reiterate ... worker is a basic wrapper around a thread with null footprint and handful methods. Is not meant to replace multithread/parallel libraries.

@yperbasis about exploring third party multithread libraries : I would like to see a clear statement here as I feel like directions seem to go back and forth. Do we have to stay **minimal** code and effort or embrace third party libs for every specific reason ? I feel like we already had a brainstorming about logging.

yperbasis:
@AndreaLanfranchi Sorry, I don't have a clear vision yet, it was just some thinking aloud. My experience with multi-threaded code is very limited.

Perhaps in the future we can think about using C++20 coroutines and libraries like Boost.Asio or cppcoro. Especially if coroutines in SilkRpc turn out to be a success. But that'll require discussion and more thought, of course, and is probably premature at this point.

Right now it's fine to have things like `Worker`, it seems to be straightforward and to do the job fine.

AndreaLanfranchi:
> Perhaps in the future we can think about using C++20 coroutines and libraries like Boost.Asio or cppcoro. Especially if coroutines in SilkRpc turn out to be a success. But that'll require discussion and more thought, of course, and is probably premature at this point.

Coroutines / threads == apples / oranges. Two completely separate contexts and models. With threads you have preemptive multitasking, with coroutines you basically start, suspend, resume tasks in a pretty serialized way where every thread can execute one coro only. Boost.Asio is basically a wrapper around call backs as each task is serialized either within the scope of Asio service or in different channels called strands but always within the Asio service (which has its own threads)

yperbasis:
> Not really. Boost's thread_pool is a little more sophisticated version of `std::vector<std::thread>`
> In this case RecoveryFarm not only handles (orchestrating them) all the recovery threads: it also implements the logic to prepare batches of work, dispatch them, and collect results from workers.

I see. Then in the future we might want to generalize `RecoveryFarm` and move it to silkworm_db or something, but for now it's fine.

yperbasis:
@AndreaLanfranchi Shall we merge the PR? It seems to be fine and we can always improve things later.

[Side note] My preference would be to have smaller PRs that we merge or abandon frequently, rather than big PRs which juxtapose too many things. <- Not necessarily about this PR, just a general thought.

AndreaLanfranchi:
> @AndreaLanfranchi Shall we merge the PR? It seems to be fine and we can always improve things later.
> 
Still polishing a couple of things.
While agree large PRs with tons of files changed are hard to track this one is pretty self-contained as is a refactor of a single process.

yperbasis (approved):

PR #176, opened 2021-03-06
Small fixes

AndreaLanfranchi:
Amend several C4456 warnings (hidden declaration)
More flexibility in parse_size to allow decimals eg "1.2TB"
Implement iequals (case insensitive string comparison) to keep core independent from Boost library
Correct an unreachable code path in Windows builds

### DIFF
--- a/core/silkworm/common/util.cpp
+++ b/core/silkworm/common/util.cpp
@@ -21,7 +21,6 @@
 #include <cstring>
 #include <iterator>
 #include <regex>
-#include <boost/algorithm/string/predicate.hpp>
 
 namespace silkworm {
 
@@ -155,6 +154,14 @@ std::optional<Bytes> from_hex(std::string_view hex) noexcept {
     return out;
 }
 
+inline bool case_insensitive_char_comparer(char a, char b) {
+    return(tolower(a) == tolower(b));
+}
+
+bool iequals(const std::string& a, const std::string& b) {
+    return (a.size() == b.size() && std::equal(a.begin(), a.end(), b.begin(), case_insensitive_char_comparer));
+}
+
 std::optional<uint64_t> parse_size(const std::string& sizestr) {
 
     if (sizestr.empty()) {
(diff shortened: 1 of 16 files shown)


core/silkworm/common/util.cpp
```diff
@@ -154,36 +154,69 @@ std::optional<Bytes> from_hex(std::string_view hex) noexcept {
...
+
 std::optional<uint64_t> parse_size(const std::string& sizestr) {
+
     if (sizestr.empty()) {
-        return 0;
+        return 0ull;
     }
 
-    static const std::regex pattern{"^([0-9]{1,})([\\ ]{0,})?(B|KB|MB|GB|TB)?$"};
+    static const std::regex pattern{"^(\\d*)(\\.\\d{1,3})?\\ *?(B|KB|MB|GB|TB)?$", std::regex_constants::icase};
```
yperbasis, line 171: It's probably better to use [C+11 raw strings](https://www.geeksforgeeks.org/raw-string-literal-c/) for regexes to avoid backslash doubling.

core/silkworm/common/util.cpp
```diff
@@ -154,36 +154,69 @@ std::optional<Bytes> from_hex(std::string_view hex) noexcept {
...
     std::smatch matches;
     if (!std::regex_search(sizestr, matches, pattern, std::regex_constants::match_default)) {
         return std::nullopt;
     };
 
-    uint64_t number{std::strtoull(matches[1].str().c_str(), nullptr, 10)};
+    std::string int_part, dec_part, suf_part;
+    uint64_t multiplier{1}; // Default for bytes (B|b)
+
+    for (size_t i = 1; i < matches.size(); i++) {
```
yperbasis, line 180: Do we need a loop here? I'd simply say 
```
int_part = matches[1].str();
if (!matches[2].str().empty()) {
  dec_part = matches[2].str().substr(1);
}
suf_part = matches[3].str();
```

core/silkworm/common/util.cpp
AndreaLanfranchi: Yep. You're right.

core/silkworm/common/util.cpp
AndreaLanfranchi: Right again. Was a leftover from another pattern. My bad

yperbasis (approved):

PR #180, opened 2021-03-11
Get actual chain config from db data

AndreaLanfranchi:
If chain is already (partially) sync'ed we should get `ChainConfig` from database.

### DIFF
--- a/db/silkworm/db/access_layer.cpp
+++ b/db/silkworm/db/access_layer.cpp
@@ -16,6 +16,7 @@
 
 #include "access_layer.hpp"
 
+#include <nlohmann/json.hpp>
 #include <boost/endian/conversion.hpp>
 #include <cassert>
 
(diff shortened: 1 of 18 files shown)


db/silkworm/db/access_layer.cpp
```diff
@@ -355,4 +356,77 @@ bool migration_happened(lmdb::Transaction& txn, const char* name) {
...
+    */
+
+    auto config_json{json::parse(config_value->c_str())};
+    ChainConfig config{};
+
+    if (config_json.contains("chainId")) {
+        config.chain_id = config_json["chainId"].get<uint64_t>();
+    }
+    if (config_json.contains("homesteadBlock")) {
+        config.homestead_block.emplace(config_json["homesteadBlock"].get<uint64_t>());
```
yperbasis, line 399: I'd simply say
```
config.homestead_block = config_json["homesteadBlock"].get<uint64_t>();
```
but that's a matter of taste.

yperbasis (approved):

db/silkworm/db/access_layer.cpp
AndreaLanfranchi: It throws if member does not exists

AndreaLanfranchi: Sorry I misread the comment ... yes maybe is a matter of taste. But as far as I understand emplace does not create a new std::optional and copies it to the member.

yperbasis: I don't understand. How
```
config.homestead_block = config_json["homesteadBlock"].get<uint64_t>();
```
is different it terms of throwing from
```
config.homestead_block.emplace(config_json["homesteadBlock"].get<uint64_t>());
```
?

AndreaLanfranchi: Read my above. Sorry I posted it late

yperbasis: No worries, both `emplace` and `=` are fine. I suspect the optimizer will produce the same code, but it's a minor thing.

PR #181, opened 2021-03-11
Use --chaindata intead of --datadir command line option for better TG compatibility

yperbasis:


AndreaLanfranchi (approved):
Yes chaindata is better

PR #185, opened 2021-03-15
Use Abseil rather than Robin Hood maps in core

yperbasis:
Rationale: Abseil is a well tested and supported library, and we already use it in silkworm_db. In my tests Abseil hash tables are slightly faster than Robin Hood ones for block execution. See also [benchmark conclusion](https://martin.ankerl.com/2019/04/01/hashmap-benchmarks-05-conclusion) by the author of Robin Hood map.


core/silkworm/common/hash_maps.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#else
+
+#include <absl/container/flat_hash_map.h>
+#include <absl/container/flat_hash_set.h>
+#include <absl/container/node_hash_map.h>
+
+#define SILKWORM_FLAT_HASH_MAP absl::flat_hash_map
+#define SILKWORM_FLAT_HASH_SET absl::flat_hash_set
+#define SILKWORM_NODE_HASH_MAP absl::node_hash_map
```
AndreaLanfranchi, line 56: I'm not a fan of shouty macros (besides preprocessor pollution).
Why not simply a template aliasing (which is actually what is meant) as of

```
template<class K, class V>
using flat_hash_map = typename absl::flat_hash_map<K, V>;
```

yperbasis: > I'm not a fan of shouty macros (besides preprocessor pollution).
> Why not simply a template aliasing (which is actually what is meant) as of
> 
> ```
> template<class K, class V>
> using flat_hash_map = typename absl::flat_hash_map<K, V>;
> ```

Yes, this is better, I agree.

AndreaLanfranchi (approved):
Good for me.
Thank you for accepting the suggestion.

PR #189, opened 2021-03-19
Split header bucket to headers, td and canonical

AndreaLanfranchi:
See https://github.com/ledgerwatch/turbo-geth/pull/1556

### DIFF
--- a/db/silkworm/db/tables.hpp
+++ b/db/silkworm/db/tables.hpp
@@ -30,8 +30,8 @@ constexpr lmdb::TableConfig kMAIN_DBI{nullptr};
 constexpr lmdb::TableConfig kAccountHistory{"hAT"};
 constexpr lmdb::TableConfig kBlockBodies{"b"};
 
-constexpr lmdb::TableConfig kHeadersHash{"canonical_headers", 0,
-                                         lmdb::cmp_fixed_len_key};  // block_num_u64 (BE) -> header_hash
+constexpr lmdb::TableConfig kCanonicalHashes{"canonical_headers", 0,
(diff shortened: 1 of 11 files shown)


db/silkworm/db/tables.hpp
```diff
@@ -29,7 +29,14 @@ namespace silkworm::db::table {
 constexpr lmdb::TableConfig kMAIN_DBI{nullptr};
 constexpr lmdb::TableConfig kAccountHistory{"hAT"};
 constexpr lmdb::TableConfig kBlockBodies{"b"};
-constexpr lmdb::TableConfig kBlockHeaders{"h"};
+
+constexpr lmdb::TableConfig kCanonicalHashes{"canonical_headers", 0,
```
yperbasis, line 33: These new tables don't use a custom key comparator in TG. Why should they in Silkworm?

AndreaLanfranchi: Cause it improves speed.
Default lexical comparator takes into account length of compared keys: but we know these tables have fixed key len so there is no point in checking both lengths (the custom comparator does only a memcmp).
Checked on bulk loading 12M canonical hashes with default comparator from LMDB 19.8 seconds while on custom comparator 19.5 seconds (average on 20 rounds). Is not that much but nevertheless a gain and does not break compatibility amongs tg/silkworm

yperbasis (approved):

db/silkworm/db/tables.hpp
yperbasis: OK, fair enough.

PR #197, opened 2021-03-25
Add Stage Hashstate(Stage 6)

Giulio2002:
Issue #178


db/silkworm/db/chaindb.cpp
```diff
@@ -524,6 +524,11 @@ int Table::get_prev_dup(MDB_val* key, MDB_val* data) { return get(key, data, MDB
 int Table::get_next(MDB_val* key, MDB_val* data) { return get(key, data, MDB_NEXT); }
 int Table::get_next_dup(MDB_val* key, MDB_val* data) { return get(key, data, MDB_NEXT_DUP); }
 int Table::get_next_nodup(MDB_val* key, MDB_val* data) { return get(key, data, MDB_NEXT_NODUP); }
+int Table::get_next_dup_unrestricted(MDB_val* key, MDB_val* data) {
```
yperbasis, line 527: Doesn't `get_next` do the same?

Giulio2002: ops, sorry was just experimenting with it a little bit

cmd/hashstate.cpp
```diff
@@ -0,0 +1,311 @@
...
+    }
+    if (field_set & 2) {
+        pos += encoded[pos++];
+    }
+    if (field_set & 4) {
+        // Incarnation has been found.
+        uint8_t len = encoded[pos++];
+        auto [incarnation, err]{rlp::read_uint64(encoded.substr(pos, len))};
+        if (err != rlp::DecodingResult::kOk) {
+            throw err;
```
AndreaLanfranchi, line 157: err is an enum ... not an exception.

db/silkworm/etl/buffer.cpp
```diff
@@ -25,7 +25,13 @@ void Buffer::put(Entry& entry) {
...
 void Buffer::sort() {
     std::sort(entries_.begin(), entries_.end(),
-              [](const Entry& a, const Entry& b) { return a.key.compare(b.key) < 0; });
+              [](const Entry& a, const Entry& b) { 
+                auto diff{a.key.compare(b.key)};
+                if (diff == 0) {
+                    return a.value.compare(b.value) < 0;
+                }
+                return diff < 0; 
+                });
```
AndreaLanfranchi, line 34: The returns are not correct imho. They should freely return -1 (for a < b) 0 (for equality) 1 (for a > b)
Here instead you're returning a boolean. I'd go for a much simpler
```suggestion
              [](const Entry& a, const Entry& b) { 
                auto diff{a.key.compare(b.key)};
                return diff ? diff : a.value.compare(b.value);
```

db/silkworm/etl/collector.cpp
```diff
@@ -104,7 +104,11 @@ void Collector::load(silkworm::lmdb::Table* table, LoadFunc load_func, unsigned
 
     // Define a priority queue based on smallest available key
     auto key_comparer = [](std::pair<Entry, int> left, std::pair<Entry, int> right) {
-        return left.first.key.compare(right.first.key) > 0;
+        auto diff{left.first.key.compare(right.first.key)};
+        if (diff == 0) {
+            return left.first.value.compare(right.first.value) > 0;
+        }
+        return diff > 0;
```
AndreaLanfranchi, line 111: Same as above comment

AndreaLanfranchi (changes requested):

db/silkworm/etl/buffer.cpp
Giulio2002: MDB_APPENDUP seems to execute without problems so this must right, otherwise  It would throw a key already exist error

cmd/hashstate.cpp
Giulio2002: that's how those exceptions are handled everywhere in silkworm, check access_db_layer.cpp in function check_err_rlp, what being throw is err.

AndreaLanfranchi: cc @yperbasis for thoughts
Agree and that's why we shouldn't do that imho.
A thrown exception can be catched by a try-catch block by inheritance through its propagation in the stack.
This example will leave the exception unhandled
```
enum class [[nodiscard]] DecodingResult{
    kOk = 0,
    kOverflow,
    kLeadingZero,
    kInputTooShort,
    kNonCanonicalSingleByte,
    kNonCanonicalSize,
    kUnexpectedLength,
    kUnexpectedString,
    kUnexpectedList,
    kListLengthMismatch,
    kUnsupportedEip2718Type,
};

try {
   throw kUnexpectedLength;
} catch (const std::exception& ex) {
  // Whatever to recover or log the error
}
```
cause `kUnexpectedLength` does not inherit from std::exception.

It could only work if the catch is modified as follows
```
try {
   throw kUnexpectedLength;
} catch (...) {
  // Whatever to recover or log the error
  // but we actually don't know what exactly happened
}
```
or 
```
try {
   throw kUnexpectedLength;
} catch (const DecodingResult& ex) {
  // Whatever to recover or log the error
  // but we actually need to catch a specific type
}
```

My suggestion is to always have thrown exception to inherit from std::exception cause in higher level calls we don't know which kind of exception might be thrown at lower levels (unless perfectly documented).

db/silkworm/etl/buffer.cpp
AndreaLanfranchi: You're right on this: sorry my fault.
Nevertheless the code could be much simpler by :

```suggestion
              [](const Entry& a, const Entry& b) { 
                return (a.key.compare(b.key) < 0) || (a.value.compare(b.value) < 0);
                }
```

AndreaLanfranchi: Disregard the above ... you're right. I've mixed up. 
std::compare returns three possible values while the compare function must be binary.

AndreaLanfranchi: This is what I meant ... I should go resting.
```suggestion
              [](const Entry& a, const Entry& b) { 
                auto diff{a.key.compare(b.key) || a.value.compare(b.value)};
                return diff < 0; 
                });
```

cmd/hashstate.cpp
yperbasis: @AndreaLanfranchi Yes, I agree, we should probably create something like `DecodingException` and throw it instead of `DecodingResult`. That was my bad example, sorry. I was lazy and just threw error codes in a couple of places.
I've created Issue #203 to exceptions rather than error codes. I suggest to proceed with the PR as is and rectify the issue later.

yperbasis (changes requested):
`tests` submodules shouldn't have been updated.

db/silkworm/etl/buffer.cpp
Giulio2002: the code you suggested does not work

yperbasis (approved):

db/silkworm/etl/buffer.cpp
yperbasis: @AndreaLanfranchi IMO the code by @Giulio2002 is fine.
We can also say
```
return a.key < b.key || (a.key == b.key && a.value < b.value);
```
but that might be slower than the original code.

yperbasis: N.B. When/if we switch to C++20, its [spaceship operator](https://en.cppreference.com/w/cpp/language/default_comparisons) will hopefully do the right thing out of the box.

PR #206, opened 2021-03-29
Magic enum

AndreaLanfranchi:
Introduce [magic-enum header](https://github.com/Neargye/magic_enum) and upgrade GCC to 9.x due to [header's limitations](https://github.com/Neargye/magic_enum/blob/master/doc/limitations.md).

Magic enum provides a useful static reflection over enums.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,17 +61,6 @@ jobs:
     machine:
       image: ubuntu-2004:202010-01
     steps:
-      - run:
-          name: "Switch to GCC 9"
-          command: |
-            sudo apt install software-properties-common
-            sudo add-apt-repository ppa:ubuntu-toolchain-r/test
-            sudo apt update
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -54,20 +54,22 @@ commands:
...
+  linux-gcc-9:
     environment:
       BUILD_TYPE: Debug
       CLANG_COVERAGE: OFF
     machine:
       image: ubuntu-2004:202010-01
     steps:
       - run:
-          name: "Switch to GCC 7"
+          name: "Switch to GCC 9"
```
yperbasis, line 65: I believe Ubuntu 20.04 comes with gcc 9 by default, so we can probably skip this step entirely.

AndreaLanfranchi: Ouch ... thought it was a 18.04 LTS. My bad

yperbasis (approved):

PR #207, opened 2021-03-29
Better Buffer

Giulio2002:


AndreaLanfranchi (commented):
I'm not personally in favor of this PR while still have to evaluate performances. Here the reasons

* Adds a discrete amount of boilerplate code
* Voids the usage of std::iterator over vectors
* At array capacity the size of the array if simply doubled
* Doubles boilerplate code for checking both length of array and size of buffer (wrt overrflow).

I understand this may produce better performance results but, in my **very very humble** opinion, is only due to the fact original buffer lacks a `reserve` on constructor which basically prevents several reallocations on multiple inserts.

I personally believe `Buffer` class could be removed in favor of `std::unique_ptr<Entry>` whcih could see the ownership moved amongst `collector` and `fileprovider` on overflow.

I reiterate: **this is not a negative review**. Only a comment.

Giulio2002:
> I'm not personally in favor of this PR while still have to evaluate performances. Here the reasons
> 
> * Adds a discrete amount of boilerplate code
> * Voids the usage of std::iterator over vectors
> * At array capacity the size of the array if simply doubled
> * Doubles boilerplate code for checking both length of array and size of buffer (wrt overrflow).
> 
> I understand this may produce better performance results but, in my **very very humble** opinion, is only due to the fact original buffer lacks a `reserve` on constructor which basically prevents several reallocations on multiple inserts.
> 
> I personally believe `Buffer` class could be removed in favor of `std::unique_ptr<Entry>` whcih could see the ownership moved amongst `collector` and `fileprovider` on overflow.
> 
> I reiterate: **this is not a negative review**. Only a comment.

Thank you @AndreaLanfranchi , I can try replacing swap with clear in `reset`and see if any performance is lost or gained. However, i think that it depends whether we prefer code integrity over performance as a whole. i will test what you suggested and let you know.

yperbasis (changes requested):
tests submodule should not change

yperbasis (commented):
I agree with @AndreaLanfranchi. Let's try the following experiment. What happens if you don't change anything, only call `entries_.reserve(kInitialBufferCap)` in `Buffer` constructors. As Andrea suggests, it might achieve the same performance affect w/o the boilerplate.

yperbasis:
As discussed, PR #212 does essentially the same optimization.

PR #208, opened 2021-03-29
std::filesystem instead of boost::filesystem

yperbasis:


db/silkworm/common/temp_dir.cpp
```diff
@@ -19,20 +19,23 @@
...
         "0123456789"
         "abcdefghijklmnopqrstuvwxyz"};
 
+    // don't count the null terminator
+    static constexpr size_t kNumberOfCharacters{sizeof(kAlphaNum) - 1};
+
     std::random_device rd;
     std::default_random_engine engine(rd());
-    std::uniform_int_distribution<size_t> uniform_dist(0, sizeof(alphanum) - 1);
+    std::uniform_int_distribution<size_t> uniform_dist(0, kNumberOfCharacters - 1);
```
AndreaLanfranchi: Am I wrong saying this removes null terminator twice ?

yperbasis: > Am I wrong saying this removes null terminator twice ?

Nope, the upper limit in uniform_int_distribution is inclusive, that is it yields random numbers `<= kNumberOfCharacters - 1`

AndreaLanfranchi: 👍

AndreaLanfranchi (approved):

Issue #203, opened 2021-03-29
Don't throw error codes, throw exceptions


yperbasis:
cc @yperbasis for thoughts
Agree and that's why we shouldn't do that imho.
A thrown exception can be catched by a try-catch block by inheritance through its propagation in the stack.
This example will leave the exception unhandled
```
enum class [[nodiscard]] DecodingResult{
    kOk = 0,
    kOverflow,
    kLeadingZero,
    kInputTooShort,
    kNonCanonicalSingleByte,
    kNonCanonicalSize,
    kUnexpectedLength,
    kUnexpectedString,
    kUnexpectedList,
    kListLengthMismatch,
    kUnsupportedEip2718Type,
};

try {
   throw kUnexpectedLength;
} catch (const std::exception& ex) {
  // Whatever to recover or log the error
}
```
cause `kUnexpectedLength` does not inherit from std::exception.

It could only work if the catch is modified as follows
```
try {
   throw kUnexpectedLength;
} catch (...) {
  // Whatever to recover or log the error
  // but we actually don't know what exactly happened
}
```
or 
```
try {
   throw kUnexpectedLength;
} catch (const DecodingResult& ex) {
  // Whatever to recover or log the error
  // but we actually need to catch a specific type
}
```

My suggestion is to always have thrown exception to inherit from std::exception cause in higher level calls we don't know which kind of exception might be thrown at lower levels (unless perfectly documented).

_Originally posted by @AndreaLanfranchi in https://github.com/torquem-ch/silkworm/pull/197#r602887209_

yperbasis:
I agree, we should probably create something like `DecodingException` and throw it instead of `DecodingResult`

PR #209, opened 2021-03-29
Enum exceptions

AndreaLanfranchi:
* Always prefer `enum class` vs `enum` to avoid clashes
* Throw exceptions instead of enum

### DIFF
--- a/cmd/hashstate.cpp
+++ b/cmd/hashstate.cpp
@@ -21,6 +21,7 @@
 #include <boost/filesystem.hpp>
 
 #include <silkworm/common/log.hpp>
+#include <silkworm/common/magic_enum.hpp>
 #include <silkworm/db/access_layer.hpp>
 #include <silkworm/db/stages.hpp>
 #include <silkworm/db/tables.hpp>
@@ -158,7 +159,7 @@ uint64_t extract_incarnation(ByteView encoded) {
         uint8_t len = encoded[pos++];
         auto [incarnation, err]{rlp::read_uint64(encoded.substr(pos, len))};
         if (err != rlp::DecodingResult::kOk) {
-            throw err;
+            throw std::runtime_error("Decoding error " + std::string(magic_enum::enum_name(err)));
         }
         return incarnation;
     }
--- a/cmd/hashstate.cpp
+++ b/cmd/hashstate.cpp
@@ -14,10 +14,12 @@
    limitations under the License.
 */
 
+#include <iostream>
+
 #include <CLI/CLI.hpp>
 #include <boost/endian/conversion.hpp>
 #include <boost/filesystem.hpp>
-#include <iostream>
+
 #include <silkworm/common/log.hpp>
 #include <silkworm/db/access_layer.hpp>
 #include <silkworm/db/stages.hpp>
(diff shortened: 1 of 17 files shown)


tg_api/silkworm_tg_api.h
```diff
@@ -40,7 +40,7 @@
 extern "C" {
 #endif
 
-enum SilkwormStatusCode {
+enum class SilkwormStatusCode {
```
yperbasis, line 43: This shouldn't be `enum class` because it's C, not C++.

db/silkworm/common/log.hpp
```diff
@@ -36,7 +36,7 @@ namespace silkworm {
 #define SILKWORM_LOG_VERBOSITY(level_) (silkworm::log_verbosity_ = (level_))
 
 // available verbosity levels
-enum LogLevels { LogTrace, LogDebug, LogInfo, LogWarn, LogError, LogCritical, LogNone };
+enum class LogLevels { LogTrace, LogDebug, LogInfo, LogWarn, LogError, LogCritical, LogNone };
```
yperbasis, line 39: I'd rename to `LogLevel` and remove the `Log` prefix from the values so that we could say e.g.
```
SILKWORM_LOG(LogLevel::Info) << "ETL Load Progress "
```
rather than
```
SILKWORM_LOG(LogLevels::LogInfo) << "ETL Load Progress "
```

yperbasis (changes requested):

yperbasis:
Related to Issue #203

cmd/hashstate.cpp
```diff
@@ -157,31 +159,31 @@ uint64_t extract_incarnation(ByteView encoded) {
         uint8_t len = encoded[pos++];
         auto [incarnation, err]{rlp::read_uint64(encoded.substr(pos, len))};
         if (err != rlp::DecodingResult::kOk) {
-            throw err; 
+            throw std::runtime_error("Decoding error " + std::string(magic_enum::enum_name(err)));
```
yperbasis, line 162: Perhaps we can create a dedicated `DecodingError` exception.

tg_api/silkworm_tg_api.h
AndreaLanfranchi: Yes right.

db/silkworm/common/log.hpp
AndreaLanfranchi: Sounds better. Agree

cmd/hashstate.cpp
AndreaLanfranchi: I was thinking the same thing but realized rlp decoding is part of core.
Need to check if magic_enum builds properly for wasm

tg_api/silkworm_tg_api.cpp
```diff
@@ -91,31 +92,33 @@ SILKWORM_EXPORT SilkwormStatusCode silkworm_execute_blocks(MDB_txn* mdb_txn, uin
...
-        SILKWORM_LOG(LogError) << "LMDB error " << e.what() << std::endl;
-        return kSilkwormLmdbError;
+        SILKWORM_LOG(LogLevel::Error) << "LMDB error " << e.what() << std::endl;
+        return SilkwormStatusCode::kSilkwormLmdbError;
     } catch (const db::MissingSenders&) {
-        SILKWORM_LOG(LogError) << "Missing or incorrect senders at block " << block_num << std::endl;
-        return kSilkwormMissingSenders;
+        SILKWORM_LOG(LogLevel::Error) << "Missing or incorrect senders at block " << block_num << std::endl;
+        return SilkwormStatusCode::kSilkwormMissingSenders;
     } catch (rlp::DecodingResult e) {
```
yperbasis, line 116: Should catch `decode_error` here

yperbasis (changes requested):

db/silkworm/db/access_layer.hpp
```diff
@@ -24,11 +24,38 @@
...
 #include <silkworm/chain/config.hpp>
+#include <silkworm/common/magic_enum.hpp>
 #include <silkworm/db/chaindb.hpp>
 #include <silkworm/db/util.hpp>
 #include <silkworm/types/account.hpp>
 #include <silkworm/types/block.hpp>
 
+namespace silkworm::rlp {
+
+class decode_error : public std::exception {
```
yperbasis, line 35: I'd name it `DecodingError` for consistency with [Google's style guide](https://google.github.io/styleguide/cppguide.html#Type_Names)

yperbasis (approved):

PR #212, opened 2021-03-30
Optimize etl::Buffer

yperbasis:
Higher-level alternative to PR #207. Kudos to @Giulio2002

### DIFF
--- a/cmake/Hunter/config.cmake
+++ b/cmake/Hunter/config.cmake
@@ -30,3 +30,10 @@ hunter_config(
   URL https://github.com/chfast/intx/archive/v0.5.1.tar.gz
   SHA1 743c46a82750143bd302a4394b7008a2112fc97b
 )
+
+hunter_config(
+    Microsoft.GSL
+    VERSION 3.1.0
+    URL https://github.com/microsoft/GSL/archive/v3.1.0.tar.gz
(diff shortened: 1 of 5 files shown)


Giulio2002:
Ok, so there have been an increase in performance compared of normal silkworm. This branch takes 57:56 minutes to complete vs 1:03 hours of normal silkworm. However, my branch is still taking 51:02 minutes. so its choosing between 7% improvement(this branch) with no boilerplate code or 18% improvement with boilerplate code. it all comes down to which trade-of we prefer (unless it can be optimized even further). We could also make a custom container to wrap the boilerplated code, this could also allow us to use it somewhere else without everything looking ugly. I also feel like stage 5 could be optimized to a certain extent by doing something like that. but that's just me wanting to experiment :).

yperbasis:
@Giulio2002 I've compared performance between the two versions on a Google Cloud machine and it is the same (the difference is less than 1%). This this be reasonably accurate as no other processes were running on the machine.
This PR: [optimize_etlb816bca.txt](https://github.com/torquem-ch/silkworm/files/6241921/optimize_etlb816bca.txt)
PR #207: [etl_buffer_experiment0ff087f.txt](https://github.com/torquem-ch/silkworm/files/6241922/etl_buffer_experiment0ff087f.txt)

PR #215, opened 2021-04-04
Initialize with genesis file

Giulio2002:


db/silkworm/common/chain_genesis.hpp
```diff
@@ -0,0 +1,9 @@
...
...c17af98dea4\":{\"balance\":\"1000000000000000000000\"},\"ac3900298dd14d7cc96d4abb428da1bae213ffed\":{\"balance\":\"24730250000000000000000\"},\"944f07b96f90c5f0d7c0c580533149f3f585a078\":{\"balance\":\"74000000000000000000\"},\"232c6d03b5b6e6711efff190e49c28eef36c82b0\":{\"balance\":\"1337000000000000000000\"},\"c87c77e3c24adecdcd1038a38b56e18dead3b702\":{\"balance\":\"8800000000000000000000\"},\"c4b6e5f09cc1b90df07803ce3d4d13766a9c46f4\":{\"balance\":\"6000000000000000000000\"},\"d44334b4e23a169a0c16bd21e866bba52d970587\":{\"balance\":\"2600000000000000000000\"},\"7757a4b9cc3d0247ccaaeb9909a0e56e1dd6dcc2\":{\"balance\":\"20000000000000000000\"},\"cf694081c76d18c64ca71382be5cd63b3cb476f8\":{\"balance\":\"1000000000000000000000\"},\"133e4f15e1e39c53435930aaedf3e0fe56fde843\":{\"balance\":\"20000000000000000000\"},\"f067fb10dfb293e998abe564c055e3348f9fbf1e\":{\"balance\":\"2000000000000000000000\"},\"94449c01b32a7fa55af8104f42cdd844aa8cbc40\":{\"balance\":\"16548000000000000000000\"},\"0e2094ac1654a46ba1c4d3a40bb8c17da7f39688\":{\"balance\":\"358000000000000000000\"},\"738ca94db7ce8be1c3056cd6988eb376359f3353\":{\"balance\":\"25500000000000000000000\"},\"0cfb172335b16c87d519cd1475530
```
yperbasis, line 4: Why wrap JSON as C++ strings? IMHO separate JSON files for mainnet, Rinkeby, etc would be better for readability.

Giulio2002: if we want to initialize a mainnet node on a future silkworm client, we should have to pre-made configuration without the user specifying a path (like it's made on tg, geth, etc...)

Giulio2002: just my two cents on the matter, i just believe that the main genesis should be native and not be inputted by the user

yperbasis: OK, fair enough. Could you please then use [raw literals](https://www.geeksforgeeks.org/raw-string-literal-c/) and formatted JSON so that it looks something like
```
std::string kMainnetGenesis = R"({"alloc":
    "3282791d6fd713f1e94f4bfd565eaa78b3a0599d":{"balance":"1337000000000000000000"},
    "17961d633bcf20a7b029a7d94b7df4da2ec5427f":{"balance":"229427000000000000000"},
    ...
})";
```
instead of extremely long lines.

yperbasis: Also, you can create nlohmann::json directly:
https://github.com/nlohmann/json#json-as-first-class-data-type
They give an example
```
// instead, you could also write (which looks very similar to the JSON above)
json j2 = {
  {"pi", 3.141},
  {"happy", true},
  {"name", "Niels"},
  {"nothing", nullptr},
  {"answer", {
    {"everything", 42}
  }},
  {"list", {1, 0, 2}},
  {"object", {
    {"currency", "USD"},
    {"value", 42.99}
  }}
};
```

Giulio2002: sure

Giulio2002: I think that the last option would be too time expensive (it's a lot of allocs) (26771 lines)

Giulio2002: now i made it more readable

Giulio2002:
now it's ready for being reviewed

yperbasis:
MSVC doesn‘t support such long strings unfortunately. CI Windows build is failing with [C2026](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2026). As an alternative to huge string literals, let‘s try to construct json directly, i.e. something like
```
nlohmann::json kMainnetGenesis = {
    "alloc": {
      "3282791d6fd713f1e94f4bfd565eaa78b3a0599d": {
        "balance": "1337000000000000000000"
      },
      "17961d633bcf20a7b029a7d94b7df4da2ec5427f": {
        "balance": "229427000000000000000"
      },
      ...
```

Perhaps that works fine with MSVC.

Another possibility is to bypass JSON altogether and to use C++ types for better type safety. For example,
```
inline constexpr intx::uint256 uint256_from_string(const char* str) { return intx::from_string<intx::uint256>(str); };

const std::map<evmc::address, intx::uint256> kMainnetGenesisAllocation = {
    {0x3282791d6fd713f1e94f4bfd565eaa78b3a0599d_address, uint256_from_string("1337000000000000000000")},
    {0x17961d633bcf20a7b029a7d94b7df4da2ec5427f_address, uint256_from_string("229427000000000000000")},
    ...
};

**Comment by yperbasis** [comment]
Also, read this article: https://www.fluentcpp.com/2019/07/23/how-to-define-a-global-constant-in-cpp/

I‘m thinking we should declare the allocation constants as `extern` in the header, and define them in `chain_genesis.cpp`, so that we avoid a huge header.

**Comment by AndreaLanfranchi** [comment]
Another option is to make use of [cmake-bin2h](https://github.com/sivachandran/cmake-bin2h).
Create the Json config specifications as plain text files and have the tool convert them into a byte array. Eventually nholman::json can parse from the byte array.

File: cmd/initialize_with_genesis.cpp [hunk 2]
```diff
@@ -0,0 +1,253 @@
...
+
+    app.add_option("--chainid", chain_id, "Specify id of the chain to generate");
+
+    CLI11_PARSE(app, argc, argv);
+
+    // Check flags
+    if (fs::exists(out)) {
+        SILKWORM_LOG(LogLevel::Error) << out << " already exist." << std::endl;
+        return -1;
+    }
```
AndreaLanfranchi, line 105: I think this should be expanded a little more.
Here you're checking the **destination** directory exists but actually should check data file (data.mdb) exists

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,253 @@
...
+    // We create the chaindata directory
+    fs::create_directories(out);
+
+    // We Initialize the database and open it
+    lmdb::DatabaseConfig db_config{out};
+    db_config.set_readonly(false);
+    std::shared_ptr<lmdb::Environment> env{lmdb::get_env(db_config)};
+    std::unique_ptr<lmdb::Transaction> txn{env->begin_rw_transaction()};
+    // We create all tables
+    db::table::create_all(*txn);
```
AndreaLanfranchi, line 120: I'd move the creation of all tables **after** all sanity checks on genesis config are made.

AndreaLanfranchi:
Besides the above ... honest question (my fault I don't get so ... sorry in advance)
Why should we embed config file(s) for **known** chains when the point of custom initialization of db is to set up **unknown** chains (i.e. private ones). where the init file should always be external ?

yperbasis:
> Besides the above ... honest question (my fault I don't get so ... sorry in advance)
> Why should we embed config file(s) for **known** chains when the point of custom initialization of db is to set up **unknown** chains (i.e. private ones). where the init file should always be external ?

TG staged sync is a genesis sync, and genesis sync requires premining allocation. For the main net and other known chains it‘s public and immutable, so it makes sense to embed it into the executable rather than rely on something external. TG also embeds mainnet premine. This PR is essentially porting a necessary piece of functionality from TG to Silkworm.

As to unknown/private chains, we can add their support to Silkworm later. For that we probably should use the same JSON format as TG does, so perhaps we shouldn‘t bypass JSON for mainnet genesis so that we have the same initialization logic, only the source of JSON is different.

AndreaLanfranchi:
Got it. I thought genesis block (block 0) would simply p2p download as all other blocks.

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+static void check_rlp_err(rlp::DecodingResult err) {
+    if (err != rlp::DecodingResult::kOk) {
+        throw err;
+    }
+}
+
+static intx::uint256 convert_string_to_uint256(std::string const& str) {
```
yperbasis, line 53: There already exists `intx::from_string<intx::uint256>(str)`

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+int main(int argc, char* argv[]) {
+    namespace fs = boost::filesystem;
+    int chain_id{-1};
+    CLI::App app{"Initializes database with genesis json file"};
+
+    std::string out;
+    std::string genesis;
+    app.add_option("--out", out, "Path to the newly created chaindata", true);
+
+    app.add_option("--genesis", genesis, "Path to the genesis json file", true)
```
AndreaLanfranchi, line 94: If file existance is what you need simply modify the check to `CLI::ExistingFile`

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+    // Check flags
+    if (fs::exists(fs::path(out) / fs::path("data.mdb"))) {
+        SILKWORM_LOG(LogLevel::Error) << out << " already exist." << std::endl;
+        return -1;
+    }
+
+    if (!fs::exists(genesis) && chain_id <= 0) {
+        SILKWORM_LOG(LogLevel::Error) << "genesis was not found." << std::endl;
+        return -1;
+    }
```
AndreaLanfranchi, line 110: After the check above for Existing file this becomes redundant

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+            case 1:
+                genesis_json = kMainnetGenesis;
+                alloc_json = kMainnetGenesisAllocation;
+                break;
+            case 4:
+                // genesis_json = nlohmann::json::parse(kRinkebyGenesis);
+                break;
+            case 5:
+                // genesis_json = nlohmann::json::parse(kGoerliGenesis);
+                break;
```
AndreaLanfranchi, line 142: Are those still to be implemented ?

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+    nlohmann::json genesis_json;
+    nlohmann::json alloc_json = {};
+    if (chain_id <= 0) {
+        std::ifstream t(genesis.data());
+        std::string str((std::istreambuf_iterator<char>(t)),
+                        std::istreambuf_iterator<char>());
+        genesis_json = nlohmann::json::parse(str);
+        if (genesis_json.contains("alloc")) {
+            alloc_json = genesis_json["alloc"];
+        }
```
AndreaLanfranchi, line 130: If provided genesis data file is not properly JSON formatted then parse will throw and there is no exception handling

AndreaLanfranchi (changes requested):

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+        res *= 10;
+    }
+    return res/10;
+}
+
+static bool is_hex(std::string const& s) {
+  return s.compare(0, 2, "0x") == 0
+      && s.size() > 2
+      && s.find_first_not_of("0123456789abcdefABCDEF", 2) == std::string::npos;
+}
```
AndreaLanfranchi, line 67: I'd invert the order of first two checks.
If size < 2 then compare tries to compare beyond string boundaries

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+static uint64_t get_uint_from_field(std::string const& field) {
+    int form = is_hex(field)? 16 : 10;
+    uint64_t res = strtoull(field.c_str(), nullptr, form);
```
AndreaLanfranchi, line 71: No need to check if isHex.
Providing 0 to the radix (third parameter) strtoull automatically detects the form as :
* if string begins with a digit not equal to `0` then is decimal
* if string begins with a digit equal to `0` then is octal
* if string begins with `0x` then is hexadecimal

By consequence I believe the above is_hex is pleonastic

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+    if (errno == EINVAL)
+    {
+        SILKWORM_LOG(LogLevel::Error) << field << " not a valid number." << std::endl;
+        throw;
+    }
+    else if (errno == ERANGE)
+    {
+        SILKWORM_LOG(LogLevel::Error) << field << " must be uint64" << std::endl;
+        throw;
+    }
```
AndreaLanfranchi, line 81: `throw` alone is used to _rethrow_ a catched exception ... nut here we don't have any exception as strotoull is guaranteed to be exception safe.
Plus, as mentioned elsewhere, there is no exception handling in all the program so throwing will cause a dump of the program

AndreaLanfranchi (changes requested):

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+        auto plainstate_table{txn->open(db::table::kPlainState)};
+        auto account_changeset_table{txn->open(db::table::kPlainAccountChangeSet)};
+        // Iterate over allocs
+        for (auto& [key, value] : alloc_json.items()) {
+            auto address_bytes{from_hex(key)};
+            if (address_bytes == std::nullopt) {
+                SILKWORM_LOG(LogLevel::Error) << "Cannot decode allocs from genesis" << std::endl;
+                return -1;
+            }
+            auto balance_str{value.get<std::string>()};
```
AndreaLanfranchi, line 172: Does your code run fine ?
This is wrong and dumps the app. Each allocation is as
```
    "0000000000000000000000000000000000000000": {
      "balance": "0x1"
    },
```
Which means the `value` binding is an JSON object while instead you're handling it as a string.
The correct way should be
```suggestion
            auto balance_str{value["balance"].get<std::string>()};
```

AndreaLanfranchi (changes requested):

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+        // Tables used
+        auto plainstate_table{txn->open(db::table::kPlainState)};
+        auto account_changeset_table{txn->open(db::table::kPlainAccountChangeSet)};
+        // Iterate over allocs
+        for (auto& [key, value] : alloc_json.items()) {
+            auto address_bytes{from_hex(key)};
+            if (address_bytes == std::nullopt) {
+                SILKWORM_LOG(LogLevel::Error) << "Cannot decode allocs from genesis" << std::endl;
+                return -1;
+            }
```
AndreaLanfranchi, line 171: This is partially wrong.
Should key be like `0x1234` from_hex would return a valid `Bytes` but that would be an invalid address to use below
(remember we're allowing the parsing of a user edited json file ... by default don't trust what you read)

AndreaLanfranchi (changes requested):

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,258 @@
...
+            auto balance_str{value.get<std::string>()};
+            intx::uint256 balance;        
+            Account account;
+            if (is_hex(balance_str)) {
+                auto balance_bytes{from_hex(balance_str)};
+                auto [balance_decoded, err]{rlp::read_uint256(*balance_bytes, /*allow_leading_zeros=*/true)};
+                check_rlp_err(err);
+                balance = balance_decoded;
+            } else {
```
AndreaLanfranchi, line 180: Are you sure about this ?
`rlp::read_uint256(...)` swaps endianness and as far as I know the balance from alloc is an hex value **not**  rlp encoded.
The following `account.rlp(kEmptyRoot)` would reswap endianness back

yperbasis:
Closing in favour of PR #219.

PR #219, opened 2021-04-18
Initialize with genesis take 2

AndreaLanfranchi:
This is alternative to #215 
Solves the problem of big header


AndreaLanfranchi:
Yet I don't like we have to deal with migration insertions as hex string.

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,335 @@
...
+using namespace silkworm;
+
+constexpr uint8_t genesis_body[] = {195, 128, 128, 192};
+constexpr uint8_t genesis_receipts[] = {246};
+constexpr evmc::bytes32 kNullOmmers{0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347_bytes32};
+
+constexpr size_t migrations_count = 20;
+std::string last_header_key = "LastHeader";
+std::pair<Bytes, Bytes> init_migrations_value[migrations_count] = {
+    {*from_hex("6163635f6368616e67655f7365745f6475705f736f72745f3138"),
```
yperbasis, line 47: 1. IMHO for better readability keys should be strings with values like "acc_change_set_dup_sort_18". When inserting, strings can be casted to `ByteView` with `byte_view_of_c_str` (in core/silkworm/common/util.hpp)
2. I wouldn't use migration values at all here. They only serve a debug purpose, don't they? If we need a value, we could use some dummy like `*from_hex("a0")` for everything.

So I'd have the following code:
```
constexpr const char* kMigration[] = {
    "stages_to_use_named_keys",
    "unwind_stages_to_use_named_keys",
    "stagedsync_to_use_stage_blockhashes",
    ...
}

File: cmd/initialize_with_genesis.cpp [hunk 2]
```diff
@@ -0,0 +1,335 @@
...
+                auto hash{keccak256(*address_bytes)};
+                account_rlp[to_bytes32(hash.bytes)] = account.rlp(kEmptyRoot);
+            }
+
+            auto it{account_rlp.cbegin()};
+            trie::HashBuilder hb{full_view(it->first), it->second};
+            for (++it; it != account_rlp.cend(); ++it) {
+                hb.add(full_view(it->first), it->second);
+            }
+            root_hash = hb.root_hash();
```
yperbasis, line 259: I think we should extract the allocation function into `silkworm/db` and add a test verifying that the state root matches what we expect for the main net, Rinkeby, etc. Presumably https://etherscan.io/block/0 has the correct state root.

cmd/initialize_with_genesis.cpp
```diff
@@ -0,0 +1,335 @@
...
+        txn->open(db::table::kBlockReceipts)->put(key.substr(0, 8), Bytes(genesis_receipts, 1));
+        txn->open(db::table::kHeadHeader)
+            ->put(Bytes(reinterpret_cast<const uint8_t*>(last_header_key.c_str()), last_header_key.size()),
+                  full_view(blockhash.bytes));
+        txn->open(db::table::kHeaderNumbers)->put(full_view(blockhash.bytes), key.substr(0, 8));
+        // Write Chain Config
+        auto chain_config{genesis_json["config"].dump()};
+        txn->open(db::table::kConfig)
+            ->put(full_view(blockhash.bytes),
+                  Bytes(reinterpret_cast<const uint8_t*>(chain_config.c_str()), chain_config.size()));
```
yperbasis, line 305: I would extract some functions into `silkworm/db` and add a test that
1. initializes a temp DB
2. reads the config with `parse_chain_config`
3. checks that the config equals kMainnetConfig, kRinkebyConfig, etc

That will ensure that `core/silkworm/chain/config.hpp` is in line with the JSON genesis files.

yperbasis (changes requested):

cmd/initialize_with_genesis.cpp
AndreaLanfranchi: Totally agree on this but I extend the question.
Given is very unlikely we(or anybody) use Silkworm tool to instantiate a new db do we really need to take into account migrations ? Afaik when in beta db _schema_ will be consolidated and I expect a check only on a db var.

AndreaLanfranchi: I mean ... as long as Silkworm is in catch-up phase with TG I don't think is worth the effort to take into account migrations.
cc @AlexeyAkhunov

cmd/initialize_with_genesis.cpp
AndreaLanfranchi: 👍

cmd/initialize_with_genesis.cpp
yperbasis: > Totally agree on this but I extend the question.
> Given is very unlikely we(or anybody) use Silkworm tool to instantiate a new db do we really need to take into account migrations ? Afaik when in beta db _schema_ will be consolidated and I expect a check only on a db var.

Right, that's a good point. Perhaps we can remove migrations from this PR altogether and revisit the matter later.

yperbasis (approved):

PR #222, opened 2021-04-21
Add thread ID in log header

canepat:
This change is useful for addressing multi-threaded issues. I've chosen not to make it configurable to keep the change as small as possible.


yperbasis:
@canepat please make it configurable and off by default for better compatibility with Turbo-Geth logging.

yperbasis (changes requested):
Please make it configurable and off by default for better compatibility with Turbo-Geth logging.

yperbasis (approved):

Issue #179, opened 2021-03-10
Port IntermediateHashes stage


yperbasis:

yperbasis:
Nope, it‘s still in progress. I’ve only implemented IH trie DB generation from scratch, but not incremental trie update yet.

PR #223, opened 2021-04-22
Generate IntermediateHash DB tables from scratch

yperbasis:
Part 1 of Issue #179, doesn't deal with incremental regeneration yet.

Also fixes a bug in `HashBuilder` when embedded (i.e. shorter than 32 bytes) nodes were additionally wrapped into an RLP byte array.


db/silkworm/db/chaindb.cpp
```diff
@@ -456,23 +457,60 @@ std::optional<ByteView> Table::get(ByteView key) {
...
+        return std::nullopt;
+    }
+    err_handler(rc);
+    db::Entry out;
+    out.key = db::from_mdb_val(mdb_key);
+    out.value = db::from_mdb_val(mdb_val);
+    return out;
+}
+
+std::optional<ByteView> Table::get_next_dup() {
```
AskAlexSharov, line 503: NextDup always returns same key which you passed as input parameter (pointer to same memory region). It means maybe you don’t need return it to App.

db/silkworm/trie/db_trie.cpp
```diff
@@ -0,0 +1,260 @@
...
+        e.key = unpacked_key;
+        e.value = marshal_node(node);
+        account_collector.collect(e);
+    };
+}
+
+// calculate_root algo:
+//	for iterateIHOfAccounts {
+//		if canSkipState
+//          goto use_account_trie
```
AskAlexSharov, line 92: Tabulation is broken in comment

AskAlexSharov (approved):

PR #226, opened 2021-04-27
Backport C++20 std::bit_cast + other casting improvements

yperbasis:
Kudos to @eb: https://discord.com/channels/687972960811745322/733288466837799002/836534650197377024

### DIFF
--- a/cmd/consensus.cpp
+++ b/cmd/consensus.cpp
@@ -30,6 +30,7 @@
 #include <silkworm/chain/blockchain.hpp>
 #include <silkworm/chain/difficulty.hpp>
 #include <silkworm/chain/validity.hpp>
+#include <silkworm/common/cast.hpp>
 #include <silkworm/common/util.hpp>
 #include <silkworm/rlp/decode.hpp>
 #include <silkworm/state/intra_block_state.hpp>
@@ -238,8 +239,7 @@ void init_pre_state(const nlohmann::json& pre, StateBuffer& state) {
         Bytes code{from_hex(j["code"].get<std::string>()).value()};
         if (!code.empty()) {
             account.incarnation = 1;
-            ethash::hash256 hash{keccak256(code)};
-            std::memcpy(account.code_hash.bytes, hash.bytes, kHashLength);
+            account.code_hash = bit_cast<evmc::bytes32>(keccak256(code));
             state.update_account_code(address, account.incarnation, account.code_hash, code);
         }
 
--- a/core/silkworm/trie/hash_builder.cpp
+++ b/core/silkworm/trie/hash_builder.cpp
@@ -22,6 +22,7 @@
 
 #include <ethash/keccak.hpp>
 
+#include <silkworm/common/cast.hpp>
 #include <silkworm/common/util.hpp>
 #include <silkworm/rlp/encode.hpp>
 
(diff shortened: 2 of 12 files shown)


chfast (commented):
Compare with https://github.com/wasmx/fizzy/blob/master/lib/fizzy/cxx20/bit.hpp (GCC & Clang only).

cmd/consensus.cpp
```diff
@@ -238,8 +239,7 @@ void init_pre_state(const nlohmann::json& pre, StateBuffer& state) {
         Bytes code{from_hex(j["code"].get<std::string>()).value()};
         if (!code.empty()) {
             account.incarnation = 1;
-            ethash::hash256 hash{keccak256(code)};
-            std::memcpy(account.code_hash.bytes, hash.bytes, kHashLength);
+            account.code_hash = bit_cast<evmc::bytes32>(keccak256(code));
```
AndreaLanfranchi, line 242: I think in this case a reinterpret_cast is more appropriate. Should save an intermediate copy.
Maybe the compiler takes care of removing memcpy but writing in the code directly makes more clear the intention

core/silkworm/trie/hash_builder.cpp
```diff
@@ -114,8 +115,7 @@ evmc::bytes32 HashBuilder::root_hash() {
     if (node_ref.length() == kHashLength) {
         std::memcpy(res.bytes, node_ref.data(), kHashLength);
     } else {
-        ethash::hash256 hash{keccak256(node_ref)};
-        std::memcpy(res.bytes, hash.bytes, kHashLength);
+        res = bit_cast<evmc::bytes32>(keccak256(node_ref));
```
AndreaLanfranchi, line 118: Same as comment above.

cmd/consensus.cpp
yperbasis: > I think in this case a reinterpret_cast is more appropriate. Should save an intermediate copy.
> Maybe the compiler takes care of removing memcpy but writing in the code directly makes more clear the intention

`reinterpret_cast` is dangerous, please see the comment in [Chromium's bit_cast](https://chromium.googlesource.com/chromium/src/base/+/refs/heads/main/bit_cast.h#15).

yperbasis: Also, I don't agree that `reinterpret_cast` reflects the intention better. The intention is to cast a type that is a wrapper over 32 bytes to another type that is another wrapper of 32 bytes. `reinterpret_cast` is a vague operation with few guarantees from the standard, while `bit_cast`'s purpose is exactly this, cast to a different type but preserve the actual bytes. How `bit_cast` is actually implemented is a separate concern.

AndreaLanfranchi: reinterpret_cast is dangerous when casting different types but in this case we have the very same type with same size and same memory alignment under a different symbol name (always 32 bytes) thus I consider heavier to
* create the destination object
* memcopy data from source
* copy contruct target 
rather than 
* treat source mem location "as if it was of target type"
* copy construct from there

Also in this case the preservation of actual bytes out of scope as the original object is temp and discarded quickly.
Anyway ... I can live with it.

AndreaLanfranchi (approved):

cmd/consensus.cpp
yperbasis: On my Linux machine (Ubuntu 20.10, gcc 10.2.0) `evmc::bytes32` (and `evmc_bytes32`) has alignment of 1, while `ethash::hash256` has alignment of 8.

AndreaLanfranchi: Cuz you're considering the alignment of ethash::hash256 **union struct** where the alignment is taken from first type (uint64_t) while evmc::bytes32 though a struct has the alignment of an array of bytes (thus 1)

```
    hash256 h{};
    assert(alignof(h) != alignof(h.bytes);
```

In fact I was reintpret casting hash **bytes** not the whole hash256

PR #227, opened 2021-04-28
Refactor ChainConfig to have an array of fork blocks instead of individual members

yperbasis:


core/silkworm/chain/config.hpp
```diff
@@ -19,164 +19,129 @@
...
-    }
-
-    // https://eips.ethereum.org/EIPS/eip-1679
-    // https://ecips.ethereumclassic.org/ECIPs/ecip-1088
-    std::optional<uint64_t> istanbul_block;
-    bool has_istanbul(uint64_t block_num) const noexcept {
-        return istanbul_block.has_value() && istanbul_block.value() <= block_num;
-    }
+    // Block numbers of forks that have an evmc_revision value
+    std::array<std::optional<uint64_t>, EVMC_MAX_REVISION> fork_blocks{};
```
mriccobene, line 48: Consider the addition of a method like this:
```
    std::optional<uint64_t> block_number(evmc_revision rev) const noexcept {
        return fork_blocks[rev];
    }
```
Although it is not necessary it better clarifies the link between fork_blocks array index and evmc_revision because in the following line it is not self evident (only the constant EVMC_MAX_REVISION suggests it):

`std::array<std::optional<uint64_t>, EVMC_MAX_REVISION> fork_blocks{};`

It would also be nice in all software to be able to use

`using BlockNum = uint64_t;`

so as to distinguish a uint64_t which is a block number from one which is something else. 
But this is a change that I would leave for the future

mriccobene (commented):
Ok for me! I left just one suggestion.

Note: I leave the approval to members more experienced than me on the project

core/silkworm/chain/config.hpp
AndreaLanfranchi: @mriccobene is this what you had n mind ?
https://github.com/torquem-ch/silkworm/pull/227/commits/68532661326c51decf8950724a8cd0195f9cb420

AndreaLanfranchi (approved):
Good for me

cmd/check_senders.cpp
```diff
@@ -596,7 +596,7 @@ class RecoveryFarm final {
             }
 
             if (transaction.chain_id) {
-                if (has_spurious_dragon) {
+                if (!has_spurious_dragon) {
```
AndreaLanfranchi, line 599: Nice catch. Thank you

AndreaLanfranchi (approved):

core/silkworm/chain/config.hpp
mriccobene: @AndreaLanfranchi yes! Now that I see my bug I think this method is necessary! Thanks!

PR #228, opened 2021-04-28
Fix class-memaccess gcc warning

yperbasis:
Follow-up to PR #226

### DIFF
--- a/cmd/consensus.cpp
+++ b/cmd/consensus.cpp
@@ -239,7 +239,7 @@ void init_pre_state(const nlohmann::json& pre, StateBuffer& state) {
         Bytes code{from_hex(j["code"].get<std::string>()).value()};
         if (!code.empty()) {
             account.incarnation = 1;
-            account.code_hash = bit_cast<evmc::bytes32>(keccak256(code));
+            account.code_hash = bit_cast<evmc_bytes32>(keccak256(code));
             state.update_account_code(address, account.incarnation, account.code_hash, code);
         }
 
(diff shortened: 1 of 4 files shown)


cmd/consensus.cpp
```diff
@@ -239,7 +239,7 @@ void init_pre_state(const nlohmann::json& pre, StateBuffer& state) {
         Bytes code{from_hex(j["code"].get<std::string>()).value()};
         if (!code.empty()) {
             account.incarnation = 1;
-            account.code_hash = bit_cast<evmc::bytes32>(keccak256(code));
+            account.code_hash = bit_cast<evmc_bytes32>(keccak256(code));
```
AndreaLanfranchi, line 242: I believe 
```suggestion
            account.code_hash = bit_cast<evmc::bytes32>(keccak256(code).bytes);
```
would have achieved the same result while preserving clearness in code with explicit casting to code_hash type (not the one it inherits from.
Mind a try ?

yperbasis: @AndreaLanfranchi That doesn't work: gcc is still complaining
```
error: ‘void* memcpy(void*, const void*, size_t)’ copying an object of non-trivial type ‘struct evmc::bytes32’ from an array of ‘const unsigned char [32]’ [-Werror=class-memaccess]
   42 |     std::memcpy(&dst, &src, sizeof(To));
      |     ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

Issue #144, opened 2021-01-25
Ethash PoW block validation


yperbasis:
As a prerequisite, we should probably restore `ethash` Hunter package. I removed it with PR #115 because it was not compatible with Wasm. Ideally we should find a way to make `ethash` package compatible with Wasm, perhaps using some macros/options, and then revert PR #115.

There's a couple of consensus tests that check Ethash PoW: [wrongMixHash](https://github.com/ethereum/tests/blob/v7.0.0/BlockchainTests/InvalidBlocks/bcInvalidHeaderTest/wrongMixHash.json) & [wrongNonce](https://github.com/ethereum/tests/blob/v7.0.0/BlockchainTests/InvalidBlocks/bcInvalidHeaderTest/wrongNonce.json). Unfortunately, they were removed in the latest release.

AndreaLanfranchi:
Refer to #240

PR #241, opened 2021-05-11
BlockHeader PoW Verification

AndreaLanfranchi:
Wrt #144

### DIFF
--- a/core/silkworm/chain/validity.cpp
+++ b/core/silkworm/chain/validity.cpp
@@ -15,7 +15,6 @@
 */
 
 #include "validity.hpp"
-#include <boost/endian.hpp>
 #include <silkworm/crypto/ecdsa.hpp>
 #include <silkworm/trie/vector_root.hpp>
 #include <ethash/ethash.hpp>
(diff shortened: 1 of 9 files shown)


core/silkworm/chain/validity.cpp
```diff
@@ -106,6 +106,29 @@ ValidationResult validate_block_header(const BlockHeader& header, const StateBuf
         }
     }
 
+    // Ethash PoW verification
+    if (header.mix_hash) {
```
yperbasis, line 110: Please add an option to ChainConfig allowing to opt out of PoW validation instead. We can opt out of PoW validation for tests, but not for `kMainnetConfig`. Otherwise there is a security risk that validate_block_header passes bogus blocks on the mainnet.

yperbasis (changes requested):

core/silkworm/chain/validity.cpp
AndreaLanfranchi: Right. Will do

AndreaLanfranchi: Done. How to reinclude the tests removed due to #144 ?
(I manually tested them adding the files and they're ok)

yperbasis (approved):

ethash/ethash/ethash.cpp
```diff
@@ -283,24 +283,36 @@ bool verify_light(const hash256& header_hash, const hash256& mix_hash, uint64_t
...
+    if (!is_equal(mix_hash, expected_mix_hash)) {
+        return VerificationResult::kInvalidMixHash;
+    }
+    return VerificationResult::kOk;
+}
+
+VerificationResult verify_full(const uint64_t block_num, const hash256& header_hash, const hash256& mix_hash,
+                               uint64_t nonce, const hash256& boundary) noexcept {
+
+    static epoch_context_ptr epoch_context{nullptr, nullptr};
```
chfast, line 303: This is not thread-safe.

AndreaLanfranchi: Yes but we actually don't need it to be.
All PoW verifications in Silkworm are single threaded

PR #244, opened 2021-05-20
Less boost

AndreaLanfranchi:
Small reduction from boost's headers

### DIFF
--- a/db/silkworm/common/os.hpp
+++ b/db/silkworm/common/os.hpp
@@ -1,0 +1,51 @@
+/*
+   Copyright 2020-2021 The Silkworm Authors
+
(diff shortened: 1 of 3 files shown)


db/silkworm/common/os.hpp
```diff
@@ -0,0 +1,51 @@
+/*
+   Copyright 2020-2021 The Silkworm Authors
```
yperbasis, line 2: Should be simply 2021

yperbasis (approved):

db/silkworm/common/os.hpp
AndreaLanfranchi: Yes. You're right. Changed

Issue #231, opened 2021-05-03
EIP-1559: Fee market change for ETH 1.0 chain


yperbasis:
Implement [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559). Part of [London hard fork](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/mainnet-upgrades/london.md).

PR #245, opened 2021-05-25
EIP-1559: Fee market change for ETH 1.0 chain

yperbasis:
Issue #231

### DIFF
--- a/core/silkworm/execution/processor.cpp
+++ b/core/silkworm/execution/processor.cpp
@@ -38,7 +38,8 @@ ValidationResult ExecutionProcessor::validate_transaction(const Transaction& txn
         return ValidationResult::kWrongNonce;
     }
 
-    intx::uint512 gas_cost{intx::umul(intx::uint256{txn.gas_limit}, txn.gas_price)};
+    // TODO[Issue 231] EIP-1559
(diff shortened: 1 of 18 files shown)


core/silkworm/execution/processor.cpp
```diff
@@ -82,7 +88,8 @@ Receipt ExecutionProcessor::execute_transaction(const Transaction& txn) noexcept
     uint64_t gas_used{txn.gas_limit - refund_gas(txn, vm_res.gas_left)};
```
AndreaLanfranchi, line 88: I believe this can be defined const too

core/silkworm/types/block.cpp
```diff
@@ -69,6 +70,9 @@ namespace rlp {
             rlp_head.payload_length += kHashLength + 1;  // mix_hash
             rlp_head.payload_length += 8 + 1;            // nonce
         }
+        if (header.base_fee_per_gas) {
+            rlp_head.payload_length += length(*header.base_fee_per_gas);
+        }
```
AndreaLanfranchi, line 75: For better readability I'd prefer always to check the `.has_value()`  method (if you don't mind).
This makes clear this is an std::optional regardless the contained value

AndreaLanfranchi (approved):
Good for me

Issue #235, opened 2021-05-03
EIP-3541: Reject new contracts starting with the 0xEF byte


yperbasis:
Implement EIP-3541.

_Warning_: it's a part of [Baikal Testnet](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/client-integration-testnets/baikal.md). Double check that this EIP gets included into London.

PR #247, opened 2021-05-27
EIP-3541: Reject new contracts starting with the 0xEF byte

yperbasis:
Implement [EIP-3541](https://eips.ethereum.org/EIPS/eip-3541) (Issue #235)


AndreaLanfranchi (approved):

AndreaLanfranchi:
Just for my education: isn't it worth checking for `ef` into 2nd byte of txn.data, thus invalidating early before calling `evm::create(...)` ?

yperbasis:
> Just for my education: isn't it worth checking for `ef` into 2nd byte of txn.data, thus invalidating early before calling `evm::create(...)` ?

Init code can be in arbitrary format, it doesn't have to start with PUSH <contract code>.

PR #255, opened 2021-06-02
EVM::execute Precondition: txn.from must be recovered

AndreaLanfranchi:
Evm test fail if txn.from is empty


yperbasis (approved):

AndreaLanfranchi:
Question: assertions gets disabled if `NDEBUG` is set.
Why not throw and invalid_argument exception ?

yperbasis:
> Question: assertions gets disabled if `NDEBUG` is set.
> Why not throw and invalid_argument exception ?

Because we don't use exceptions in core for WebAssembly compatibility.

AndreaLanfranchi:
> Because we don't use exceptions in core for WebAssembly compatibility.

Yeah right ... always forget this

PR #257, opened 2021-06-04
Fix access to already deallocated memory in cmd/hashstate

yperbasis:
`MDB_val mdb_key{db::to_mdb_val(Bytes(8, '\0'))};`
contains a bug. `Bytes(8, '\0')` creates a temporary object, which is destroyed after the line. And later we try to access an already deallocated memory via the view. (`MDB_val` is a view, essentially the same as `std::string_view`)


AndreaLanfranchi (approved):
Yes right

Giulio2002 (approved):

PR #260, opened 2021-06-06
Code Refactoring for StagedSync

Giulio2002:


cmd/blockhashes.cpp
```diff
@@ -38,102 +39,12 @@ int main(int argc, char* argv[]) {
...
-        } else {
-            SILKWORM_LOG(LogLevel::Info) << "Nothing to process" << std::endl;
-        }
-
-        SILKWORM_LOG(LogLevel::Info) << "All Done" << std::endl;
+        lmdb::DatabaseConfig db_config{db_path};
+        db_config.set_readonly(false);
+        std::shared_ptr<lmdb::Environment> env{lmdb::get_env(db_config)};
+        std::unique_ptr<lmdb::Transaction> txn{env->begin_ro_transaction()};
+        stagedsync::stage_blockhashes(db_path, txn.get());
```
AndreaLanfranchi, line 47: I think every single stage should return some sort of code stating their work has completed correctly or some condition has happened. I see the stage_blockhashes is in a try catch block but this is limited to a proper program flow execution. I mean we don't have info back from the stage whether or not we can recover from some condition.
My two cents.

Giulio2002:
@AndreaLanfranchi I agree with you, Do you agree on these error codes? @yperbasis any thoughts on this PR?

yperbasis:
With respect to error handling my thinking is the following. When it's something we can do nothing about, like the database is corrupted, we should print an error message and exit the process (e.g. by throwing an exception), since there's little the program can do besides. However, when an error is recoverable—for example, a bad block whose state root calculated at [Stage 7](https://github.com/ledgerwatch/erigon/tree/devel/eth/stagedsync#stage-7-compute-state-root-stage) doesn't match the root declared in its header—we should report the condition and recover from it correctly. Whether it's reported via an error code or a dedicated exception I'm on the fence. Perhaps error codes are better.

Also, apart from stage progress functions, we'll need unwind functions as well. See https://github.com/ledgerwatch/interfaces/blob/master/_docs/staged-sync.md & https://github.com/ledgerwatch/erigon/tree/devel/eth/stagedsync. So a stage could be described by something like
```
struct {
    id,
    progress_func,
    unwind_func,
    // possibly prune func
};
```
P.S. We should use Erigon for inspiration when it makes sense (see [stagebuilder.go](https://github.com/ledgerwatch/erigon/blob/devel/eth/stagedsync/stagebuilder.go)), though sometimes Erigon is over-engineered IMHO.

Giulio2002:
now it should be looking better, if you agree on this i will make it for all the other stages

yperbasis:
> now it should be looking better, if you agree on this i will make it for all the other stages

Sounds good. We'll revisit staged sync in the future, but this is a good first step IMHO.

By the way, `kStageInvalidHashLength` means that the DB is corrupted and isn't really recoverable.

Giulio2002:
The PR is finished. @yperbasis @AndreaLanfranchi could you please review now? feel free to make changes to the code if you want to.

yperbasis:
`tests` submodule shouldn't change

db/silkworm/stagedsync/util.hpp
```diff
@@ -0,0 +1,47 @@
...
+    kStageInvalidRange,
+    kStageAborted
+};
+
+void check_stagedsync_error(StageResult code);
+
+ /*
+ * Extract the incarnation from an encoded account object without fully decoding it.
+ */
+uint64_t extract_incarnation(ByteView);
```
yperbasis, line 40: Pls move to `core/silkworm/types/account.hpp`

db/silkworm/stagedsync/util.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+// Return db formatted (key, value) pair
+std::pair<Bytes, Bytes> convert_to_db_format(Bytes& key, Bytes& value);
```
yperbasis, line 43: What does this method do? The name and the comment are too generic, I struggle to understand what it's for.
Also, please take the arguments as `ByteView`  or `const Bytes&` unless you're changing them. Rationale: [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) and https://rules.sonarsource.com/cpp/RSPEC-6009

db/silkworm/stagedsync/stagedsync.hpp
```diff
@@ -0,0 +1,85 @@
...
+/* HashState Promotion Functions*/
+
+/*
+    * Operation is used to distinguish what bucket we want to generated
+    * HashAccount is for genenerating HashedAccountBucket
+    * HashStorage is for genenerating HashedStorageBucket
+    * Code generates hashed key => code_hash mapping
+
+*/
+enum class HashstateOperation {
```
yperbasis, line 53: `HashstateOperation` shouldn't be in `db/silkworm/stagedsync/stagedsync.hpp` since it's not something pertaining to all stages, but specific for a particular stage.

db/silkworm/stagedsync/stagedsync.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <silkworm/stagedsync/util.hpp>
+#include <vector>
+
+#ifndef SILKWORM_STAGEDSYNC_HPP_
+#define SILKWORM_STAGEDSYNC_HPP_
+
+namespace silkworm::stagedsync {
+
+extern bool write_receipts;
+constexpr size_t kBatchSize = 512 * kMebi;
```
yperbasis, line 27: `write_receipts` & `kBatchSize` shouldn't be in `db/silkworm/stagedsync/stagedsync.hpp` since it's not something pertaining to all stages, but specific for a particular stage.

yperbasis (changes requested):

db/silkworm/stagedsync/stagedsync.hpp
Giulio2002: I would like to be able to use promotions in the cmd freely tool in the cmd directory in case we need to test them in the future,  so that's why i kept cmd/hashstate the same and did not use stage_hashstate function there. as you can see you need to specify an Operation in promote

db/silkworm/stagedsync/stagedsync.hpp
Giulio2002: write_receipts needs to be modified in case receipts are disabled or not outside of stage_execution, and that is why it is an extern. kBatchSize is both used in stage_senders and stage_execution.

yperbasis: Fine about `kBatchSize`, we can have it in `stagedsync.hpp` for now.

But `write_receipts` shouldn't be there. The current logic in master in `cmd/execute.cpp` is
```
bool write_receipts{db::read_storage_mode_receipts(*txn)};
```
Let's keep the current logic and obtain the value of `write_receipts` from the database in the execution code (no need for an `extern` variable).

Giulio2002: oh, i did not notice that

db/silkworm/stagedsync/util.hpp
Giulio2002: I put a comment that makes more sense now. it is the FromDbFormat function from erigon.

yperbasis (approved):
Looks good to me as a first step towards fully-fledged staged sync in Silkworm.

@AndreaLanfranchi what are your thoughts?

AndreaLanfranchi (approved):
I believe is good for me too.

Personally I think some more checks should be performed to enforce db consistency.
For example `stage_blockhashes` does a read of blocks until MDB_NOTFOUND but is not completely accurate : we should check for instance that **all** blocks processed by block downloader gets processed.
Say we have downloaded 10 blocks but for any reason `stage_blockhashes` reads up to block 9 and returns NOT FOUND on block 10 ... actually the routine returns no error and persist 9 into the stage progress but in reality we've stumbled into a db inconsistency and on the next loop this stage could not go beyond block 9

This said we can adjust later while avoiding this PR gets too big.

PR #282, opened 2021-07-11
max_size instead of inmemory in db::EnvConfig

yperbasis:
max_size is more precise than inmemory.
Also `db_test` now runs fine under Address Sanitizer.

### DIFF
--- a/db/silkworm/db/access_layer_test.cpp
+++ b/db/silkworm/db/access_layer_test.cpp
@@ -85,6 +85,7 @@ namespace db {
         // Empty dir
         std::string empty{};
         db::EnvConfig db_config{empty};
+        db_config.max_size = 64 * kMebi;
         REQUIRE_THROWS_AS(db::open_env(db_config), std::invalid_argument);
 
         // Conflicting flags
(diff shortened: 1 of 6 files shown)


AndreaLanfranchi (commented):
I don't get the point. Inmem is consistent with erigon and has a single point for setting default sizes. Better a bool then having the need to compute the size. Sorry but don't agree in this.

AndreaLanfranchi:
Also you have removed the setting of MDBX_NOMETASYNC flag

AndreaLanfranchi:
I also don't get the point of setting max size to 64MB in the places where it was set to inmem if the default value is already 64MB.
So any db can't grow beyond 64MB ?

yperbasis:
@AndreaLanfranchi `inmemory` is only used in tests, right? There are two issues with it. First, the name is not accurate: a file is still created, so it's not completely in-memory. Second, it conflates two things: `MDBX_NOMETASYNC` & memory mapping geometry. I don't see any benefit of having `MDBX_NOMETASYNC` in tests, but not in production. What's the point? But a small memory geometry is important for test performance, especially under AddressSanitizer.

So the purpose of this PR is to remove `MDBX_NOMETASYNC` (used only in tests and IMHO unnecessary), and to have a clearer name for the existing functionality. But if you are still not convinced, that's fine, I can rework this PR to leave `inmemory` as is. What do you say?

yperbasis:
> I also don't get the point of setting max size to 64MB in the places where it was set to inmem if the default value is already 64MB.
> So any db can't grow beyond 64MB ?

That's only set in tests so that run fast under Address Sanitizer.

AndreaLanfranchi:
> @AndreaLanfranchi `inmemory` is only used in tests, right? 

Actually yes but it **might** be used in small migrations too. 

> There are two issues with it. First, the name is not accurate: a file is still created, 

Yes a file is always created but the name states `the whole map fits in RAM`

>so it's not completely in-memory. Second, it conflates two things: `MDBX_NOMETASYNC` & memory mapping geometry. I don't see any benefit of having `MDBX_NOMETASYNC` in tests, 

We don't care in tests if they got interrupted by a power outage so durability for a test db is not an issue. I.e. don't care if METADATA and ROOT are Fsynced to disk.

Yet I don't know what ASAN triggers. Can you elaborate ?

> So the purpose of this PR is to remove `MDBX_NOMETASYNC` (used only in tests and IMHO unnecessary), and to have a clearer name for the used functionality. But if you are still not convinced, that's fine, I can rework this PR to leave `inmemory` as is. What do you say?

I strongly believe, again, a bool with controlled hardcoded sizes is way better than having the freedom to choose any arbitrary value. Why not 128MB or 1GB instead ?

yperbasis:
OK, closing in favour of PR #284 and #285.

PR #284, opened 2021-07-12
Fix double deletion of a TemporaryDirectory

yperbasis:
`TEST_CASE("Db Opening")` was crashing or triggering Address Sanitizer for me.


AndreaLanfranchi (approved):
Yes this a way which however inserts a non-copyable constraint.
Another way could have been to release/delete owned directories not in destructor rather in specific method (say `dispose`) and delete directories only if not disposed (hence preventing double delete).
However this looks good to me.

PR #286, opened 2021-07-12
Clarify the difference between code and recipient

yperbasis:
An `evmc_message` contains only two addresses (sender and "destination").
However, in case of `DELEGATECALL` we need 3 addresses (sender, recipient, and code), so we recover the missing 3rd address from the `address_stack_`.

See Section 8 "Message Call" of the Yellow Paper.
`destination` in `evmc_message` can mean either code or recipient, depending on the context.

### DIFF
--- a/core/silkworm/execution/evm.cpp
+++ b/core/silkworm/execution/evm.cpp
@@ -161,7 +161,7 @@ evmc::result EVM::call(const evmc_message& message) noexcept {
     }
 
     // See Section 8 "Message Call" of the Yellow Paper for the difference between code & recipient.
-    // destination in evmc_message can mean either code or recipent, depending on the context.
+    // destination in evmc_message can mean either code or recipient, depending on the context.
     const evmc_address code_address{message.destination};
     const evmc_address recipient_address{recipient_of_call_message(message)};
 
--- a/core/silkworm/execution/evm.cpp
+++ b/core/silkworm/execution/evm.cpp
@@ -230,9 +230,9 @@ evmc_address EVM::recipient_of_call_message(const evmc_message& message) noexcep
         return message.sender;
     } else if (message.kind == EVMC_DELEGATECALL) {
         // An evmc_message contains only two addresses (sender and "destination").
-        // However, in case of DELEGATECALL we need 3 addresses (sender, recipient, and code),
-        // so we recover the missing 3rd address from the caller_stack_.
-        return caller_stack_.top();
+        // However, in case of DELEGATECALL we need 3 addresses (sender, code, and recipient),
+        // so we recover the missing recipient address from the address_stack_.
+        return address_stack_.top();
     } else {
         assert(message.kind == EVMC_CALL);
         return message.destination;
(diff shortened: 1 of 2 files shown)


chfast:
This issue was brought up some other time (cannot find it now): the single `DELEGATECALL` case requires the `evmc::Host` to track the recipient address because this information cannot be recovered from the `evmc::message`. Otherwise the Host implementation can be "stateless".

I think it is time to solve it on EVMC level by adding one more `evmc::address` field.
If you have good spec insights, can you suggest names for these?

core/silkworm/execution/evm.cpp
```diff
@@ -154,62 +154,63 @@ evmc::result EVM::create(const evmc_message& message) noexcept {
...
 
-    auto snapshot{state_.take_snapshot()};
+    const auto snapshot{state_.take_snapshot()};
 
     if (message.kind == EVMC_CALL) {
         if (message.flags & EVMC_STATIC) {
             // Match geth logic
             // https://github.com/ethereum/go-ethereum/blob/v1.9.25/core/vm/evm.go#L391
-            state_.touch(message.destination);
+            state_.touch(recipient_address);
```
chfast, line 182: This seems to be fixing a consensus bug. Are there any tests failing?

yperbasis: > This seems to be fixing a consensus bug. Are there any tests failing?

Not really because this code is executed only when `message.kind == EVMC_CALL`

yperbasis:
> This issue was brought up some other time (cannot find it now): the single `DELEGATECALL` case requires the `evmc::Host` to track the recipient address because this information cannot be recovered from the `evmc::message`. Otherwise the Host implementation can be "stateless".
> 
> I think it is time to solve it on EVMC level by adding one more `evmc::address` field.
> If you have good spec insights, can you suggest names for these?

I'd suggest to split "destination" into "recipient" and "code", in line with Section 8 "Message Call" of the Yellow Paper.

core/silkworm/execution/evm.cpp
```diff
@@ -224,7 +225,22 @@ evmc::result EVM::call(const evmc_message& message) noexcept {
     return res;
 }
 
+evmc_address EVM::recipient_of_call_message(const evmc_message& message) noexcept {
```
chfast, line 228: This is similar code for it in Aleth: https://github.com/ethereum/aleth/blob/master/libevm/ExtVMFace.cpp#L206-L208

I'm usually confused by this, but it seems that for the `CALLCODE` the answer is also `message.destination`.

chfast:
> I'd suggest to split "destination" into "recipient" and "code", in line with Section 8 "Message Call" of the Yellow Paper.

Sounds good except I cannot name it "code" :). Then `code_address`, `recipient_address`, `sender_address`? Or `code_address`, `recipient`, `sender`? Or `code_account`?

yperbasis:
> > I'd suggest to split "destination" into "recipient" and "code", in line with Section 8 "Message Call" of the Yellow Paper.
> 
> Sounds good except I cannot name it "code" :). Then `code_address`, `recipient_address`, `sender_address`? Or `code_address`, `recipient`, `sender`? Or `code_account`?

My preference would be `code_address`, `recipient`, `sender`

yperbasis:
@chfast I think a better idea would be not to change `evmc::message`, but to add `code_address` as an argument to `evmc_host_interface::call`. I'll quickly try if it works.

yperbasis:
@chfast Oh well, extending `evmc_host_interface::call` seems to be too much work on the evmc side.

I've created https://github.com/ethereum/evmc/pull/611, https://github.com/ethereum/evmone/pull/360, and https://github.com/torquem-ch/silkworm/pull/288 that add `code_address` to `evmc_message`.

PR #289, opened 2021-07-14
Fix AddressSanitizer: stack-use-after-scope in read_chain_config

yperbasis:
```
auto key{to_slice(block_key(0))};
```
is problematic.

`block_key(0)` returns a temporary `Bytes` instance, which is then destroyed at the end of the line, so we end up with a view of a destroyed (byte)string.


AndreaLanfranchi (approved):
Right

PR #290, opened 2021-07-15
magic_enum as submodule instead of file in db/silkworm/common

yperbasis:
magic_enum is an external library, and we don't modify it in any way, so it's better not to copy it to our code base verbatim.

Also, https://app.codecov.io/gh/torquem-ch/silkworm excludes code from submodules automatically.


AndreaLanfranchi (approved):
Fair enough

PR #291, opened 2021-07-15
Db misc

AndreaLanfranchi:
* Remove ambiguity on datapath passed by command line (whether it is a file or a directory)
* Introduce `DataDirectory` class which holds and manage correct paths (actually calling method parent() on datadir returns different result whether original input has a trailing slash or not - eg. parent of c:\test\data\ is still c:\test\data while parent of c:\test\data is c:\test)
* Introduce db::for_each
* Introduce db::for_count


core/silkworm/common/util.cpp
```diff
@@ -231,4 +226,18 @@ size_t prefix_length(ByteView a, ByteView b) {
     }
     return len;
 }
+
+std::vector<std::string> split(std::string source, std::string delimiter) {
```
yperbasis, line 230: I know it's trivial, but could you please add a test case?

db/silkworm/db/mdbx.cpp
```diff
@@ -28,6 +27,9 @@ ::mdbx::env_managed open_env(const EnvConfig& config) {
 
     // Check datafile exists if create is not set
     fs::path db_path{config.path};
+    if (db_path.has_filename()) {
+        db_path += std::filesystem::path::preferred_separator;  // Remove amibiguity. It hash to be a directory
```
yperbasis, line 31: "hash" typo

yperbasis (approved):

core/silkworm/common/util.cpp
AndreaLanfranchi: Yeh right. Which also made emerge a couple of cases I did not think of.
Thank you

db/silkworm/db/mdbx.cpp
AndreaLanfranchi: And "amibiguity" too

PR #293, opened 2021-07-15
Use iov_base to avoid problems with mdbx::byte type aliasing

### LINKED ISSUE #292: C++20 and MDBX
Due to this aliasing in MDBX

```
namespace mdbx {

// Functions whose signature depends on the `mdbx::byte` type
// must be strictly defined as inline!
#if defined(DOXYGEN) || (defined(__cpp_char8_t) && __cpp_char8_t >= 201811)
// Wanna using a non-aliasing type to release more power of an optimizer.
using byte = char8_t;
#else
// Wanna not using std::byte since it doesn't add features,
// but add inconvenient restrictions.
using byte = unsigned char;
#endif /* __cpp_char8_t >= 201811*/
```

with C++20 mdbx::byte is now a different type : `char8_t` != `unsigned char`

By consequence `byte_ptr()` property of `slice` returns a different type pointer incompatible with our `Byteview` and and `boost:endian`

Always use `iov_base` (which is void*) and cast it to required type.

canepat:
Add path for included CMake files to fix build as submodule
Fixes #292


db/silkworm/db/access_layer.cpp
```diff
@@ -35,11 +35,11 @@ std::optional<version_t> get_schema_version(mdbx::txn& txn) noexcept {
...
     auto data{src.current()};
     assert(data.value.length() == 12);
-    auto Major{boost::endian::load_big_u32(data.value.byte_ptr())};
+    auto Major{boost::endian::load_big_u32(static_cast<unsigned char*>(data.value.iov_base))};
     data.value.remove_prefix(sizeof(uint32_t));
-    auto Minor{boost::endian::load_big_u32(data.value.byte_ptr())};
+    auto Minor{boost::endian::load_big_u32(static_cast<unsigned char*>(data.value.iov_base))};
     data.value.remove_prefix(sizeof(uint32_t));
-    auto Patch{boost::endian::load_big_u32(data.value.byte_ptr())};
+    auto Patch{boost::endian::load_big_u32(static_cast<unsigned char*>(data.value.iov_base))};
```
AndreaLanfranchi, line 42: Please be consistent ... as in all other cases use uint8_t

silkworm.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
+	"folders": [
+		{
+			"path": "."
+		}
+	]
+}
```
AndreaLanfranchi, line 7: Please add .code-workspace to .gitignore

AndreaLanfranchi (changes requested):

AndreaLanfranchi:
Addresses #292

db/silkworm/db/access_layer.cpp
canepat: OK

silkworm.code-workspace
canepat: OK

AndreaLanfranchi (approved):
Good for me

PR #306, opened 2021-07-19
Use db::get_storage_mode() instead of reading single modes individually

AndreaLanfranchi:


db/silkworm/db/access_layer.cpp
```diff
@@ -96,7 +96,7 @@ storage_mode_t get_storage_mode(mdbx::txn& txn) noexcept {
     return ret;
 }
 
-void set_storage_mode(mdbx::txn& txn, storage_mode_t& val) {
+void set_storage_mode(mdbx::txn& txn, StorageMode& val) {
```
yperbasis, line 99: Should it be **const** StorageMode& ?

yperbasis (approved):

db/silkworm/db/access_layer.cpp
AndreaLanfranchi: Right

PR #311, opened 2021-07-20
Use Intel intrinsics for SHA256

yperbasis:
Use [Intel SHA extensions](https://en.wikipedia.org/wiki/Intel_SHA_extensions) for SHA256 when available. Fallback to [BMI instructions](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set).


AndreaLanfranchi:
This looks good to me but not unsure about the rationale.
Is it due to improved performance ?

yperbasis:
> This looks good to me but unsure about the rationale.
> Is it due to improved performance ?

Yes, the goal is to improve performance.
I've compared SHA256 performance against the baseline code without CPU extensions on my MacBook (Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz).
The SHA version is about 80% faster than the baseline, and the BMI version is about 15% faster.

core/silkworm/crypto/sha-256.c
```diff
@@ -23,99 +50,426 @@
...
+__attribute__((target("bmi,bmi2"))) static void sha_256_x86_bmi(uint32_t h[8], const void *input, size_t len) {
+    sha_256_implementation(h, input, len);
+}
+
+// The following function was adapted from https://github.com/noloader/SHA-Intrinsics
+/*   Intel SHA extensions using C intrinsics               */
+/*   Written and place in public domain by Jeffrey Walton  */
+/*   Based on code from Intel, and by Sean Gulley for      */
+/*   the miTLS project.                                    */
+__attribute__((target("sha,sse4.1"))) static void sha_256_x86_sha(uint32_t h[8], const void *input, size_t len) {
```
chfast, line 245: Did you try to use function multi-versioning? https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html

yperbasis: > Did you try to use function multi-versioning? https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html

I've just tried and my clang say "error: function multiversioning doesn't support feature 'sha'"

PR #312, opened 2021-07-21
Cleaner Senders stage

Giulio2002:


AndreaLanfranchi:
I believe g_should_stop should stay atomic and be used in RecoveryWorker too (it was used there).

yperbasis (changes requested):
I concur with @AndreaLanfranchi.
@Giulio2002 could you please make `g_should_stop` into `std::atomic_bool` again.

yperbasis (approved):

PR #314, opened 2021-07-22
MDBX geometry: match Erigon better

yperbasis:


db/silkworm/db/mdbx.cpp
```diff
@@ -90,7 +92,10 @@ ::mdbx::env_managed open_env(const EnvConfig& config) {
     op.max_readers = config.max_readers;
 
     ::mdbx::env_managed ret{db_path.native(), cp, op, config.shared};
-    // TODO (Andrea) C++ bindings don't have setoptions
+
+    // C++ bindings don't have setoptions
+    ::mdbx::error::success_or_throw(::mdbx_env_set_option(ret, MDBX_opt_rp_augment_limit, 32 * 1024 * 1024));
+
```
AndreaLanfranchi, line 98: I'm not sure this is the correct way (i'm stil testing).
If you notice erigon's code you'll notice options are set **before** env is opened whilest the creation of instance of env_managed immediately opens the underlying data file

yperbasis: I see that `rp_augment_limit` is used in `mdbx_page_alloc()`, so it't probably OK to set it after the environment is created.

AndreaLanfranchi: Sorry probably we're not on the same page. Env created != Env opened.

yperbasis: What I mean is that mdbx_page_alloc() is called most of the time during normal operations, after the environment is created and opened, so it should be OK to set MDBX_opt_rp_augment_limit later.

AndreaLanfranchi (approved):

PR #315, opened 2021-07-22
More erigon's settings on db opening

AndreaLanfranchi:


cmd/dbtool.cpp
```diff
@@ -129,6 +129,7 @@ struct dbFreeInfo {
 
 struct DbOptions {
     std::string datadir{DataDirectory{}.get_chaindata_path().string()};  // Where data file is located
+    bool shared{false}; // Whether db has to be opened in shared mode
```
yperbasis, line 132: This field is not passed into EnvConfig

yperbasis (changes requested):

yperbasis (approved):

PR #319, opened 2021-07-23
Restore stage uniformity

yperbasis:
Stage uniformity is useful for future work (something similar to `erigon/eth/stagedsync/default_stages.go`).

### DIFF
--- a/cmd/execute.cpp
+++ b/cmd/execute.cpp
@@ -36,13 +36,22 @@ int main(int argc, char* argv[]) {
     app.add_option("--chaindata", chaindata, "Path to a database populated by Erigon", true)
         ->check(CLI::ExistingDirectory);
 
+    std::string batch_size_str{"512MB"};
+    app.add_option("--batch", batch_size_str, "Batch size of DB changes to accumulate before committing", true);
+
     CLI11_PARSE(app, argc, argv);
+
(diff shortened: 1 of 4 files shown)


AndreaLanfranchi:
In my personal opinion there's more work to do on this matter.
I understand having the same signature for common methods is useful but I don't think is acceptable to have only `EnvConfig` as parameter. We should normalize stages passing **all** useful informations and making them suitable to fit in a stage loop (with a common transaction.

AndreaLanfranchi:
Also we should agree on common methods naming: eg `forward` and [`unwind`|`rewind`]

yperbasis:
> In my personal opinion there's more work to do on this matter.
> I understand having the same signature for common methods is useful but I don't think is acceptable to have only `EnvConfig` as parameter. We should normalize stages passing **all** useful informations and making them suitable to fit in a stage loop (with a common transaction.

Yes, absolutely, more work and more parameters will be required. For instance, I'm going to add `external_txn` (useful we we only have a handful of blocks and want to do several stages inside a single transaction). This PR is just a stepping stone.

AndreaLanfranchi (approved):
I approve with reserve: not happy to loose command line arguments specifically the one which could set the batchsize

PR #326, opened 2021-07-26
Mdbx as submodule (Release v0.10.2)

yperbasis:
Include mdbx as a submodule (https://github.com/erthink/libmdbx).
Update to [Release v0.10.2](https://github.com/erthink/libmdbx/releases/tag/v0.10.2).


db/silkworm/db/access_layer_test.cpp
```diff
@@ -127,11 +127,6 @@ namespace db {
         db_config.exclusive = true;
         REQUIRE_NOTHROW(env = db::open_env(db_config));
         env.close();
-
-        // Check readers
-        db_config.create = false;
-        REQUIRE_NOTHROW(env = db::open_env(db_config));
-        env.close();
```
AndreaLanfranchi: Why this ?
Never happened before (in my env). Can you elaborate ?

yperbasis: > Why this ?
> Never happened before (in my env). Can you elaborate ?

When running tests under AddressSanitizer, mdbx v0.10.2 now throws `MDBX_TOO_LARGE` if DB is very big (with is the case with inmemory=false).

AndreaLanfranchi: Hmm ... under address Sanitizer throws and w/o ASAN does not ?
That does not sound good.
Besides Erigon does same checl_readers

AndreaLanfranchi: See https://github.com/erthink/libmdbx/issues/73 where it reports ASAN could generate false positives
Maybe worth opening an issue on libmdbx repo rather than removing the code

yperbasis: I believe it was done intentionally. From v0.10.2 release notes: "Raise MDBX_TOO_LARGE under Valgrind/ASAN if being opened DB is 100 larger than RAM (to avoid hangs and OOM)."

Previously tests with `inmemory=false` did pass even under ASAN, but were very slow.

AndreaLanfranchi: Not sure if in the end we keep it or not.
This is useful to remove stale readers.

AndreaLanfranchi (approved):

PR #328, opened 2021-07-27
Support external transaction in staged sync functions

yperbasis:
`external_txn` is useful for running several stages on a handful of blocks atomically.
This functionality exists in Erigon.

Also fixes a couple of bugs:
- `stage_senders` didn't commit the transaction
- `history_index_stage` started from a wrong block number


db/silkworm/stagedsync/stage_hashstate.cpp
```diff
@@ -323,15 +333,21 @@ void hashstate_unwind(mdbx::txn& txn, uint64_t unwind_to, HashstateOperation ope
     }
 
     (void)db::for_each(changeset_table, unwind_func);
-
 }
 
-StageResult unwind_hashstate(db::EnvConfig db_config, uint64_t unwind_to) {
+StageResult unwind_hashstate(db::EnvConfig db_config, uint64_t unwind_to, mdbx::txn* external_txn) {
```
AndreaLanfranchi, line 338: I'd have approached the matter in a different way.
Two different function signatures for function
* version A with db_config as parameter which eventually opens a transaction and calls B and commits on return)
* version B with txn as parameter which never commits

Higher functions either call A or B on behalf of how wide the transaction should be.
Also I'd change db_config to env

yperbasis: I've reworked this PR. Now there's a new helper class, TransactionManager, that encapsulates the differences between the normal mode and the external transaction mode.

cmd/blockhashes.cpp
```diff
@@ -42,7 +42,10 @@ int main(int argc, char* argv[]) {
 
     try {
         db::EnvConfig db_config{chaindata};
-        auto result_code{stagedsync::stage_blockhashes(db_config)};
+        auto env{db::open_env(db_config)};
+        stagedsync::TransactionManager tm{env};
+        auto data_dir{DataDirectory::from_chaindata(chaindata)};
+        auto result_code{stagedsync::stage_blockhashes(tm, data_dir.get_etl_path())};
```
AndreaLanfranchi, line 48: I'd rely more on DataDirectory as :

```
        auto data_dir{DataDirectory::from_chaindata(chaindata)};
        data_dir.create_tree();
        db::EnvConfig db_config{data_dir.get_chaindata_path().string()};
        auto env{db::open_env(db_config)};
        stagedsync::TransactionManager tm{env};
        auto result_code{stagedsync::stage_blockhashes(tm, data_dir.get_etl_path())};
        check_stagedsync_error(result_code);
```

This will make easier to switch back command line argument --chaindata to --datadir

db/silkworm/stagedsync/transaction_manager.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    void commit() {
+        if (external_txn_ == nullptr) {
+            managed_txn_.commit();
+            managed_txn_ = env_->start_write();  // renew transaction
+        } else {
+            // external_txn is useful for running several stages on a handful of blocks atomically.
+            // The code that invokes the stages is responsible for committing external_txn_ later on.
+        }
+    }
```
AndreaLanfranchi, line 53: I don't like the idea this mettod silently returns if transaction is external.
Throwing an exception here helps locate coding errors.

yperbasis: > I don't like the idea this mettod silently returns if transaction is external.
> Throwing an exception here helps locate coding errors.

This method does nothing for the external transaction by design; it's not a coding error. The idea is to move bits of conditional logic like the one in https://github.com/ledgerwatch/erigon/blob/v2021.07.05/eth/stagedsync/stage_execute.go#L341 from multiple stages to a single central place.

db/silkworm/common/data_dir.hpp
```diff
@@ -18,18 +18,12 @@
...
 namespace silkworm {
 
 class DataDirectory {
   public:
-    DataDirectory(std::filesystem::path base_path, bool create = false);
-    DataDirectory(bool create = false)
-        : DataDirectory::DataDirectory(DataDirectory::get_default_storage_path(), create){};
-    DataDirectory(std::string base_path_str, bool create = false)
-        : DataDirectory::DataDirectory(std::filesystem::path(base_path_str), create){};
-    ~DataDirectory() = default;
```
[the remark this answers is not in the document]
yperbasis: > Don't understand why remove those overloaded constructors

Right, I've put them back.

db/silkworm/db/mdbx.cpp
```diff
@@ -37,10 +37,13 @@ ::mdbx::env_managed open_env(const EnvConfig& config) {
             throw std::runtime_error("Unable to locate " + db_file.string() + ". Must exist has been set");
         }
     } else {
+        // Strangely create_directories occasionally returns false on macOS,
+        // despite having created the directories successfully.
+        // As a workand, we first call fs::create_directories and then fs::exists
+        fs::create_directories(db_path);
```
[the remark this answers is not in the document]
yperbasis: > This is actually a misunderstanding I fell into as well.
> Problem is create_directories() returns true only if **all** directories in the tree have been created by the call.
> If, on the other hand, any of the intermediate paths in the tree already exists then the call returns false.
> Please note that we're using here the call version which is meant to throw if there's an error hence the check of return value is only intended as : true == "I've created all directories"; false == "Some or all of the paths in the tree already existed"

I didn't know that, thanks for the clarification. I've simplified the code accordingly.

cmd/blockhashes.cpp
yperbasis: > I'd rely more on DataDirectory as :
> 
> ```
>         auto data_dir{DataDirectory::from_chaindata(chaindata)};
>         data_dir.create_tree();
>         db::EnvConfig db_config{data_dir.get_chaindata_path().string()};
>         auto env{db::open_env(db_config)};
>         stagedsync::TransactionManager tm{env};
>         auto result_code{stagedsync::stage_blockhashes(tm, data_dir.get_etl_path())};
>         check_stagedsync_error(result_code);
> ```
> 
> This will make easier to switch back command line argument --chaindata to --datadir

Fair enough; done.

AndreaLanfranchi (approved):

Issue #297, opened 2021-07-16
Header downloader


yperbasis:
This figure shows the overall architecture: 
![image](https://user-images.githubusercontent.com/9080285/136594066-ed63d84c-5f34-4344-82f4-8f980df6991d.png)

One component/thread (BlockProvider) is devoted to replay to external requests.
One component (HeaderDownloader) is devoted to chain downloading.

This figure details the  HeaderDownloader structure:

![architecture](https://user-images.githubusercontent.com/9080285/136594134-ffcbfb5c-9518-44a2-98e5-4a2d98ba2e21.png)

mriccobene:
### RPC
We have a class for each rpc of the sentry interface. Each class has this responsibility: prepare & execute a rpc

![image](https://user-images.githubusercontent.com/9080285/127617084-af0c0af7-f715-4bc3-8766-be742e71bc63.png)

### Packets

Messages (=packets) are not specified in the proto of sentry interface, they are in the Erigon code. We have a struct for each packet:

![image](https://user-images.githubusercontent.com/9080285/127618224-f565ff40-f461-4f33-aa08-2f0908267999.png)

### Messages

We have messages that encapsulate:

- incoming packet pre-processing
- outgoing packet pre-processing

(header processing & requesting logic is elsewhere)

![image](https://user-images.githubusercontent.com/9080285/127618371-69ad7a72-8af9-473b-9451-d65d7bd4abff.png)

### Logic and "main"
The other classes are in the picture below:

![image](https://user-images.githubusercontent.com/9080285/127618619-9dda936a-7a50-40c3-9a39-abe67f5c6b45.png)

PR #331, opened 2021-07-30
Header downloader - step 1

mriccobene:
This partially implement #297

Here we have the "responder" part of the downloader: it is able to connect to the sentry and respond to external request.
Also we have almost all packets (with test) and rpcs.


Giulio2002:
I think you should change the stage1 directory name to downloader, i think we might want to reuse it to put all of the future work to be done on block bodies.

cmake/Hunter/core_packages.cmake
```diff
@@ -19,3 +19,5 @@ hunter_add_package(intx)
 hunter_add_package(Microsoft.GSL)
 hunter_add_package(nlohmann_json)
 hunter_add_package(ethash)
+hunter_add_package(gRPC)
+hunter_add_package(Protobuf)
```
yperbasis, line 23: These 2 should go into `extra_packages`, not `core_packages`. Silkworm core is compatible with WebAssembly, while gRPC and Protobuf are probably not.

AndreaLanfranchi: Yeah right.

yperbasis (changes requested):
I agree with @Giulio2002 : we should put stage 1 code alongside the rest of silkworm db. "silkworm db" is a bit unfortunate (and we should probably rename it) but basically "silkworm db" means everything else but "silkworm core" (Silkworm core primarily deals with the EVM execution logic and is compatible with WebAssembly, while "silkworm db" is not  compatible with WebAssembly).

For example, instead of adding
```
stages/stage1/messages/InboundGetBlockBodies.hpp
```
add something like
```
db/downloader/messages/inbound_get_block_bodies.hpp
```
Please spend some time ensuring that your code fits nicely into existing "silkworm db" code.  Please follow the conventions in https://github.com/torquem-ch/silkworm#code-style

As to protobuf-generated files, it's probably OK for them to sit in the `interfaces` directory, no need to move to db.

mriccobene:
Done:
- [x] moved code to silkworm/db
- [x] added build target for test in silkworm/db - all test (47) pass, 10 are about the downloader
- [x] fixed some code-style issues
- [x] removed c++20 dependencies - refactored the entire rlp::encode overload set (of the downloader) - tested with gcc 10.2

Todo:
- [ ] fix circle/ci issues
- [ ] fix MSVC build issues
- [ ] check clang compile/build

Future improvements of the code base:
- rework the rlp::encode function template overload set (remove template?) and avoid code duplication [encode(vector), encode(packet66)]
- add interfaces as submodule https://github.com/ledgerwatch/interfaces

Please consider this post that can help the reviewing process https://github.com/torquem-ch/silkworm/issues/297#issuecomment-889698577

appveyor.yml
```diff
@@ -32,7 +32,7 @@ before_build:
   - git submodule sync
   - git submodule update --init --recursive
   - SET INCLUDE=C:\Tools\vcpkg\installed\x64-windows\include;%INCLUDE%
-  - SET PATH=C:\Tools\vcpkg\installed\x64-windows\bin;%PATH%
+  - SET PATH=C:\Tools\vcpkg\installed\x64-windows\bin;C:\Strawberry\perl\bin;%PATH%
```
yperbasis, line 35: Why is perl required?

cmd/CMakeLists.txt
```diff
@@ -106,6 +106,12 @@ if(NOT SILKWORM_CORE_ONLY)
   add_executable(trie trie.cpp)
   target_link_libraries(trie PRIVATE silkworm_db CLI11::CLI11)
 
+  find_package(gRPC CONFIG REQUIRED)
```
yperbasis, line 109: No need to find gRPC since it's a public dependency of silkworm_db.

cmd/CMakeLists.txt
```diff
@@ -106,6 +106,12 @@ if(NOT SILKWORM_CORE_ONLY)
...
+  add_executable(download_headers download_headers.cpp)
+  target_link_libraries(download_headers PRIVATE silkworm_db silkworm_core CLI11::CLI11)
+  target_include_directories(download_headers PRIVATE ${CMAKE_SOURCE_DIR})
+  set_property(TARGET download_headers PROPERTY CXX_STANDARD 20)
```
yperbasis, line 113: Can we do w/o C++ 20 for download_headers?

appveyor.yml
mriccobene: It is required to build OpenSSL that is a dependency of gRPC

yperbasis (approved):

PR #332, opened 2021-08-01
Reuse receipts in block execution to reduce memory allocation

yperbasis:
A small performance optimization to reduce unnecessary memory deallocations with subsequent new allocations. Decreases the total execution time by about 2%.


yperbasis:
FYI, here's a perf report of entire mainnet block execution after this change:
[perf_report.txt](https://github.com/torquem-ch/silkworm/files/6916274/perf_report.txt)

Linux andrew-NUC8i7HNK 5.11.0-25-generic #27-Ubuntu SMP Fri Jul 9 23:06:29 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

core/silkworm/execution/execution.cpp
```diff
@@ -22,11 +22,9 @@
...
 namespace silkworm {
 
-std::pair<std::vector<Receipt>, ValidationResult> execute_block(const Block& block, StateBuffer& buffer,
-                                                                const ChainConfig& config,
-                                                                AnalysisCache* analysis_cache,
-                                                                ExecutionStatePool* state_pool,
-                                                                evmc_vm* exo_evm) noexcept {
+ValidationResult execute_block(const Block& block, StateBuffer& buffer, const ChainConfig& config,
+                               std::vector<Receipt>& receipts, AnalysisCache* analysis_cache,
+                               ExecutionStatePool* state_pool, evmc_vm* exo_evm) noexcept {
```
AndreaLanfranchi, line 27: Just a question: why not make execute_block a std::pair<std::vector<receipt>, ValidationResult> ?
If I read it correctly the reference to receipts is passed through all the stack where `ExecutionProcessor::execute_block` eventually clears it and reserve. I.e. basically is ExecutionProcessor::execute_block which creates the instance of vector.

yperbasis: > Just a question: why not make execute_block a std::pair<std::vector, ValidationResult> ?
> If I read it correctly the reference to receipts is passed through all the stack where `ExecutionProcessor::execute_block` eventually clears it and reserve. I.e. basically is ExecutionProcessor::execute_block which creates the instance of vector.

Well, yes, `ExecutionProcessor::execute_block` clears the receipts, but it doesn't call `shrink_to_fit`, so it avoids deallocating memory and benefits from its subsequent reuse.

AndreaLanfranchi (approved):

PR #333, opened 2021-08-03
Fix for_each & for_count

yperbasis:
`data` was not updated when calling the `WalkFunc`


AndreaLanfranchi:
Good catch (totally missed it).
Let me simplify a little bit

AndreaLanfranchi (approved):

PR #334, opened 2021-08-03
Execute deployment code with baseline rather than advanced interpreter

yperbasis:
This is a performance optimization. For one-off execution (such as deployment code) evmone baseline interpreter is generally faster. In my tests this change decreased the total execution time of all mainnet blocks by 3%.

### DIFF
--- a/core/silkworm/execution/evm.cpp
+++ b/core/silkworm/execution/evm.cpp
@@ -249,9 +249,10 @@ evmc::result EVM::execute(const evmc_message& msg, ByteView code, std::optional<
     if (exo_evm) {
         EvmHost host{*this};
         res = exo_evm->execute(exo_evm, &host.get_interface(), host.to_context(), rev, &msg, code.data(), code.size());
-    } else if (analysis_cache) {
+    } else if (code_hash != std::nullopt && advanced_analysis_cache != nullptr) {
         res = execute_with_default_interpreter(rev, msg, code, code_hash);
     } else {
+        // for one-off execution baseline interpeter is generally faster
         res = execute_with_baseline_interpreter(rev, msg, code);
     }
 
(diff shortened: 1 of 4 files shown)


core/silkworm/execution/evm.cpp
```diff
@@ -287,17 +288,12 @@ evmc_result EVM::execute_with_baseline_interpreter(evmc_revision rev, const evmc
...
-        analysis = analysis_cache->get(*code_hash, rev);
-        if (!analysis) {
-            analysis = std::make_shared<evmone::AdvancedCodeAnalysis>(evmone::analyze(rev, code.data(), code.size()));
-            analysis_cache->put(*code_hash, analysis, rev);
-        }
-    } else {
-        // don't cache deployment code
+    assert(code_hash != std::nullopt && advanced_analysis_cache != nullptr);
+
+    std::shared_ptr<evmone::AdvancedCodeAnalysis> analysis{advanced_analysis_cache->get(*code_hash, rev)};
```
chfast, line 293: This upgrades weak pointer to shared pointer?

core/silkworm/execution/evm.cpp
```diff
@@ -249,9 +249,10 @@ evmc::result EVM::execute(const evmc_message& msg, ByteView code, std::optional<
     if (exo_evm) {
         EvmHost host{*this};
         res = exo_evm->execute(exo_evm, &host.get_interface(), host.to_context(), rev, &msg, code.data(), code.size());
-    } else if (analysis_cache) {
+    } else if (code_hash != std::nullopt && advanced_analysis_cache != nullptr) {
```
chfast, line 252: When is `advanced_analysis_cache` not null?

yperbasis: It's not null for mainnet block execution (`cmd/execute`), but it's null for the consensus tests (`cmd/consensus`).

I haven't implemented analysis cache for baseline interpreter yet (on my TODO list). Then it'll be interesting to compare baseline & advanced interpreter for mainnet block execution when they both have analysis caching. When I compare advanced interpreter with analysis cache against baseline interpreter w/o one, advanced interpreter wins.

AndreaLanfranchi (approved):

PR #343, opened 2021-08-05
Make collector's paths as std::filesystem::path

### LINKED ISSUE #342: Treat paths as std::filesystem:path
Need to treat paths as effective paths instead of passing them as string or char pointers

AndreaLanfranchi:
Make collector's paths as std::filesystem::path and delete directory only if owned
Fixes #342

### DIFF
--- a/db/silkworm/etl/collector.hpp
+++ b/db/silkworm/etl/collector.hpp
@@ -37,9 +37,9 @@ class Collector {
     Collector& operator=(const Collector&) = delete;
 
     explicit Collector(const std::filesystem::path work_path, size_t optimal_size = kOptimalBufferSize)
-        : work_path_managed_{true}, work_path_{set_work_path(work_path)}, buffer_{optimal_size} {}
+        : work_path_managed_{false}, work_path_{set_work_path(work_path)}, buffer_{optimal_size} {}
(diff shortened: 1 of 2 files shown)


AndreaLanfranchi:
Fixes #342

db/silkworm/etl/collector.hpp
```diff
@@ -36,8 +36,10 @@ class Collector {
     Collector(const Collector&) = delete;
     Collector& operator=(const Collector&) = delete;
 
-    explicit Collector(const char* work_path = nullptr, size_t optimal_size = kOptimalBufferSize)
-        : work_path_{set_work_path(work_path)}, buffer_{optimal_size} {}
+    explicit Collector(const std::filesystem::path work_path, size_t optimal_size = kOptimalBufferSize)
+        : work_path_managed_{true}, work_path_{set_work_path(work_path)}, buffer_{optimal_size} {}
```
yperbasis, line 40: Shouldn't `work_path_managed_` be the other way round? I'd think that it should be true when not specified externally and created internally, and false when specified externally.

AndreaLanfranchi: Damn you're right. I need a couple of days off

yperbasis (approved):

PR #345, opened 2021-08-06
Don't copy code from DB to IntraBlockState

yperbasis:
Performance optimization. Reduces execution time by about 2%.

Continuation of PR #329.


AndreaLanfranchi (approved):
Please ... on std::optional let's try to use .has_value() for bool checks.
if (!some_var) is confusing

PR #352, opened 2021-08-10
Test senders

Giulio2002:
For some reason the other PR was not updating commits


Giulio2002:
The Unwind senders was also simplified

node/silkworm/stagedsync/util.hpp
```diff
@@ -43,6 +44,7 @@ enum class [[nodiscard]] StageResult {
 
 void check_stagedsync_error(StageResult code);
 
+void unwind_table_from(mdbx::cursor& table, Bytes& starting_key);
```
yperbasis, line 47: I would rename it to something like `truncate_table_from`

node/silkworm/db/access_layer.cpp
```diff
@@ -17,6 +17,7 @@
 #include "access_layer.hpp"
 
 #include <cassert>
+#include <iostream>
```
yperbasis, line 20: An unnecessary include

yperbasis (approved):

PR #361, opened 2021-08-13
StateBuffer -> State, MemoryBuffer -> InMemoryState

yperbasis:
InMemoryState is not a buffer to something; it holds entire state in memory.


canepat (commented):
Probably also interface `StateBuffer` should be renamed (`State`?)

yperbasis:
> Probably also interface `StateBuffer` should be renamed (`State`?)

Yes, makes sense.

PR #365, opened 2021-08-15
Use of transaction manager in unwind_execution test

Giulio2002:


Giulio2002:
now it's more standard

yperbasis (commented):
Well, this variant works, but the previous one was also fine. In the previous variant `TransactionManager` was created in the external transaction mode, implying that `stagedsync::unwind_execution()` doesn't actually commit the transaction. In the new variant `stagedsync::unwind_execution()` does commit and then `TransactionManager` renews the transaction. Both variants are valid, so not sure what we achieve with this change.

Giulio2002:
@yperbasis In my opinion, it is more clear now, before we needed to make a transaction manager to be passed to unwind_execution in the middle of the code and not at the beggining, thus i think now it's more standard compared to the others unwind tests

yperbasis (approved):
OK, fine, I'm happy with the new variant as well.

Issue #295, opened 2021-07-16
Run Sanitizers in CI


yperbasis:
[ASAN](https://clang.llvm.org/docs/AddressSanitizer.html) & [UB](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) sanitizers are super useful and help with bringing C++ safety to that of Rust, provided good [code coverage](https://app.codecov.io/gh/torquem-ch/silkworm). We should enable them in our Continuous Integration.

We can do something similar to `build-gcc-sanitizers` & `build-clang-sanitizers` in https://github.com/ethereum/evmc/blob/master/circle.yml

Also, MSan or valgrind might be a good idea (see https://github.com/ethereum/evmone/pull/370) unless it's prohibitively slow.

[ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html) would also be super-useful.

PR #380, opened 2021-08-19
Added pruned execution

Giulio2002:


node/silkworm/stagedsync/util.cpp
```diff
@@ -44,7 +44,11 @@ std::pair<Bytes, Bytes> convert_to_db_format(const ByteView& key, const ByteView
 }
 
 void truncate_table_from(mdbx::cursor& table, Bytes& starting_key, bool reverse) {
-    if (table.lower_bound(db::to_slice(starting_key), false)) {
+    auto current{table.lower_bound(db::to_slice(starting_key), false)};
```
yperbasis, line 47: Please document what truncate table is supposed to do in the reverse case. When `reverse` is false, it deletes all entries whose `key>= starting_key`, correct? What is it supposed to do when `reverse` is true?

node/silkworm/stagedsync/prune_execution_test.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // This contract initially sets its 0th storage to 0x2a
+    // and its 1st storage to 0x01c9.
+    // When called, it updates its 0th storage to the input provided.
+    Bytes contract_code{*from_hex("600035600055")};
+    Bytes deployment_code{*from_hex("602a6000556101c960015560068060166000396000f3") + contract_code};
+
+    block.transactions.resize(1);
+    block.transactions[0].data = deployment_code;
+    block.transactions[0].gas_limit = block.header.gas_limit;
+    block.transactions[0].max_priority_fee_per_gas = 0;  // EIP-1559
```
yperbasis, line 69: Please change this test similar to my changes in PR #381. Either use `kLondonTestConfig` and leave EIP-1559 transactions (what I did in `unwind_execution_test`), or keep legacy transactions and equate `max_priority_fee_per_gas` with `max_fee_per_gas` (what I did in `history_index_test`).

Giulio2002:
code updated

node/silkworm/stagedsync/util.cpp
yperbasis: @Giulio2002 documentation for truncate_table_from is still missing. From the code I take it that it  deletes all entries whose `key >= starting_key` when reverse is false, and all entries whose `key < starting_key` when reverse is true. Is that the intention? It should be documented.

node/silkworm/stagedsync/util.cpp
```diff
@@ -44,7 +44,11 @@ std::pair<Bytes, Bytes> convert_to_db_format(const ByteView& key, const ByteView
...
+    if (reverse) {
+        current = table.to_previous(false);
```
yperbasis, line 49: What if the cursor is already at the very first position? Will `to_previous` silently do nothing? Better add a unit case.

yperbasis (changes requested):
Please add documentation for `truncate_table_from`

node/silkworm/stagedsync/util.cpp
Giulio2002: to_previous should return false and current will have nothing it, thus avoiding the entire loop, i can add some unit tests for it though

node/silkworm/stagedsync/util.cpp
Giulio2002: so the thing is this, there is no such thing as an upper_bound function, so my solution for whenever we need to reverse is we take key+1 and go to previous element so that we are at the upper bound

Giulio2002: forgot to address this sorry

Giulio2002:
ok now tests are here @yperbasis

node/silkworm/stagedsync/util.cpp
```diff
@@ -43,8 +43,19 @@ std::pair<Bytes, Bytes> convert_to_db_format(const ByteView& key, const ByteView
     return {a, b};
 }
 
+/*
+ * Truncate a table by starting from a certain key, either up or down.
+ * If reverse is set to false, then every key either higher or equal to starting key will be erased.
+ * If reverse is set to true, then every key either lower or equal to starting key will be erased.
```
yperbasis, line 49: Well, this is not what the function does, is it? If reverse is set to true, then every key strictly lower than `starting_key` will be erased, not `<= starting_key`.

I'd also remove or rephrase the NOTE. The fact that we use `cursor. lower_bound()` is just an implementation detail, not relevant to the API. We could implement this function differently and achieve `<= starting_key` for the reverse case with something like
```
if (current && current.key == db::to_slice(starting_key)) {
    table.erase(/*whole_multivalue=*/true);
}
```

yperbasis (approved):

node/silkworm/stagedsync/util.cpp
Giulio2002: ok, i fixed the comments as you said

yperbasis (changes requested):
Hold on a sec, ASAN is complaining for me, let's investigate.

yperbasis (approved):
Ok, fixed now. Code like
```
auto value{db::to_slice(db::block_key(0))};
```
is problematic because it takes a view of a temporary byte string, which is deleted at the end of the line.

We should enable Address Sanitizer in CI really (Issue #295).

PR #388, opened 2021-08-23
remove unnecessary length_ member in Buffer

greg7mdp:
I think this should not change the behavior.


yperbasis (approved):

node/silkworm/etl/buffer.cpp
```diff
@@ -22,25 +22,17 @@ namespace silkworm::etl {
...
+    buffer_.push_back(entry);
 }
 
-void Buffer::sort() { std::sort(buffer_.data(), buffer_.data() + length_); }
+void Buffer::sort() { std::sort(buffer_.begin(), buffer_.end()); }
 
 size_t Buffer::size() const noexcept { return size_; }
 
-gsl::span<const Entry> Buffer::entries() const noexcept { return {buffer_.data(), length_}; }
+    gsl::span<const Entry> Buffer::entries() const noexcept { return {buffer_.data(), buffer_.size()}; }
```
AndreaLanfranchi, line 32: I believe with this mod the gsl::span is not needed and we could return a direct const reference to buffer_ to iterate the items.

AndreaLanfranchi (changes requested):

node/silkworm/etl/buffer.cpp
yperbasis: > I believe with this mod the gsl::span is not needed and we could return a direct const reference to buffer_ to iterate the items.

True, but `gsl::span` is still valid, so it is just a matter of preference. I am happy with either. In C++20 span is in the standard (`std::span`).

AndreaLanfranchi: As here we don't return a subset of items returning a non-owning container with the same set of items of the original container does not make any sense semantically. Imho

greg7mdp: Yes I agree with @AndreaLanfranchi , since we mean to return the whole vector it is just more natural (and more efficient) to return a reference to the vector itself, rather than calculating the vector's length and constructing a new object. I have checked in the change.

AndreaLanfranchi (approved):

greg7mdp:
As a summary:
- removed unnecessary length_ member
- moved function to header file so they can be inlined for better performance
- added version of put() accepting rvalues to avoid unnecessary etl::Entry copies when possible
- added the (modified) test that Andrea provided. I removed the sort() so we were testing only the Buffer filling time, and I added a version with the move that destroys the source.

greg7mdp:
move is faster :-)
image

node/silkworm/etl/buffer.cpp
```diff
@@ -20,30 +20,5 @@
 
 namespace silkworm::etl {
```
AndreaLanfranchi, line 21: I believe buffer.cpp can be safely removed.

AndreaLanfranchi (changes requested):

node/silkworm/etl/buffer.cpp
greg7mdp: yes, will do

AndreaLanfranchi:
Sorry to bother again. Please rename cmd/buffer executable to bench_buffer so we'll use bench_ prefix for all this tiny tools 
After that I believe is ready to merge

greg7mdp:
> Sorry to bother again. Please rename cmd/buffer executable to bench_buffer so we'll use bench_ prefix for all this tiny tools
> After that I believe is ready to merge

should I rename also the buffer.cpp file to bench_buffer.cpp? I think it would be better.

AndreaLanfranchi:
Yes please

greg7mdp:
unrelated, but I upgraded to cmake 3.21.1 (for an unrelated project) and now I have the following issue when rebuilding the cmake cache:

```
1> [CMake] C:\greg\.hunter\_Base\6c9b2bc\61b5df2\9d920f1\Build\Boost\Source>bootstrap.bat
1> [CMake] Building Boost.Build engine
1> [CMake] 'C:/Program' is not recognized as an internal or external command,
1> [CMake] operable program or batch file.
1> [CMake] File Not Found
1> [CMake] 
1> [CMake] Failed to build Boost.Build engine.
1> [CMake] Please consult bootstrap.log for further diagnostics.
1> [CMake] 
1> [CMake] ninja: build stopped: subcommand failed.
1> [CMake] 
1> [CMake] [hunter ** FATAL ERROR **] Build step failed (dir: C:/greg/.hunter/_Base/6c9b2bc/61b5df2/9d920f1/Build/Boost
1> [CMake] [hunter ** FATAL ERROR **] [Directory:C:/greg/.hunter/_Base/Download/Hunter/0.23.288/6c9b2bc/Unpacked/cmake/projects/Boost]
```

AndreaLanfranchi:
> unrelated, but I upgraded to cmake 3.21.1 (for an unrelated project) and now I have the following issue when rebuilding the 

Ninja generator has issues with boost libraries, besides on Windows and using MSVC I always use `"generator": "Visual Studio 16 2019 Win64"` in CMakeSettings.json ... I believe has something to do with boost's b2 not reading properly the pathname (does not accept spaces)

AndreaLanfranchi (approved):

Issue #195, opened 2021-03-23
Etl load inefficiency ?


AndreaLanfranchi:
For tracking purposes.

Etl load() method exhibit a great inefficiency when is granted all items are properly sorted across **all** flushed buffers.
This case is what sender's recovery implements: items are collected in the very same order they will be inserted into db.
In this case scenario there is no need to go through a priority queue traversing all flushed buffers to check which one has the lower key on every item being processed.

See https://github.com/torquem-ch/silkworm/blob/1d454b11c9efba66ac81eb3b39c7c9ed67aefc36/db/silkworm/etl/collector.cpp#L102-L157

greg7mdp:
@AndreaLanfranchi not sure how to test the speed, but the following changes may help:
https://github.com/greg7mdp/silkworm/commit/cacbeb7b3db4bbd76f70b8951912519fd31e9055

greg7mdp:
@AndreaLanfranchi I have made another change using `std::map` instead of `std::priority_queue`. I think it may well be faster than using a vector-based priority_queue, but again I don't know how to test it. I have built the code but not tested it.

AndreaLanfranchi:
Hi @greg7mdp,
sorry for late response.
First of all thank you for spending time on the issue. 

I personally think the difference of priority_queue vs map (in this case) is negligible.  At most the container will hold 1 element for each filecollector so the numbers are really low. However a good instrumentation could clear any doubt: unfortunately current Silkworm master branch is off from erigon wrt data paths. I'd suggest to wait till we merge #383 and will eventually show you how to instrument this.

Besides this issue is actually slightly different ... if we know keys are already sorted in every file buffer there is no need for a priority queue nor for a map.

Off topic question : I understand you're using Visual Studio. Did you experience any problem with latest updates ? (I personally can't have the whole project load correctly since update to 16.11.1)

Thank you

greg7mdp:
No problem @AndreaLanfranchi. Yes it would be good to be able to reproduce the issue you mentioned, to validate if any change indeed fixes it.

One issue I corrected in my first checkin was that the comparison function was taking values instead of references. This would cause string copies (and memory allocations) for every comparison, which is inefficient.

> At most the container will hold 1 element for each filecollector so the numbers are really low.

Then the fact that it would be slow is really surprising?

> if we know keys are already sorted in every file buffer there is no need for a priority queue nor for a map

that's true, but you have to make a special case which should not be necessary.

[...]

PR #391, opened 2021-08-24
issue #195 - speedup etl by avoiding unnecessary copies

greg7mdp:
I haven't tested it besides running core_test and node_test, but it is simple enough that it should work and avoid unnecessary etl::Entry copies.
Each etl::Entry copy will cause two `malloc()` when allocating the new std::basic_string, two data copies,  and two `free()` when the original is destructed. Doing a move avoids all this.

### DIFF
--- a/node/silkworm/etl/collector.cpp
+++ b/node/silkworm/etl/collector.cpp
@@ -57,6 +57,14 @@ void Collector::collect(const Entry& entry) {
     }
 }
 
+void Collector::collect(Entry&& entry) {
+    buffer_.put(std::move(entry));
+    ++size_;
+    if (buffer_.overflows()) {
+        flush_buffer();
+    }
+}
+
 void Collector::load(mdbx::cursor& target, LoadFunc load_func, MDBX_put_flags_t flags, uint32_t log_every_percent) {
     const auto overall_size{size()};  // Amount of work
 
--- a/node/silkworm/etl/collector.hpp
+++ b/node/silkworm/etl/collector.hpp
@@ -44,6 +44,7 @@ class Collector {
     ~Collector();
 
     void collect(const Entry& entry);  // Store key-value pair in memory or on disk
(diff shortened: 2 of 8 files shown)


Giulio2002:
Master has following performance: 

``` Done in 29.354s

 [Buffer-Default] First loop ...
 [Buffer-Default copy build] Done in 898.685ms
 [Buffer-Default] Second loop ...
 [Buffer-Default copy build] Done in 606.608ms


 [Buffer-Default] First loop ...
 [Buffer-Default move build] Done in 66.327ms
 [Buffer-Default] Second loop ...
 [Buffer-Default move build] Done in 32.103ms

```

Now we achieve:

```
 Feeding buffer base ...
 Done in 29.137s

 [Buffer-Default] First loop ...
 [Buffer-Default copy build] Done in 865.616ms
 [Buffer-Default] Second loop ...
 [Buffer-Default copy build] Done in 716.927ms


 [Buffer-Default] First loop ...
 [Buffer-Default move build] Done in 68.371ms
 [Buffer-Default] Second loop ...
 [Buffer-Default move build] Done in 25.281ms
```

Giulio2002 (approved):

Giulio2002:
LGTM @AndreaLanfranchi, @yperbasis, anything to say here?

greg7mdp:
@Giulio2002 that test you ran (bench_buffer.cpp) does not measure the change from the PR. It tests only the buffer itself, not inserting rvalue entries into a collector.

AndreaLanfranchi:
Good for me too. Would be super-nice if you could also conver in collector_test.cpp the lines highlighted for codecoverage

greg7mdp:
I think this last change should do it @AndreaLanfranchi

AndreaLanfranchi (approved):
Nice ! Ty @greg7mdp

PR #393, opened 2021-08-24
Header downloader step1 - Improve responsibility partitioning

mriccobene:
This PR aims at improved responsibility partitioning and improved relationships between components. 

It separates block providing and headers downloading. Block providing is a name given to the code that process inbound request of block headers or block bodies from other peers (suggestions for better names are welcome). This code should always be running.

Doing this simplifies code and enable us to gives to stage1 a shape similar to other stages (essentially an interface with two functions: wind and unwind).

Also the threading model is visible in the "main" (download_headers.cpp).

Note: some class names have changed but the files have not been renamed for ease of review. In the next PR they will be aligned.


node/silkworm/downloader/ActiveComponent.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * has also other things to do).
+ * Here we prefer not to provide a thread facility and let the user provide one more suitable to the context,
+ * so perhaps a better name is LongRunningComponent.
+ */
+
+class ActiveComponent {
+  public:
+    virtual void execution_loop() = 0;
+
+    void need_exit() { exiting_.store(true); }
```
yperbasis, line 36: I'd rename this method to `stop`

node/silkworm/downloader/ActiveComponent.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    bool exiting() { return exiting_.load(); }
```
yperbasis, line 38: I'd rename this method to `stopping`  and make it `const`.

yperbasis (approved):

node/silkworm/downloader/ActiveComponent.hpp
mriccobene: ok! I will do it in the next PR

node/silkworm/downloader/ActiveComponent.hpp
mriccobene: ok! I will do it in the next PR

node/silkworm/downloader/ActiveComponent.hpp
yperbasis: @mriccobene I've already done it in PR #401, could you review it please.

PR #394, opened 2021-08-25
added benchmarks for ETL

Giulio2002:


Giulio2002:
CC @greg7mdp

greg7mdp:
Hi @Giulio2002 , thanks for the test, but I think the problem referred to in issue #195 is with collector.cpp, not with `Buffer`. I don't think `Buffer` would care about the ordering because it inserts the entries in the order received and doesn't care about their content.
Let me see if I can use inspiration from your code to write a similar benchmark for `Collector`.

Giulio2002:
yeah sure go ahead, for the sake of simplicity you can commit on my branch

AndreaLanfranchi:
> Transform will never tamper with key prefixes.

It seems to me here [we do indeed tamper with extracted keys](https://github.com/torquem-ch/silkworm/blob/master/node/silkworm/stagedsync/stage_log_index.cpp#L38-L62)

> if pretty much all data will be sorted we wont need ETL anyway

That's not completely true : etl acts as a separator amongst reads and writes into db. If we interleave reads and writes performance degrades

Giulio2002:
we tamper the suffix not the prefix, so nothing happens there in terms of order. regarding the sorting we could put a boolean in the load function that prevents sorting in this specific case.

node/silkworm/etl/collector.hpp
```diff
@@ -57,10 +56,11 @@ class Collector {
     void load(mdbx::cursor& target, LoadFunc load_func = nullptr,
               MDBX_put_flags_t flags = MDBX_put_flags_t::MDBX_UPSERT, uint32_t log_every_percent = 100u);
 
+    void load_serialized(mdbx::cursor& target, LoadFunc load_func = nullptr,
```
yperbasis, line 59: What's the difference between `load` and `load_serialized`?

AndreaLanfranchi: Load serialized consumes one flushed after the other (wuthout passing from priority queue)
Normal load instead pulls an item from each flushed file and picks the lowest key to be processed

AndreaLanfranchi: Anyway this issue, imo, is very low priority and I would not tamper too much with etl load for the moment

Giulio2002:
#404 is taking care of the benchmarks

PR #401, opened 2021-08-29
Gather concurrency things in silkworm/concurrency

yperbasis:
Move `ActiveComponent`, `ThreadSafeQueue`, and `Worker` into silkworm/concurrency. More consistent function names between `ActiveComponent` and `Worker`.


AndreaLanfranchi (commented):
I believe active component can be easily implemented into worker scope

mriccobene (approved):
The previous ActiveComponent methods nomenclature expressed the desire for any threads to terminate since a clean exit cannot be imposed on a thread. But I realize that the names were unusual.

PR #402, opened 2021-08-30
Pruned log index stage

Giulio2002:


Giulio2002:
#396

node/silkworm/stagedsync/stage_log_index.cpp
```diff
@@ -215,4 +215,59 @@ StageResult unwind_log_index(TransactionManager& txn, const std::filesystem::pat
...
+            // check if prune can be applied
+            if (bm.maximum() >= prune_from) {
+                // Erase elements that are below prune_from
+                bm &= roaring::Roaring(roaring::api::roaring_bitmap_from_range(prune_from, last_processed_block + 1, 1));
+                Bytes new_bitmap(bm.getSizeInBytes(), '\0');
+                bm.write(byte_ptr_cast(&new_bitmap[0]));
+                // generates new key
+                Bytes new_key(key.size(), '\0');
+                std::memcpy(&new_key[0], key.data(), key.size());
+                endian::store_big_u32(&new_key[new_key.size() - 4], UINT32_MAX);
```
yperbasis, line 246: What's the logic with `UINT32_MAX` here?

Giulio2002: "mis-pasting" on my part and i did not include a case for it

Giulio2002: now i fixed it

yperbasis: Well, `UINT32_MAX` is still here. If it's correct, please add more comments - otherwise the logic is not immediately clear.

node/silkworm/stagedsync/stage_history_index.cpp
```diff
@@ -239,7 +239,6 @@ StageResult history_index_prune(TransactionManager& txn, const std::filesystem::
                 // generates new key
                 Bytes new_key(key.size(), '\0');
                 std::memcpy(&new_key[0], key.data(), key.size());
-                endian::store_big_u32(&new_key[new_key.size() - 4], UINT32_MAX);
                 // replace with new index
                 etl::Entry entry{new_key, new_bitmap};
```
yperbasis, line 243: `new_key` is the same as `key`, and you don't actually need it. You can do something like
```
etl::Entry entry{Bytes{key}, new_bitmap};
```

yperbasis (changes requested):

node/silkworm/stagedsync/stage_history_index.cpp
Giulio2002: ops, forgot to git add it

yperbasis (approved):

PR #408, opened 2021-08-31
Fix keys in history_index_unwind & unwind_log_index

yperbasis:
As far as I understand it, this fixes the issue similar to the one in PR #402.


Giulio2002:
This logic in unwind is correct

Giulio2002:
It is not a mistake we need to make sure that the non-deleted bitmaps but modified ones are always "last"... that just won't apply to pruning because we "unwind" downwards

Giulio2002:
In anyway bitmap with a minimum too high are wiped off so the ones that just are modified are always the last ones

PR #410, opened 2021-09-01
Header downloader step1 - Sync gRPC model

mriccobene:
**This PR reverts to a synchronous gRPC model that greatly simplify the code.**
It has been favored by the current partitioning of responsibilities but I think it can be applied regardless of that.

(the responsibilites are: 
- BlockProvider answers external requests accessing the database, 
- HeaderDownloader  (not present in this code yet) requests and processes headers, accessing the database to save the newly received ones; 

so the 2 world are partitioned and can run in independent threads)

The sync model is NOT recommended for performance sensitive SERVERS but not for clients. And silkworm is a client of the sentry. 
The sync model is LESS flexible than the asynchronous one but it seems adequate at the moment.

I think that if we will want to use the asynchronous model in the future we will be able to implement it with coroutines or futures leaving the code with the current form, which is simpler.

Futermore Erigon uses a synchronous model for what I can understand.

What do you think?

Here a diagram of the relevant classes and thread scopes.
image

In the PR the additions are more than the removals because I left the old file with the asynchronous gRPC abstractions, only to compare it with the sync one.


yperbasis:
Right, I think it makes sense to start with simpler synchronous code and get it working.

yperbasis (approved):

PR #414, opened 2021-09-03
Fix stage_hashstate for storage

yperbasis:
Our code has to be different from Erigon code because we don't have its `AutoDupSortKeysConversion` feature.

### DIFF
--- a/node/silkworm/db/access_layer.cpp
+++ b/node/silkworm/db/access_layer.cpp
@@ -17,6 +17,7 @@
 #include "access_layer.hpp"
 
 #include <cassert>
+
 #include <nlohmann/json.hpp>
 
 #include <silkworm/common/endian.hpp>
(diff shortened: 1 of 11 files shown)


Giulio2002:
ok, so you altered the db format for storage keys, in erigon is `address hash + incarnation + key hash` you made it become `address hash+incarnation`, so you put a load function which fixes the above but i do not see how that is different from just accessing it when we extract data, in other words what changed with what the code was before. Erigon does not seems to act like this either

Giulio2002:
one more thing, the contract we deploy in test has only one location not 2, it should just have a location 0 not a location 1

Giulio2002:
in the test you are just testing Address hash + incarnation not Address hash + incarnation + key hash

yperbasis:
> ok, so you altered the db format for storage keys, in erigon is `address hash + incarnation + key hash` you made it become `address hash+incarnation`, so you put a load function which fixes the above but i do not see how that is different from just accessing it when we extract data, in other words what changed with what the code was before. Erigon does not seems to act like this either

I haven't altered the DB format from that of Erigon. However, Erigon does special magic with `MDBX_DUPSORT` entries (see `AutoDupSortKeysConversion`), which we don't have in Silkworm. So the code in Erigon looks as if storage key were `address hash + incarnation + location hash`, but it's not true on the MDBX level. On the MDBX level storage key is `address hash + incarnation`, while storage (multi)value is `location hash + zeroless value`.

Giulio2002 (approved):

PR #416, opened 2021-09-05
added test_util format

Giulio2002:


node/silkworm/common/test_util.cpp
```diff
@@ -0,0 +1,39 @@
...
+    transactions[0].to = 0xe5ef458d37212a06e3f59d40c454e76150ae7c32_address;
+    transactions[0].value = 1'027'501'080 * kGiga;
+    transactions[0].data = {};
+    static_cast<void>(transactions[0].set_v(27));
+    transactions[0].r =
+        intx::from_string<intx::uint256>("0x48b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d495a36649353");
+    transactions[0].s =
+        intx::from_string<intx::uint256>("0x1fffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804");
+
+    transactions[1].type = Transaction::Type::kEip1559;
```
AndreaLanfranchi, line 22: What if we add this transaction to a block number which is pre-Eip1559 ?

Giulio2002: we could just exclude and not add this specific transaction

yperbasis (approved):

PR #417, opened 2021-09-05
fixed imprecision in boundary hash

Giulio2002:


core/silkworm/types/block.cpp
```diff
@@ -37,6 +37,9 @@ ethash::hash256 BlockHeader::boundary() const {
 
     if (difficulty > 1u) {
         auto result{intx::bswap(dividend / difficulty)};
+        if ((difficulty & (difficulty - 1)) == 0) {
+            result += 1;
+        }
```
AndreaLanfranchi, line 42: This is not correct as you're adding 1 **after** the byte swapping while instead you should add 1 to the quotient.
```suggestion
        unsigned int add_one{(difficulty & (difficulty - 1)) == 0 ? 1u : 0};
        auto result{intx::bswap((dividend / difficulty) + add_one)};
```

Also I realized `if (difficulty > 1u)` should be instead `if (difficulty >= 1u)`

Please also add tests

AndreaLanfranchi (changes requested):

core/silkworm/types/block.cpp
Giulio2002: ops, i misunderstood what you said in the issue, test are coming shortly

Giulio2002:
now i added tests to it

core/silkworm/types/block_test.cpp
```diff
@@ -188,4 +189,36 @@ TEST_CASE("EIP-1559 Header RLP") {
...
+    auto expected_boundary{intx::bswap(
+        (intx::from_string<intx::uint256>("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff") /
+        h.difficulty) + 1u
+    )};
+    Bytes expected_boundary_view(kHashLength, '\0');
+    std::memcpy(&expected_boundary_view[0], intx::as_bytes(expected_boundary), kHashLength);
+
+    CHECK(boundary_view.compare(expected_boundary_view) == 0);
+}
+
```
AndreaLanfranchi, line 222: I see several problems here (sorry for being pedantic)

1. Please don't duplicate test cases : 1 test case ensemble one logic unit of test. If you need to test two (or more) different behaviors of the same unit use `SECTION`. See [examples](https://github.com/catchorg/Catch2/blob/devel/docs/tutorial.md#test-cases-and-sections)
2. The writing of this test is not relieable: `expected_boundary` is computed exactly as `header.boundary()` does so there absolutely no chance expected boundary might result different from computed boudary. What we need instead is compare a well known boundary value (or a set of) with the outcome of `header.boundary()` and see if we've done the function correctly;
3. Variable naming : if you write `Bytes expected_boundary_view(kHashLength, '\0');` I should expect the initialized variable to be a `ByteView` not a `Byte`
4. For coverage we should also implement the test for difficulty <= 1u

AndreaLanfranchi (changes requested):

core/silkworm/types/block_test.cpp
Giulio2002: now all of the above they should good.  (do not worry about being pedantic, it is for the greater good)

core/silkworm/types/block.cpp
```diff
@@ -29,17 +29,21 @@ evmc::bytes32 BlockHeader::hash(bool for_sealing) const {
...
     static intx::uint256 dividend{
         intx::from_string<intx::uint256>("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")};
 
     ethash::hash256 ret{};
 
-    if (difficulty > 1u) {
-        auto result{intx::bswap(dividend / difficulty)};
-        std::memcpy(ret.bytes, intx::as_bytes(result), 32);
+    if (difficulty >= 1u) {
+        unsigned int add_one{(difficulty & (difficulty - 1)) == 0 ? 1u : 0u};
```
chfast, line 42: This is interesting way of fixing it, but did you consider to just use `auto dividend = uint320{1} << 256`?

core/silkworm/types/block_test.cpp
```diff
@@ -188,4 +188,28 @@ TEST_CASE("EIP-1559 Header RLP") {
     CHECK(decoded == h);
 }
 
+TEST_CASE("Hash header boundary computation") {
+    BlockHeader h;
+    h.difficulty = intx::from_string<intx::uint256>("0x13009de5666753258eb9306157680dc5da0d");
```
chfast, line 193: Intx has support for literals: `= 0x13009de5666753258eb9306157680dc5da0d_uint256`.

core/silkworm/types/block.cpp
```diff
@@ -29,17 +29,21 @@ evmc::bytes32 BlockHeader::hash(bool for_sealing) const {
...
+        auto result{intx::bswap((dividend / difficulty) + add_one)};
+        std::memcpy(&ret.bytes[0], intx::as_bytes(result), kHashLength);
```
chfast, line 44: intx has API for this: `intx::be::store<ethash::hash256>(result)`.

core/silkworm/types/block_test.cpp
Giulio2002: how to enable intx literals?

core/silkworm/types/block.cpp
Giulio2002: wouldn't that be bigger than an uint256? wouldn't that overflow?

Giulio2002: that would be 0x1000....0000 would it not?

Giulio2002: oh now i see

chfast: This is `2^256` as defined in the spec. Then you need to cast it back to `uint256`. This will not overflow if `difficulty >= 2`. For cases `0` and `1` you need to return the constant `0xffff...ffff`.

core/silkworm/types/block_test.cpp
chfast: I believe with `using namespace intx;`

Giulio2002: `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff_uint256` does not work

Giulio2002: it was _u256

core/silkworm/types/block_test.cpp
```diff
@@ -17,9 +17,12 @@
 #include "block.hpp"
 
 #include <catch2/catch.hpp>
-
+#include <iostream>
```
yperbasis, line 20: An unnecessary include

core/silkworm/types/block.cpp
```diff
@@ -23,25 +23,31 @@
...
-
-    ethash::hash256 ret{};
+    auto dividend{intx::uint320{1} << 256};
 
-    if (difficulty > 1u) {
-        auto result{intx::bswap(dividend / difficulty)};
-        std::memcpy(ret.bytes, intx::as_bytes(result), 32);
-    } else {
-        std::memcpy(ret.bytes, intx::as_bytes(dividend), 32);
+    if (difficulty >= 1u) {
```
yperbasis, line 41: Should be strictly greater than 1. Otherwise the result is 0 for `difficulty == 1`, which is wrong.

yperbasis (changes requested):

core/silkworm/types/block.cpp
```diff
@@ -23,25 +23,31 @@
...
+
 evmc::bytes32 BlockHeader::hash(bool for_sealing) const {
     Bytes rlp;
     rlp::encode(rlp, *this, for_sealing);
     return bit_cast<evmc_bytes32>(keccak256(rlp));
 }
 
+//! \brief Calculates header's boundary.
+//! \return Header's boundary.
+//! \remarks When difficulty is an exact power of 2, we need to adjust the quotient by one.
```
yperbasis, line 37: The remark is obsolete now, please remove.

core/silkworm/types/block.cpp
Giulio2002: it already is >=

yperbasis: It should be `>` instead of `>=`. Currently `BlockHeader::boundary()` returns 0 if difficulty == 1 (when you convert `intx::uint320{1} << 256` to `intx::uint256`, you get 0).

Giulio2002: ok, i misunderstood, now it is fixed

yperbasis (approved):

yperbasis:
LGTM now. @AndreaLanfranchi what do you think?

AndreaLanfranchi (approved):
Ok for me now

PR #422, opened 2021-09-07
more general lookup encoding

Giulio2002:
Might be useful in the future if we want to implement more lookups or efficient ways to store blocks... it could also be used for numerical keys since the encoding of lookups is little endian @AndreaLanfranchi . nonetheless, we will need this.


node/silkworm/stagedsync/stage_tx_lookup.cpp
```diff
@@ -75,9 +55,9 @@ StageResult stage_tx_lookup(TransactionManager& txn, const std::filesystem::path
         auto body_rlp{db::from_slice(bodies_data.value)};
         auto body{db::detail::decode_stored_block_body(body_rlp)};
         Bytes block_number_as_bytes(static_cast<uint8_t*>(bodies_data.key.iov_base), 8);
-        // we compact block number
-        auto lookup_block_data{compact(block_number_as_bytes)};
         block_number = endian::load_big_u64(&block_number_as_bytes[0]);
+        // we compact block number
+        auto lookup_block_data{db::encode_lookup(block_number)};
```
AndreaLanfranchi, line 60: I don't think this is correct. 
In original implementantion compaction is performed on **bigendian** representation of block_number pulled from body data:
```
        Bytes block_number_as_bytes(static_cast<uint8_t*>(bodies_data.key.iov_base), 8);
        // we compact block number
        auto lookup_block_data{compact(block_number_as_bytes)};
```
While here instead you pull block_number doing an endianness conversion using
```
block_number = endian::load_big_u64(&block_number_as_bytes[0]);
```
and then passing the number in little endian to your new function `encode_lookup`

If we assume a block number == 12345678u we have the production of data to be stored in lookup table as

- original implementation `0xbc614e`
- your new implementation `0x4e61bc` (note the bytes are swapped)

node/silkworm/db/util_test.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <vector>
+
+#include <catch2/catch.hpp>
+
+#include "util.hpp"
+
+namespace silkworm::db {
+
+TEST_CASE("block key") {
```
AndreaLanfranchi, line 25: Ok for these tests but the error depicted invalidates tx_lookup tests.

AndreaLanfranchi (changes requested):

node/silkworm/db/util.hpp
```diff
@@ -154,6 +154,10 @@ Bytes storage_prefix(ByteView address, uint64_t incarnation);
 // Erigon CanonicalHeadersKey / ReceiptsKey
 Bytes block_key(uint64_t block_number);
 
+// Erigon TxIndex value
+Bytes encode_lookup(uint64_t lookup);
```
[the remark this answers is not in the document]
Giulio2002: they encode lookups, lookups are generic numbers just "compacted". i was planning on doing some experimentation with new lookups tables and i just put it there because in the future it is likely we will add more of those anyway.

node/silkworm/stagedsync/stage_tx_lookup.cpp
Giulio2002: thank you, i am going to fix this

Giulio2002: now it is in big endian

AndreaLanfranchi: Though the result is correct you've introdced a big inefficiency.
Look:

1. You extract from data key the bytes representing the bigendian form of block number : `Bytes block_number_as_bytes(static_cast<uint8_t*>(bodies_data.key.iov_base), 8);`
2. you parse block number as little endian doing a byteswap by `block_number = endian::load_big_u64(&block_number_as_bytes[0]);
3. you pass `block_number` to `encode_lookup` which does again a byteswap of block_number and transforms it into a byte array (`block_key(lookup)`). **You're getting the very same value you had in step 1**
4. eventually the bytearray is compacted

You can easily see how steps 2 and 3 add unnecessary overhead.

Giulio2002: Removed the overhead for that specific case.

AndreaLanfranchi: Giulio ! ... sorry for being harsh but ... stop a second and **think**. Don't rush things with copy and paste... **think**.
Let's see what you've achieved with last commit :

1. Basically you've removed the only place where `encode_lookup` was used (not counting tests) hence voiding its presence. What you should have done instead was to create an overload for `encode_lookup` to receive an array of bytes
2. This line `block_number_as_bytes = block_number_as_bytes.substr(offset);` creates a copy of the Bytes. and reassigns. Why not simply use `block_number_as_bytes.erase()` to remove the prefix ?
3. The compaction should be performed **only** if body:txn_count > 0 (otherwise is unnecessary)

Giulio2002: okay, i pushed some changes, is that what you mean?

AndreaLanfranchi: Sort of but there are still sanity errors. For instance this is very unsafe for a _generic_ function
```
uint64_t decode_lookup(ByteView encoded_lookup) {
    uint64_t decoded_lookup{0};
    uint8_t  position{0};
    for (auto i = encoded_lookup.rbegin(); i != encoded_lookup.rend(); ++i) {
        decoded_lookup |= uint64_t((*i) << position);
        position += 8;
    }

    return decoded_lookup;
}
```
Should it receive as input a ByteView with length > sizeof(uint64_t) you'd produce martian values as from the 8th cycle you'd begin left switching bits by 72 on a 64 bit value ... which is nonsense. And this should be covered in tests

I'm pushing a commit now to show what I meant

AndreaLanfranchi (approved):

PR #426, opened 2021-09-09
Merge to/from_big_compact with its RLP cousins

yperbasis:
Merge the two sets of methods from rlp & db/util that deal with compact big endian into one. Move the unified set into silkworm::endian.


core/silkworm/common/endian.cpp
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+#include "endian.hpp"
+
+#include <silkworm/common/util.hpp>
+
+namespace silkworm::endian {
+
+Bytes to_big_compact(const uint64_t value) {
+    uint8_t full_be[sizeof(uint64_t)];
```
AndreaLanfranchi, line 24: Can't this be static ?

core/silkworm/common/endian.cpp
```diff
@@ -0,0 +1,69 @@
...
+    store_big_u64(&full_be[0], value);
+    return Bytes{zeroless_view(full_be)};
+}
+
+Bytes to_big_compact(const intx::uint256& value) {
+    uint8_t full_be[sizeof(intx::uint256)];
```
AndreaLanfranchi, line 30: Can't this be static ?

core/silkworm/common/endian.cpp
```diff
@@ -0,0 +1,69 @@
...
+static std::optional<T> from_big_compact(const ByteView& data, bool allow_leading_zeros) {
+    if (data.length() > sizeof(T)) {
+        return std::nullopt;
+    }
+
+    T x{0};
+
+    if (data.empty()) {
+        return x;
+    }
```
AndreaLanfranchi, line 45: Probably worth gather the sanity checks which return std::nullopt in a single place
```suggestion
static std::optional<T> from_big_compact(const ByteView& data, bool allow_leading_zeros) {

    if (data.length() > sizeof(T) ||                               // Exceeds size of T
        (!data.empty() && (!allow_leading_zeros && data[0] == 0))  // Leading zeroes not allowed
    ) {
        return std::nullopt;
    }

    T x{0};
    if (data.empty()) {
        return x;
    }

```

core/silkworm/common/endian.hpp
```diff
@@ -37,6 +33,11 @@ In addition, SILKWORM_BSWAP16, SILKWORM_BSWAP32, and SILKWORM_BSWAP64 macros are
 as compiler intrinsics to swap bytes in 16-bit, 32-bit, and 64-bit integers respectively.
 */
 
+#include <stdint.h>
```
AndreaLanfranchi, line 36: Clion's clang-tidy suggests stdint.h -> cstdint

AndreaLanfranchi (approved):
Besides comments above seems a pretty good code rationalization

core/silkworm/common/endian.cpp
yperbasis: Hmm. I find the separate checks more readable.

core/silkworm/common/endian.cpp
yperbasis: Right, I'll make it thread_local to be on the safe side

core/silkworm/common/endian.cpp
AndreaLanfranchi: Yes I know but the _less readable_ form saves the creation of instance of T which is eventually wasted cause we return std::nullopt

PR #427, opened 2021-09-10
Make ByteView constructible from array. Kill full_view

yperbasis:
Make `ByteView` implicitly constructible from arrays, thus eliminating the need for `full_view`. That simplifies the code a little bit.


core/silkworm/common/util.cpp
```diff
@@ -83,20 +83,14 @@ evmc::bytes32 to_bytes32(ByteView bytes) {
     return out;
 }
 
-ByteView zeroless_view(const ByteView& data) {
+ByteView zeroless_view(ByteView data) {
```
AndreaLanfranchi, line 86: Why remove const reference ?
This way we have a new copy instance of data passed as argument and evenually a third instance returned as result

core/silkworm/common/util.cpp
```diff
@@ -83,20 +83,14 @@ evmc::bytes32 to_bytes32(ByteView bytes) {
...
     unsigned offset{0};
     while (offset < data.length() && data[offset] == 0) {
         ++offset;
     }
     return {data.data() + offset, data.length() - offset};
```
AndreaLanfranchi, line 91: Probably more basic_string_view style to return `data.substr(offset)`

core/silkworm/common/util.cpp
yperbasis: Well, `ByteView` is just a pointer + size, it's cheap to pass it by value. Shouldn't matter much either way.

AndreaLanfranchi: Agree is lightweight but nevertheless there's a copy ... which is eventually havier than a ref.
Semantically then, as we also don't tamper with input data, I'd keep the const ref

mriccobene: I also prefer to pass ByteView by copy, the method signature is more readable. The copy is very cheap: we have a pointer + length copy vs a pointer copy.

AndreaLanfranchi:
Another suggestion : at this point why don't make `zeroless_view` a method of `ByteView` class ? Say ByteView.zeroless()

core/silkworm/common/util.cpp
yperbasis: Right, makes sense

yperbasis:
> Another suggestion : at this point why don't make `zeroless_view` a method of `ByteView` class ? Say ByteView.zeroless() or more consistently `ByteView.zeroless_substr()`

That would work too, but one drawback is that we'll have to explicitly convert arguments to `ByteView`, even if they are implicitly convertible. For example, now we have
```
Bytes to_big_compact(const uint64_t value) {
    SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(uint64_t)];
    store_big_u64(&full_be[0], value);
    return Bytes{zeroless_view(full_be)};
}
```
The last line would have to change to somewhat clumsier
```
return Bytes{ByteView{full_be}.zeroless_substr()};
```

core/silkworm/common/util.cpp
AndreaLanfranchi: Ok. I won't die on this hill

node/silkworm/downloader/packets/NewBlockHashesPacket.hpp
```diff
@@ -21,18 +21,17 @@
...
 
-    using NewBlockHashesPacket = std::vector<NewBlockHash>;
+using NewBlockHashesPacket = std::vector<NewBlockHash>;
 
 namespace rlp {
 
     inline void encode(Bytes& to, const NewBlockHash& from) noexcept {
-        rlp::Header rlp_head{true,
-                             rlp::length(from.hash) + rlp::length(from.number)};
+        rlp::Header rlp_head{true, rlp::length(ByteView{from.hash}) + rlp::length(from.number)};
```
mriccobene, line 34: The formatting changes had prevented me from seeing the semantic change from `rlp::length(Hash)` to 
`rlp::length(ByteView)` (perhaps it would be better for the reviewer to have different changes in different commits).
Anyway why this change? 

We should carefully consider whether to add this overload
`constexpr ByteView(const evmc::bytes32& hash) noexcept;`
I feel it causes overload resolution problems elsewhere

node/silkworm/downloader/packets/NewBlockHashesPacket.hpp
```diff
@@ -41,16 +40,14 @@ namespace rlp {
     }
 
     inline size_t length(const NewBlockHash& from) noexcept {
-        rlp::Header rlp_head{true,
-                             rlp::length(from.hash) + rlp::length(from.number)};
+        rlp::Header rlp_head{true, rlp::length(ByteView{from.hash}) + rlp::length(from.number)};
```
mriccobene, line 43: See previous comment

node/silkworm/downloader/packets/packet_coding_test9.cpp
```diff
@@ -0,0 +1,44 @@
...
+    packet.request.push_back(Hash::from_hex("9f20a871bf5151959fff4c88783bf4ef27b170a4cbe92b8f63ca1fe7d6ab829c"));
+
+    Bytes encoded;
+    rlp::encode(encoded, packet);
+
+    REQUIRE(to_hex(encoded) ==
+            "f84d88ae9405dbeebf3f01f842a0a36b1595c5acd878b63f83d3b62f6882edd27b757582f5319aebc17bc3e98246a09f20a871bf51"
+            "51959fff4c88783bf4ef27b170a4cbe92b8f63ca1fe7d6ab829c");
+
+    auto len = rlp::length(packet);
```
mriccobene, line 39: Here MSVC reports "ambiguous call to overloaded function" 

```
..\silkworm-win\core\silkworm/rlp/encode.hpp(82): error C2668: 'silkworm::rlp::length': ambiguous call to overloaded function
..\silkworm-win\node\silkworm\downloader\packets\HashOrNumber.hpp(43): note: could be 'size_t silkworm::rlp::length(const silkworm::HashOrNumber &)'
..\silkworm-win\core\silkworm/rlp/encode.hpp(67): note: or 'size_t silkworm::rlp::length(silkworm::ByteView)'
```

Maybe it's due to the newly added member in ByteView?

`constexpr ByteView(const evmc::bytes32& hash)`

mriccobene (commented):
I think that killing full_view is ok but I have concerns on the automatic conversion from address and hash to ByteView

mriccobene:
See my comments in the review. Anyway I am going to get rid of errors on the downloader, probably due to slight different overload resolution rules on different compilers, by deleting all the eth66 generic rlp handling code.

mriccobene (commented):
I would prefer to leave these two methods out of ByteView:

```
constexpr ByteView(const evmc::address& address) 

constexpr ByteView(const evmc::bytes32& hash) 
```

and let the author of evmc::address or evmc::bytes32 or Hash or other class to decide if provide a conversion to ByteView in the form of a conversion operator 

`operator ByteView () `

so automatically getting rlp handling or implement custom rlp handling functions.

So we can keep ByteView clean and unaware of other classes (can be many). But other classes can benefit of a conversion to ByteView.

yperbasis:
> I would prefer to leave these two methods out of ByteView:
> 
> ```
> constexpr ByteView(const evmc::address& address) 
> 
> constexpr ByteView(const evmc::bytes32& hash) 
> ```
> 
> and let the author of evmc::address or evmc::bytes32 or Hash or other class to decide if provide a conversion to ByteView in the form of a conversion operator
> 
> `operator ByteView () `
> 
> so automatically getting rlp handling or implement custom rlp handling functions.
> 
> So we can keep ByteView clean and unaware of other classes (can be many). But other classes can benefit of a conversion to ByteView.



> I would prefer to leave these two methods out of ByteView:
> 
> ```
> constexpr ByteView(const evmc::address& address) 
> 
> constexpr ByteView(const evmc::bytes32& hash) 
> ```
> 
> and let the author of evmc::address or evmc::bytes32 or Hash or other class to decide if provide a conversion to ByteView in the form of a conversion operator
> 
> `operator ByteView () `
> 
> so automatically getting rlp handling or implement custom rlp handling functions.
> 
> So we can keep ByteView clean and unaware of other classes (can be many). But other classes can benefit of a conversion to ByteView.

@mriccobene Well, the purpose of `ByteView` is not to be universally useful, but rather to be useful in Silkworm.  `evmc::address` or `evmc::bytes32` are defined in evmc, which doesn't and shouldn't depend on Silkworm, and thus can't use `ByteView`. Silkworm, on the other hand, uses `evmc::address` and `evmc::bytes32` a lot, so IMHO it's a good thing to simplify their use in our codebase.

yperbasis:
Closing in favour of PR #494.

PR #428, opened 2021-09-11
Added Clique Support

Giulio2002:
Things added:
* Consensus Engine System
* POA voting snapshots
* Clique rewards(nothing)
* Configurations for periods, epoch, etc...
* integration of consensus engine framework
* doxygen comments


yperbasis (commented):
Please add tests. The test cases from EIP-225 are a good starting point.

yperbasis:
`context.block_coinbase` in `EvmHost::get_tx_context` should be set accordingly

Giulio2002:
> `context.block_coinbase` in `EvmHost::get_tx_context` should be set accordingly

you mean that it should be different that the address(0) aka. vmHost::get_tx_context == header.beneficiary? shouldn't it be checked in POW as well? or perhaps as a default condition in execution?

yperbasis:
> you mean that it should be different that the address(0) aka. vmHost::get_tx_context == header.beneficiary? shouldn't it be checked in POW as well? or perhaps as a default condition in execution?

Per EIP-225 "Transaction execution must use the actual block signer (see extraData) for the COINBASE opcode and transaction fees must be attributed to the signer account." I see that you've implemented that for transaction fees, but not for the COINBASE opcode (which uses `context.block_coinbase`).

Giulio2002:
> > you mean that it should be different that the address(0) aka. vmHost::get_tx_context == header.beneficiary? shouldn't it be checked in POW as well? or perhaps as a default condition in execution?
> 
> Per EIP-225 "Transaction execution must use the actual block signer (see extraData) for the COINBASE opcode and transaction fees must be attributed to the signer account." I see that you've implemented that for transaction fees, but not for the COINBASE opcode (which uses `context.block_coinbase`).

Oh Man, that is why execution fails miserably... Thank you.

Giulio2002:
@yperbasis This PR only contains Validation of Headers, rewards/evm related stuff will be done later. I think this is already enough to unpack... now it is done, it is ready to be reviewed.

Giulio2002:
Consensus Tests passed BTW, they are implemented in ´clique_test.cpp´

core/silkworm/consensus/validation.hpp
```diff
@@ -64,29 +56,17 @@ enum class [[nodiscard]] ValidationResult{
...
     // See [YP] Section 11.2 "Transaction Validation", Eq (160)
     kWrongBlockGas = 27,  // BHg ≠ l(BR)u
-
     kInvalidSignature = 28,  // EIP-2
-
     kWrongChainId = 29,  // EIP-155
-
-    kUnsupportedTransactionType = 30,  // EIP-2718
+    kInvalidVotingSegment = 30, // Invalid voting segment in Clique/Aura
+    kUnhauthorizedSigner = 31,  // Handling an unhautorized voting signer
```
yperbasis, line 62: Typo: unhautorized

core/silkworm/consensus/validation.hpp
```diff
@@ -64,29 +56,17 @@ enum class [[nodiscard]] ValidationResult{
 
     // See [YP] Section 11.2 "Transaction Validation", Eq (160)
     kWrongBlockGas = 27,  // BHg ≠ l(BR)u
-
     kInvalidSignature = 28,  // EIP-2
-
     kWrongChainId = 29,  // EIP-155
-
-    kUnsupportedTransactionType = 30,  // EIP-2718
+    kInvalidVotingSegment = 30, // Invalid voting segment in Clique/Aura
```
yperbasis, line 61: kInvalidVotingSegment is never returned. Is corresponding validation not implemented yet or is it a mistake?

core/silkworm/consensus/clique/clique_snapshot.hpp
```diff
@@ -0,0 +1,115 @@
...
+        // cast adds a new vote into the tally.
+        void cast(const evmc::address& address, const evmc::address& signer, bool authorize);
+        // uncast removes a previously cast vote from the tally.
+        void uncast(const evmc::address& address, const evmc::address& signer);
+        // uncast_all removes every cast vote from the signer.
+        void uncast_all(const evmc::address& signer);
+        // update update snapshots and cleans it up from db corruption, etc..
+        void update(const uint64_t& block_number, const evmc::bytes32& hash);
+
+        uint64_t block_number_;                             // Block number where the snapshot was created
```
yperbasis, line 102: Please default-initialize `block_number_` to 0 to avoid potentially unitialized value.

core/silkworm/consensus/clique/clique_snapshot.hpp
```diff
@@ -0,0 +1,115 @@
...
+                       std::deque<evmc::address> recents):
+                        block_number_{block_number}, hash_{hash}, signers_{signers}, recents_{recents} {}
+
+        //! \brief Updated snapshot by adding headers
+        //! \param headers: list of headers to add.
+        //! \param config: clique config.
+        ValidationResult add_header(const BlockHeader& header, const evmc::address& signer, const CliqueConfig& config);
+        //! \brief Verify seal for header
+        //! \param header: header to verify.
+        ValidationResult verify_seal(BlockHeader header, const evmc::address& signer);
```
yperbasis, line 63: `header` should be passed by `const &` (`BlockHeader` is not so small, it's expensive to pass it by value). Also, `verify_seal()` should be marked `const` since it's a read-only method.

core/silkworm/consensus/consensus_engine.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    return is_kin(branch_header, *mainline_parent, mainline_header.parent_hash, n - 1, state, old_ommers);
+}
+
+ConsensusEngine& get_consensus_engine(SealEngineType engine_type) {
+    ConsensusEngine* engine;
+    switch (engine_type) {
+        case SealEngineType::kEthash:
+            engine = new Ethash();
```
yperbasis, line 155: This code leaks memory: you allocate memory with new, but never delete it. Return `std::unique_ptr< ConsensusEngine>` instead since [unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr) will automatically delete its memory when no longer in use.

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+std::array<uint8_t, 8> kNonceAuthorize =   {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff};
+std::array<uint8_t, 8> kNonceUnauthorize = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+//! \brief Updated snapshot by adding headers
+//! \param headers: list of headers to add.
+ValidationResult CliqueSnapshot::add_header(const BlockHeader& header, const evmc::address& signer, const CliqueConfig& config) {
+    auto hash{header.hash()};
+    auto tmp_recents{recents_}; // We only modify snapshot after checks are done.
+    // Delete the oldest signer from the recent list to allow it signing again
+    if (recents_.size() > 0 && recents_.size() > signers_.size() / 2) {
```
AndreaLanfranchi, line 34: This double check is pleonastic. 
`recents_.size() > signers_.size() / 2` already implies `rcents_.size() > 0`

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+//! \brief Getter method for hash_.
+//! \return Snapshot's hash.
+const evmc::bytes32& CliqueSnapshot::get_hash() const noexcept {
+    return hash_;
+}
+
+//! \brief Convert the snapshot in JSON.
+//! \return The resulting JSON.
+Bytes CliqueSnapshot::to_bytes() const noexcept {
```
AndreaLanfranchi, line 139: Seems to me the comment is misleading. Says it returns JSON bu the function returns Bytes

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+    for (const auto &recent: recents_) {
+        std::memcpy(&ret[signers_size * kAddressLength + i * kAddressLength + 1], recent.bytes, kAddressLength);
+        i++;
+    }
+    return ret;
+}
+
+//! \brief Decode snapshot from json format.
+//! \return Decoded snapshot.
+CliqueSnapshot CliqueSnapshot::from_bytes(ByteView& b, uint64_t& block_number, const evmc::bytes32& hash) noexcept {
```
AndreaLanfranchi, line 162: Description says `Decode snapshot from JSON format` but function signature has no json input

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+    auto signers_count{b[0]};
+    auto recents_count{(b.size() - signers_count * kAddressLength - 1) / kAddressLength};
```
AndreaLanfranchi, line 164: Possible underflow here ?

core/silkworm/common/util.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include <cassert>
 #include <regex>
+#include <iostream>
```
AndreaLanfranchi, line 21: Probably was left after some debugging ?

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,178 @@
...
+	// Ensure that the extra-data contains a signer list on checkpoint, but none otherwise
+	uint64_t signers_length{header.extra_data.size() - kHashLength - kSignatureLength - 1};
+	if (!checkpoint && signers_length != 0) {
+		return ValidationResult::kUnhauthorizedSigner;
+	}
+	if (checkpoint && (signers_length % kAddressLength) != 0) {
+		return ValidationResult::kMissingSigner;
+	}
+	// Ensure that the mix digest is zero as we don't have fork protection currently
+	if (header.mix_hash != 0x0000000000000000000000000000000000000000000000000000000000000000_bytes32) {
```
AndreaLanfranchi, line 68: evmc::bytes32 can be checked for zeroes using either `is_zero` method or its `bool()` operator.
Same result but semantically more readable

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,178 @@
...
+	if (header.number > 0) {
+		if (header.difficulty == 0 || (header.difficulty != kDiffInTurn && header.difficulty != kDiffNoTurn)) {
+			return ValidationResult::kInvalidGasLimit;
+		}
+	}
+	
+	// All basic checks passed, verify cascading fields
+    if (header.number == 0) {
+        std::vector<evmc::address> signers;
+        size_t signers_count{(header.extra_data.size() - kHashLength - kSignatureLength -1) / kAddressLength};
```
AndreaLanfranchi, line 85: Possible underflow here if `kHashLength + kSignatureLength + 1` exceeds extra_data.size()

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,178 @@
...
+            evmc::address signer;
+            std::memcpy(signer.bytes, 
+                        &header.extra_data[kHashLength + (i * kAddressLength)],
+                        kAddressLength);
+            signers.push_back(signer);
+        }
+        last_snapshot_ = CliqueSnapshot{header.number, header.hash(), signers, {}};
+        state.write_snapshot(header.number, header.hash(), last_snapshot_);
+        return ValidationResult::kOk;
+    } else if (last_snapshot_.get_hash() == 0x0000000000000000000000000000000000000000000000000000000000000000_bytes32) {
```
AndreaLanfranchi, line 96: Maybe use `is_zero` or bool operator

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto header_hash{header.hash()};
+    if (sig_cache_.find(header_hash) != sig_cache_.end()) {
+        return sig_cache_[header_hash];
+    }
+    // Run Ecrecover and get public key
+    auto recovered{ecdsa::recover(full_view(header_hash), signature, v)};
+    if (recovered == std::nullopt) {
+        return std::nullopt;
+    }
+    auto hash{keccak256(recovered->substr(1))};
```
AndreaLanfranchi, line 179: If address recover works as normally I think a check is missing. Can you verify ?
recovered.at[0] should be == 4u
See erigon's reference [here](https://github.com/ledgerwatch/erigon/blob/devel/core/types/transaction_signing.go#L304-L329) and silkworm's recovery [here](https://github.com/torquem-ch/silkworm/blob/master/node/silkworm/stagedsync/recovery/recovery_worker.cpp#L76)

core/silkworm/consensus/clique/clique_snapshot.hpp
```diff
@@ -0,0 +1,115 @@
...
+    private:
+        // is_vote_valid returns whether it makes sense to cast the specified vote in the
+        // given snapshot context (e.g. don't try to add an already authorized signer).
+        bool is_vote_valid(const evmc::address& address, bool authorize) const noexcept;
+        // cast adds a new vote into the tally.
+        void cast(const evmc::address& address, const evmc::address& signer, bool authorize);
+        // uncast removes a previously cast vote from the tally.
+        void uncast(const evmc::address& address, const evmc::address& signer);
+        // uncast_all removes every cast vote from the signer.
+        void uncast_all(const evmc::address& signer);
```
AndreaLanfranchi, line 98: I'd avoid using `cast`, `uncast` and `uncast_all`. In C++ when you cast something you expect a change in the shape of the memory object you point to.
I suggest 
* cast -> increment_votes; 
* uncast -> decrement_votes;
* uncast_all -> clear_votes

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+    } else if (header.nonce == kNonceUnauthorize) {
+        authorize = false;
+    } else {
+        return ValidationResult::kInvalidVote;
+    }
+    
+    tmp_recents.push_front(signer);
+    recents_ = tmp_recents;
+
+    if (header.beneficiary == 0x0000000000000000000000000000000000000000_address) {
```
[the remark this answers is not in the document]
yperbasis: I'd suggest `if (is_zero(header.beneficiary)) {`

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    update(header.number, hash);
+    // Success
+    return ValidationResult::kOk;
+}
+
+//! \brief Verify seal for header
+//! \param header: header to verify.
+ValidationResult CliqueSnapshot::verify_seal(BlockHeader header, const evmc::address& signer) {
+    // Check difficituly
```
AndreaLanfranchi: Typo

Giulio2002:
I am merging master, then all comments are resolved

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,192 @@
...
+    }
+    // Run Ecrecover and get public key
+    auto recovered{ecdsa::recover(full_view(header_hash), signature, v)};
+    if (recovered == std::nullopt) {
+        return std::nullopt;
+    }
+
+    if (recovered->at(0) != 4u) {
+        return std::nullopt;
+    }
```
AndreaLanfranchi, line 182: Group sanity check
```suggestion
    if (!recovered.has_value() || recovered->at(0) !=4u) {
        return std::nullopt;
    }
```

AndreaLanfranchi (changes requested):

AndreaLanfranchi (changes requested):

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,246 @@
...
+namespace silkworm {
+
+std::array<uint8_t, 8> kNonceAuthorize =   {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff};
+std::array<uint8_t, 8> kNonceUnauthorize = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+//! \brief Updated snapshot by adding headers
+//! \param headers: list of headers to add.
+ValidationResult CliqueSnapshot::add_header(const BlockHeader& header, const evmc::address& signer, const CliqueConfig& config) {
+    auto hash{header.hash()};
+    auto tmp_recents{recents_}; // We only modify snapshot after checks are done.
```
AndreaLanfranchi, line 32: There is definitely no need to create an (expensive) copy of recents only to simulate if the last in queue gets popped out the queue.

Simply track if this signer **would be popped out** and act accordingly.

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,246 @@
...
+    if (header.nonce == kNonceAuthorize) {
+        authorize = true;
+    } else if (header.nonce == kNonceUnauthorize) {
+        authorize = false;
+    } else {
+        return ValidationResult::kInvalidVote;
+    }
+    
+    tmp_recents.push_front(signer);
+    recents_ = tmp_recents;
```
AndreaLanfranchi, line 62: Besides the fact this should not be here this is dramatically expensive.
Should you need to move contents of tmp_recents to recents_ (and eventually discard tmp_recents_) use 

`recents_.swap(tmp_recents);`

which only moves pointers to their respective data.

AndreaLanfranchi (changes requested):

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,246 @@
...
+        i++;
+    }
+    // We add the recents
+    i = 0;
+    for (const auto &recent: recents_) {
+        std::memcpy(&ret[signers_size * kAddressLength + i * kAddressLength + 1], recent.bytes, kAddressLength);
+        i++;
+    }
+    return ret;
+}
```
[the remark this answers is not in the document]
AndreaLanfranchi: > I also would advise not to use recents_ as deque as there is apparently no reason for that. Maybe you can elaborate the rationale of the choice

Can you please elaborate on this ?

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,248 @@
...
+//! \param block_number: Block to check.
+//! \param address: Address to check.
+//! \return if a signer at a given block height is in charge or not.
+bool CliqueSnapshot::is_authority(uint64_t block_number, evmc::address address) const noexcept {
+    uint64_t offset{0};
+	while (offset < signers_.size() && signers_[offset] != address) {
+		++offset;
+	}
+	return (block_number % signers_.size()) == offset;
+}
```
AndreaLanfranchi, line 117: This is wrong. You increase complexity to O(n) while complexity is O(1).
This time I won't put the suggestion here as I wish you reason a bit on that

AndreaLanfranchi (changes requested):

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -111,11 +111,8 @@ ValidationResult CliqueSnapshot::verify_seal(const BlockHeader& header, const ev
 //! \param address: Address to check.
 //! \return if a signer at a given block height is in charge or not.
 bool CliqueSnapshot::is_authority(uint64_t block_number, evmc::address address) const noexcept {
-    uint64_t offset{0};
-	while (offset < signers_.size() && signers_[offset] != address) {
-		++offset;
-	}
-	return (block_number % signers_.size()) == offset;
+	return (block_number % signers_.size())
+           == static_cast<uint64_t>(std::find(signers_.begin(), signers_.end(), address) - signers_.begin());
```
AndreaLanfranchi, line 115: This is still complexity O(n).

AndreaLanfranchi (changes requested):

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,188 @@
...
+    if (!recovered.has_value() || recovered->at(0) !=4u) {
+        return std::nullopt;
+    }
+    auto hash{keccak256(recovered->substr(1))};
+    // Convert public key to address
+    evmc::address signer{};
+    std::memcpy(signer.bytes, &hash.bytes[12], kAddressLength);
+    sig_cache_.emplace(header_hash, signer);
+    return signer;
+}
```
AndreaLanfranchi, line 185: This whole block is not covered by tests which means the tests do not cover address recovery from signature. How can we rely on correctness of this code ?

Besides: though I understand you're passing header by copy cause you tamper with extra_data I would suggest to explore the opportunity to override the `header.hash()` method so it produces a hash accordingly to consensus engine type

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "clique_snapshot.hpp"
+#include <silkworm/common/endian.hpp>
+#include <silkworm/common/util.hpp>
+#include <silkworm/crypto/ecdsa.hpp>
+#include <cstring>
+
+namespace silkworm {
+
+std::array<uint8_t, 8> kNonceAuthorize =   {0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff};
+std::array<uint8_t, 8> kNonceUnauthorize = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
```
AndreaLanfranchi, line 26: These two constants are better named in erigon.
- kNonceAuthorize -> kNonceAuthVote
- kNonceUnauthorize -> kNonceDropVote

Are also immediately readable

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,249 @@
...
+    return ValidationResult::kOk;
+}
+
+//! \brief Verify seal for header
+//! \param header: header to verify.
+ValidationResult CliqueSnapshot::verify_seal(const BlockHeader& header, const evmc::address& signer) {
+    // Check difficulty
+    auto authority{is_authority(header.number, signer)};
+    if (authority && header.difficulty != kDiffInTurn && header.difficulty != kDiffNoTurn) {
+        return ValidationResult::kIntrinsicGas;
```
AndreaLanfranchi, line 103: Why this return value ? What does it mean ? As you're validating seal I'd expect, on validation error, to return `ValidationResult::kInvalidSeal`

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    update(header.number, hash);
+    // Success
+    return ValidationResult::kOk;
+}
+
+//! \brief Verify seal for header
+//! \param header: header to verify.
+ValidationResult CliqueSnapshot::verify_seal(const BlockHeader& header, const evmc::address& signer) {
```
AndreaLanfranchi, line 99: This whole function is not covered by tests which means test cases never verify seal.

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    return ValidationResult::kOk;
+}
+
+//! \brief Checks for authority
+//! \param block_number: Block to check.
+//! \param address: Address to check.
+//! \return if a signer at a given block height is in charge or not.
+bool CliqueSnapshot::is_authority(uint64_t block_number, evmc::address address) const noexcept {
```
AndreaLanfranchi, line 113: This whole function is not covered by tests which means test cases never verify seal.

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+//! \brief Getter method for hash_.
+//! \return Snapshot's hash.
+const evmc::bytes32& CliqueSnapshot::get_hash() const noexcept {
+    return hash_;
+}
+
+//! \brief Convert the snapshot in Bytes.
+//! \return The resulting Bytes.
+Bytes CliqueSnapshot::to_bytes() const noexcept {
```
AndreaLanfranchi, line 140: Whole function not covered by tests

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,249 @@
...
+        std::memcpy(&ret[offset], recent.bytes, kAddressLength);
+        offset += kAddressLength;
+    }
+
+    return ret;
+}
+
+//! \brief Decode snapshot from bytes format.
+//! \return Decoded snapshot.
+CliqueSnapshot CliqueSnapshot::from_bytes(ByteView& b, uint64_t& block_number, const evmc::bytes32& hash) noexcept {
```
AndreaLanfranchi, line 162: Whole function not covered by tests

AndreaLanfranchi (changes requested):

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,256 @@
...
+const evmc::bytes32& CliqueSnapshot::get_hash() const noexcept { return hash_; }
+
+//! \brief Convert the snapshot in Bytes.
+//! \return The resulting Bytes.
+Bytes CliqueSnapshot::to_bytes() const noexcept {
+    auto signers_bytes_count{signers_.size() * kAddressLength};
+    auto recents_bytes_count{recents_.size() * kAddressLength};
+    Bytes ret(signers_bytes_count + recents_bytes_count + 8 /* byte for signers count*/, '\0');
+
+    // We specify how many are signers
```
AndreaLanfranchi, line 150: We must check the case when `signers_.empty()` and/or `recents_.empty()` otherwise we get errors (index-out-bounds) in encoding/decoding tests. Specifically test "Recents without signers snapshot encoding/decoding"

core/silkworm/consensus/clique/clique_snapshot.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+    return ret;
+}
+
+//! \brief Decode snapshot from bytes format.
+//! \return Decoded snapshot.
+CliqueSnapshot CliqueSnapshot::from_bytes(ByteView& b, const uint64_t& block_number,
+                                          const evmc::bytes32& hash) noexcept {
+    // Consume signers count
+    auto signers_count{endian::load_big_u64(&b[0])};
```
AndreaLanfranchi, line 171: We must check the case when `signers_.empty()` and/or `recents_.empty()` otherwise we get errors (index-out-bounds) in encoding/decoding tests. Specifically test "Recents without signers snapshot encoding/decoding"

AndreaLanfranchi (changes requested):

core/silkworm/consensus/clique/clique.cpp
```diff
@@ -0,0 +1,182 @@
...
+	// Ensure that the block's difficulty is meaningful (may not be correct at this point)
+	if (header.number > 0) {
+		auto parent{state.read_header(header.number - 1, header.parent_hash)};
+		if (parent == std::nullopt) {
+			return ValidationResult::kUnknownParent;
+		}
+		if (parent->timestamp > header.timestamp - clique_config_.period) {
+			return ValidationResult::kInvalidTimestamp;
+		}
+		if (header.difficulty != kDiffInTurn && header.difficulty != kDiffNoTurn) {
```
yperbasis, line 78: Should be `||` instead of `&&`

yperbasis (approved):

yperbasis:
Looks good to me now. @AndreaLanfranchi what do you think?

AndreaLanfranchi:
@yperbasis not for me yet.
Still checking :
- check_poa.ccp does not work if you set an arbitrary initial block (access to uninitialized optional)
- I don't see where the consensus engine effectively verifies the signer is the most voted

Imho still work to be done

Giulio2002:
> @yperbasis not for me yet.
> Still checking :
> 
> * check_poa.ccp does not work if you set an arbitrary initial block (access to uninitialized optional)
> * I don't see where the consensus engine effectively verifies the signer is the most voted
> 
> Imho still work to be done

What does the signer is the most voted mean? external address receives votes not already present signer

AndreaLanfranchi:
Closing this PR as

1. Implementation completely overlooks the fact Clique snapshots are stored in a separate database (subfolder clique). This would break database compatibility with erigon
2. Implemented logic of snapshot pretend to be similar to Erigon's but instead overlooks tons of logic: for instance in Erigon snapshot are fixed every 10 blocks (i.e. in an epoch we have 300 snapshots) while in Silkworm's implementation snapshots are saved for every block cause their persistance are bound to state changes (every after a block has been executed)
3. There is no notion of in memory snapshots and a time/height window to keep those snapshots hot
4. Snapshots, when stored in db, only record the list of signers and the list of recent voters: tallies of votes are not stored which makes the future load of the snapshot totally unuseful 
5. Implementation totally overlooks all other non-clique related block checks (i.e. gas checks, fork checks, parent checks, timestamps etc) 
6. All tests written only take into account the voting mechanics and nothing else
7) Too much code is an inplace word-by-word translation from GO to C++ 
8) Code excessively convoluted

Besides I think the actual logic (even erigon's one) can be simplified on behalf of these considerations.
1) Clique engine basically keeps a state of votes which is reset every 30K blocks (same as ethash epoch)
2) Being a "state" it can be forwarded (applying new votes for every new block) as well as unrolled in case of reorgs. 
3) If we save a snapshot of clique state on every LAST block of each epoch we have the data available to unroll at any point

Due to all of the above is preferrable a rewrite.

PR #436, opened 2021-09-25
Consensus engine refactor

AndreaLanfranchi:
By this PR we redefine the structure of consensus_engine.

- `core/silkworm/consensus/engine.hpp` defines `IConsensusEngine` : the interface all consensus engines must implement
- class `ConsensusEngineBase` in `core/silkworm/consensus/base/engine.hpp` implements common validation methods for Header and Block and a default value `kOk` for `validate_seal`
- class `ConsensusEngineEthash` in `core/silkworm/consensus/ethash/engine.hpp` inherits from `ConsensusEngineBase` **and** fully implements (by override) the ProofOfWork validation
- class `ConsensusEngineNoProof` in `core/silkworm/consensus/noproof/engine.hpp` inherits from `ConsensusEngineEthash` **and** overrides PoW verification with always success result

This work is preparatory to future tasks aimed to implement new consensus engines as
- move specialized checks from `ConsensusEngineBase` to lower level classes : for example when we insert Clique consensus we'll need to move **actual** checks for difficulty from `ConsensusEngineBase` to `ConsensusEngineEthash` and implement the repurposed checks in `ConsensusEngineClique`  so each derived class will evaluate results from parent class plus the ones implemented in instance.

Questions you may ask : why `ConsensusEngineNoProof` inherits from `ConsensusEngineEthash` and not from `ConsensusEngineBase` ? Cause for the sake of tests (see consensus.cpp) noproof seal engine always inherits all checks from default ethash consensus engine with the only exception of seal verification.


AndreaLanfranchi:
Hmm need to review transaction validation in consensus

AndreaLanfranchi:
Ok. Ready for review guys when you have time.

core/silkworm/consensus/engine.hpp
```diff
@@ -0,0 +1,94 @@
...
+  public:
+    explicit IConsensusEngine(const ChainConfig& chain_config) : chain_config_{chain_config} {};
+    virtual ~IConsensusEngine() = default;
+
+    //! \brief Performs validation of block header & body that can be done prior to sender recovery and execution.
+    //! \brief See [YP] Sections 4.3.2 "Holistic Validity", 4.3.4 "Block Header Validity", and 11.1 "Ommer Validation".
+    //! \param [in] block: block to pre-validate.
+    //! \param [in] state: current state.
+    //! \note Shouldn't be used for genesis block.
+    virtual ValidationResult validate_block(const Block& block, State& state) = 0;
```
yperbasis, line 41: I'd keep `pre_validate_block` because this is not the full validation. There are some other checks performed in `ExecutionProcessor::validate_transaction` & `execute_and_write_block`.

core/silkworm/consensus/engine.hpp
```diff
@@ -0,0 +1,94 @@
...
+    //! \brief See https://eips.ethereum.org/EIPS/eip-1559
+    virtual std::optional<intx::uint256> expected_base_fee_per_gas(const BlockHeader& header,
+                                                                   const BlockHeader& parent) = 0;
+};
+
+//! \brief Performs a validation of a transaction.
+//! \return Any of kIntrinsicGas, kInvalidSignature, kWrongChainId, kUnsupportedTransactionType, or kOk.
+//! \remarks Should sender of transaction not yet recovered a check on signature's validity is performed
+//! \remarks These function is agnostic to whole block validity
+ValidationResult validate_transaction(const Transaction& txn, uint64_t block_number, const ChainConfig& config,
```
yperbasis, line 86: I'd keep `pre_validate_transaction` and the comment "Performs validation of a transaction that can be done prior to sender recovery and block execution" because it's not the full validation. There are also checks in `ExecutionProcessor::validate_transaction`.

yperbasis (changes requested):

core/silkworm/consensus/engine.hpp
AndreaLanfranchi: I understand the point but let me argue that :

- in the context of "consensus_engine" this is the only validation: any "pre" (or "post") suffix implies there is also a "non-pre" (or "non-post") validation while instead there are none in **the context**
- I understand the concept of consensus in the chain involves both the "grammar and syntax" of blocks and transactions (which actually is what consensus_engine checks) **bundled** with the correctness of the state which is delegated to execution phase(s)

This said we're in a situation where naming is tricky. 
Wrt `validate_transaction` actual code in `processor.cpp` only invokes it as assertion (which is skipped in release build) so, again, the "pre" prefix does not make much sense.

I am also assuming that in a "stage-loop" context like ours "consensus_engine" checks come before execution.

So either we rename `consensus_engine` to `pre_execution_consensus_engine` or we restore the "pre" suffix (which I definitely don't like but eventually can live with).

yperbasis: `pre_execution_consensus_engine` doesn't sound right because consensus engine also covers post-execution block rewards.

`pre_validate_...` is perhaps not great, but it highlights that returned `true` for a transaction doesn't necessarily mean that the transaction is certainly valid.

AndreaLanfranchi: Or again we must extend conensus_engine methods to include other checks performed outside its scope. Otherwise consensus_engine (as a name which defines a scope) seems cripple to me.

AndreaLanfranchi: > `pre_execution_consensus_engine` doesn't sound right because consensus engine also covers post-execution block rewards.

Right but that is an "action" not a "check"

> 
> `pre_validate_...` is perhaps not great, but it highlights that returned `true` for a transaction doesn't necessarily mean that the transaction is certainly valid.

Ok ... but be my strong objection put on record (LOL)

core/silkworm/consensus/blockchain.cpp
```diff
@@ -22,15 +22,19 @@
...
 namespace silkworm::consensus {
 
-Blockchain::Blockchain(State& state, ConsensusEngine& engine, const ChainConfig& config, const Block& genesis_block)
-    : state_{state}, engine_{engine}, config_{config} {
-    evmc::bytes32 hash{genesis_block.header.hash()};
-    state_.insert_block(genesis_block, hash);
-    state_.canonize_block(genesis_block.header.number, hash);
+Blockchain::Blockchain(State& state, const ChainConfig& config, const Block& genesis_block)
+    : state_{state}, config_{config}, engine_{consensus::engine_factory(config)} {
+    prime_state_with_genesis(genesis_block);
```
yperbasis, line 27: Perhaps add
```
assert(engine_ != nullptr);
```

core/silkworm/consensus/engine.hpp
```diff
@@ -28,47 +28,66 @@
...
+    virtual ValidationResult validate_block_header(const BlockHeader& header, State& state) = 0;
 
-    //! \brief See [YP] Section 11.3 "Reward Application".
+    //! \brief Validates the seal of the header
+    virtual ValidationResult validate_seal(const BlockHeader& header) {
+        (void)header;
+        return ValidationResult::kOk;
+    }
+
+    //! \brief Finalizes consensus tasks by applying changes in the state of accounts or of the consensus itself
```
yperbasis, line 55: I'd say "Finalizes block execution" instead of "consensus tasks"

core/silkworm/consensus/engine.hpp
```diff
@@ -28,47 +28,66 @@
...
+    //! \remarks For Ethash See [YP] Section 11.3 "Reward Application".
+    virtual void finalize(IntraBlockState& state, const Block& block, const evmc_revision& revision);
 
     //! \brief See [YP] Section 11.3 "Reward Application".
     //! \param [in] header: Current block to get beneficiary from
     virtual evmc::address get_beneficiary(const BlockHeader& header) = 0;
 
-    virtual ~ConsensusEngine() = default;
+    //! \brief Returns parent header (if any) of provided header
+    virtual std::optional<BlockHeader> get_parent_header(const State& state, const BlockHeader& header) = 0;
```
yperbasis, line 67: Should get_parent_header be in the interface? Looks to me like an implementation detail that belongs to `ConsensusEngineBase` but not `IConsensusEngine`.

core/silkworm/consensus/engine.hpp
```diff
@@ -28,47 +28,66 @@
...
+
+  protected:
+    const ChainConfig& chain_config_;
```
yperbasis, line 70: Should `IConsensusEngine` have any protected stuff? IMHO it should only define pure interface and all protected stuff (incl. chain_config_) can be moved to `ConsensusEngineBase`.

core/silkworm/consensus/engine.hpp
```diff
@@ -28,47 +28,66 @@
...
     //! \param [in] state: current state.
-    //! \param [in] config: current chain config.
     //! \note Shouldn't be used for genesis block.
-    virtual ValidationResult validate_block_header(const BlockHeader& header, State& state,
-                                                   const ChainConfig& config) = 0;
+    virtual ValidationResult validate_block_header(const BlockHeader& header, State& state) = 0;
 
-    //! \brief See [YP] Section 11.3 "Reward Application".
+    //! \brief Validates the seal of the header
+    virtual ValidationResult validate_seal(const BlockHeader& header) {
```
yperbasis, line 50: I'd keep this method (and entire `IConsensusEngine`) pure virtual.

core/silkworm/types/transaction.cpp
```diff
@@ -375,21 +375,22 @@ namespace rlp {
 }  // namespace rlp
 
 void Transaction::recover_sender() {
-    from.reset();
-
+    if (from.has_value()) {
```
yperbasis, line 378: IMHO this memoization is a bit dangerous. What if the transaction is modified between invocations of `recover_sender`.

AndreaLanfranchi: Should this be the case we'd be doing something very wrong elsewhere. Removing the check would invoke an expensive extra check of the signature

core/silkworm/consensus/blockchain.cpp
AndreaLanfranchi: Either a fully backed check for nullptr or leave as is. `asserts` completely loose their meaning on release builds (i.e. NDEBUG is defined)

yperbasis: `assert` also serves as documentation. But yeah, at some point we can probably introduce a version of assert that is triggered in release builds as well. Let's leave as is for now.

yperbasis (approved):

core/silkworm/consensus/engine.hpp
AndreaLanfranchi: Resolved by 6526c60

core/silkworm/consensus/engine.hpp
AndreaLanfranchi: Resolved by 6526c60

core/silkworm/consensus/engine.hpp
AndreaLanfranchi: Resolved by 6526c60

PR #441, opened 2021-09-29
Ecdsa and Sender's recovery refactor

AndreaLanfranchi:
- Extend core's ecdsa.cpp 
- Have checks for recovery of address in a single place
- Have each worker in recovery farm to handle its own ecdsa context
- Simplify recovery farm and workers


core/silkworm/crypto/ecdsa.cpp
```diff
@@ -86,4 +94,34 @@ std::optional<Bytes> recover(ByteView message, ByteView signature, bool odd_y_pa
...
+
+std::optional<evmc::address> recover_address(ByteView message, ByteView signature, bool odd_y_parity) {
+    const auto recovered_public_key{recover(message, signature, odd_y_parity)};
+    if (!recovered_public_key.has_value()) {
+        return std::nullopt;
+    }
+    return public_key_to_address(recovered_public_key.value());
+}
+
+std::optional<evmc::address> recover_address(secp256k1_context* context, ByteView message, ByteView signature,
```
yperbasis, line 117: Perhaps make `context` the last argument and default it to `nullptr` in order to avoid duplicated versions of `recover_address`? (and `recover` would call `create_context` if the passed one is `nullptr`)

AndreaLanfranchi: Yeah right.

AndreaLanfranchi: Hmm wait ... need a bit of refactor then ... that would break the static context instantiated in `recover`

yperbasis: > Hmm wait ... need a bit of refactor then ... that would break the static context instantiated in `recover`

Well, we can have something like
```
std::optional<Bytes> recover(ByteView message, ByteView signature, bool odd_y_parity,
                                                  secp256k1_context* context) {
    static secp256k1_context* static_context{create_context()};
    if (context == nullptr) {
         context = static_context;
    }
    ...
}
```

AndreaLanfranchi: Commit 3f60bd8

yperbasis (approved):

PR #442, opened 2021-09-30
Refactor and extend cursor walking functions

AndreaLanfranchi:
- rename `for_each` and `for_count` so they have `cursor` prefix
- introduce walk direction for the above (forward or backward)
- introduce `cursor_erase` to delete records on behalf of `cursor_for_each` and `cursor_for_count` (also made tests)
- remove method `truncate_table_from` in favor of the more generic `cursor_erase`


node/silkworm/db/mdbx.cpp
```diff
@@ -18,6 +18,27 @@
...
 namespace silkworm::db {
 
+namespace detail {
+
+    //! \brief Returns data of current cursor position or moves it to begin or end of table on behalf of provided
+    //! direction.
+    //! \param [in] _c : A reference to an open cursor
+    //! \param [in] _d : Direction cursor should have
+    //! \return ::mdbx::cursor::move_result
+    static inline ::mdbx::cursor::move_result cursor_data_factory(::mdbx::cursor& _c, CursorMoveDirection _d) {
```
yperbasis, line 28: I'm confused about this function. It seems to do two unrelated thing: conditionally move the cursor and return its data. The name is not particularly enlightening either. What's the purpose of this function? Perhaps remove the last bit, i.e. have something like
```
/// \brief Moves unitialized cursor to the beginning or the end of the table on behalf of provided
//! direction.
static inline void initialize_cursor((::mdbx::cursor& _c, CursorMoveDirection _d) {
        if (_c.eof()) {
            return (_d == CursorMoveDirection::Forward) ? _c.to_first(false) : _c.to_last(false);
        }
}
```

AndreaLanfranchi: Maybe looking at cursor_for_each can help
```
size_t cursor_for_each(::mdbx::cursor& cursor, const WalkFunc& walker, const CursorMoveDirection direction) {
    const mdbx::cursor::move_operation move_operation{direction == CursorMoveDirection::Forward
                                                          ? mdbx::cursor::move_operation::next
                                                          : mdbx::cursor::move_operation::previous};

    size_t ret{0};
    auto data{detail::cursor_data_factory(cursor, direction)};
    while (data.done) {
        ++ret;
        if (!walker(cursor, data)) {
            break;  // Walker function has returned false hence stop
        }
        data = cursor.move(move_operation, /*throw_notfound=*/false);
    }
    return ret;
}
```

Particularly this bit
```
    auto data{detail::cursor_data_factory(cursor, direction)};
    while (data.done) {
```

cursor_data_factory basically returns the first meaningful `data` (if any) on behalf of move direction.
Your proposed solution has two issues imo
- making it void forces higher level code to always issue a `cursor.current()`
- making it void this `return (_d == CursorMoveDirection::Forward) ? _c.to_first(false) : _c.to_last(false);` has no meaning besides wasting a move result

AndreaLanfranchi: > two unrelated thing: conditionally move the cursor and return its data.

They're not unrelated. Methods to_first and to_last are not voids. They do return data

yperbasis: I strongly prefer clarity over tiny (perhaps immaterial) gains in performance. If we absolutely must to have a single function, let's name it something more explicit than `cursor_data_factory`. Also, what's wrong with `cursor_for_each` calling `cursor.current()`? `cursor_for_each` doesn't seem too high-level for that :)

AndreaLanfranchi: Honestly I don't follow. Say apply your suggestion.
for each would become

```
size_t cursor_for_each(::mdbx::cursor& cursor, const WalkFunc& walker, const CursorMoveDirection direction) {
    const mdbx::cursor::move_operation move_operation{direction == CursorMoveDirection::Forward
                                                          ? mdbx::cursor::move_operation::next
                                                          : mdbx::cursor::move_operation::previous};

    size_t ret{0};
    detail::initialize_cursor(cursor, direction);
    auto data{cursor.current(/*throw_not_found=*/ false);
    while (data.done) {
        ++ret;
        if (!walker(cursor, data)) {
            break;  // Walker function has returned false hence stop
        }
        data = cursor.move(move_operation, /*throw_notfound=*/false);
    }
    return ret;
}
```

I honestly don't get the difference and for sure we call cursor-move twice.

I really don't understand what you don't find clear

yperbasis: I'd like to get an idea (at least a rough one) of what a function does just by looking at its name, without reading its comments or source code. That's not always possible, of course, but at least IMHO we should strive towards very readable function names. "cursor_data_factory" doesn't give me a clue about it does.

yperbasis: Also using `cursor.eof()` is a bit unfortunate here because we really want to only position cursors that haven't been positioned. If the cursor was manually positioned beyond the last entry prior to calling `cursor_for_each(..., CursorMoveDirection::Forward)`, it's probably quite unexpected that cursor_for_each moves the cursor back to beginning and starts iterating anew.

yperbasis: I correct my suggestion: the function name should be `position_cursor` instead of `initialize_cursor` to avoid confusion with MDBX `C_INITIALIZED`.

AndreaLanfranchi: > cursor.eof() is a bit unfortunate here because we really want to only position cursors that haven't been positioned

Unfortunately cursor_eof is the only method available. Whether it is beyond highest key or not positioned at all makes no difference for MDBX. A freshly opened cursor (without any previous move operation) always has EOF == True
Comment from mdbx.h about mdbx_cursor_eof
```
/** \brief Determines whether the cursor is pointed to a key-value pair or not,
 * i.e. was not positioned or points to the end of data.
 * \ingroup c_cursors
 *
 * \param [in] cursor    A cursor handle returned by \ref mdbx_cursor_open().
 *
 * \returns A \ref MDBX_RESULT_TRUE or \ref MDBX_RESULT_FALSE value,
 *          otherwise the error code:
 * \retval MDBX_RESULT_TRUE    No more data available or cursor not
 *                             positioned
 * \retval MDBX_RESULT_FALSE   A data is available
 * \retval Otherwise the error code */
 ```

Comment (AndreaLanfranchi) [reply node/silkworm/db/mdbx.cpp:28]: > I'd like to get an idea (at least a rough one) of what a function does just by looking at its name, without reading its comments or source code. That's not always possible, of course, but at least IMHO we should strive towards very readable function names. "cursor_data_factory" doesn't give me a clue about what it does.

Ok. Please update on behalf of your needs cause I'm not sure I can capture the spirit

Comment (AndreaLanfranchi) [reply node/silkworm/db/mdbx.cpp:28]: Maybe you'll also want to take a look at things like [this](https://github.com/torquem-ch/silkworm/blob/master/node/silkworm/stagedsync/util.hpp#L54> which you have previoulsy approved.
I bet no one can get a clue of what is converted to what without reading the context.
Just for the sake of precision

Comment (yperbasis) [reply node/silkworm/db/mdbx.cpp:28]: `convert_to_db_format` is a bad name, agreed.

File: node/silkworm/db/mdbx.hpp [hunk 2]
```diff
@@ -54,33 +56,116 @@ struct EnvConfig {
...
+//! \brief Erases map records by cursor until any record is found
+//! \param [in] cursor : A reference to a cursor opened on a map
+//! \param [in] direction : Whether the cursor should navigate records forward (default) or backwards
+//! \return The overall number of erased records
+//! \remarks If the provided cursor is *not* positioned on any record it will be moved to either the beginning or the
+//! end of the table on behalf of the move criteria.
+//! \warning Might nuke all your table records if not used properly
+size_t cursor_erase(::mdbx::cursor& cursor, const CursorMoveDirection direction = CursorMoveDirection::Forward);
+
+//! \brief Erases map records by cursor until any record is found
```
yperbasis, line 140: This sentence sounds unclear to me. To my mind the old description was clearer:
```
 * Truncate a table by starting from a certain key, either up or down.
 * If reverse is set to false, then every key either higher or equal than starting key will be erased.
 * If reverse is set to true, then every key either lower(not equal) than starting key will be erased.
 ```
Now it's not specified whether the entry with `set_key` is erased or not.

File: node/silkworm/db/mdbx.cpp [hunk 3]
```diff
@@ -20,22 +20,25 @@ namespace silkworm::db {
...
-    //! \param [in] _d : Direction cursor should have
-    //! \return ::mdbx::cursor::move_result
-    static inline ::mdbx::cursor::move_result cursor_data_factory(::mdbx::cursor& _c, CursorMoveDirection _d) {
-        if (_c.eof()) {
-            return (_d == CursorMoveDirection::Forward) ? _c.to_first(false) : _c.to_last(false);
+    //! \brief Returns data of current cursor position or moves it to the beginning or the end of the table based on
+    //! provided direction if the cursor is not positioned.
+    //! \param [in] c : A reference to an open cursor
+    //! \param [in] d : Direction cursor should have \return ::mdbx::cursor::move_result
+    static inline ::mdbx::cursor::move_result data_of_positioned_cursor(::mdbx::cursor& c, CursorMoveDirection d) {
```
AndreaLanfranchi, line 27: This is semantically wrong.
`data_of_positioned_cursor == cursor.current`

Here instead the function

1. cursor is moved
2. data is pulled

node/silkworm/db/mdbx.hpp
yperbasis: Ah, OK, my bad, didn't notice that. However, "Erases map records by cursor until any record is found" sounds a bit weird to me, but it's a small thing.

node/silkworm/db/mdbx.cpp
yperbasis: Fair enough, `adjust_cursor_postion_if_unpositioned_and_return_data` is more accurate

AndreaLanfranchi: Ok. Maybe 8f88f5b ?
We can start a new coding pattern ?

yperbasis: Well, function name should reflect its intention. "erase" clearly means that it erases something. If I want more detail, then I can read the doc or the source code. But I get some rough idea from "erase" alone. The problem with "cursor_data_factory" is that doesn't give a hint (at least not to me) about the function's intention to move the cursor to the beginning/end of the table if the cursor is not positioned.

AndreaLanfranchi: Well. I'll stop fooling around here cause we're building a case over the name of a function which :

1. Is extensively commented
2. Is in detail namespace of mdbx.cpp file (so not exposed for other source files)
3. Used only in the same file where paging up few lines is not a big effort
4. has "factory" in its name which clearly states a **creational pattern**

I wish we could not waste more hours on things like this cause, honestly, our code is disseminated of things like this (eg. hash() is the verb implying an action - to hash something - or the getter of a property ?). 
Anyway ... your will be done and let's move on.

yperbasis (approved):

PR #445, opened 2021-10-01
convert_to_db_format -> change_set_to_plain_state_format

yperbasis:
`convert_to_db_format` was a bad name because it converts from one DB format (AccountChangeSet or StorageChangeSet) into another DB format (PlainState).

### DIFF
--- a/node/silkworm/stagedsync/stage_execution.cpp
+++ b/node/silkworm/stagedsync/stage_execution.cpp
@@ -32,9 +32,9 @@
 namespace silkworm::stagedsync {
 
 // block_num is input-output
-static StageResult execute_batch_of_blocks(mdbx::txn& txn, const ChainConfig& config, const uint64_t max_block,
+static StageResult execute_batch_of_blocks(mdbx::txn& txn, const ChainConfig& config, const BlockNum max_block,
                                            const db::StorageMode& storage_mode, const size_t batch_size,
-                                           uint64_t& block_num, uint64_t prune_from) noexcept {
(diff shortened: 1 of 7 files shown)


node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -197,24 +198,24 @@ static void revert_state(Bytes key, Bytes value, mdbx::cursor& plain_state_table
...
 // For given changeset cursor/bucket it reverts the changes on states buckets
 static void unwind_state_from_changeset(mdbx::cursor& source, mdbx::cursor& plain_state_table,
                                         mdbx::cursor& plain_code_table, uint64_t unwind_to) {
-    uint64_t block_number{0};
     auto src_data{source.to_last(/*throw_notfound*/ false)};
     while (src_data) {
         Bytes key(db::from_slice(src_data.key));
         Bytes value(db::from_slice(src_data.value));
-        block_number = endian::load_big_u64(&key[0]);
+        const uint64_t block_number = endian::load_big_u64(&key[0]);
```
AndreaLanfranchi, line 214: For block numbers we should consistently use the proper alias

```suggestion
        const BlockNum block_number = endian::load_big_u64(&key[0]);
```

AndreaLanfranchi (changes requested):

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: Right, done.

AndreaLanfranchi (approved):

PR #449, opened 2021-10-02
Enforce FutureBlock validation

### LINKED ISSUE #448: Future timestamps on block header verification
Erigon [does check for future timestamps in block header](https://github.com/ledgerwatch/erigon/blob/8d9e058eb727b853845625db759006125a326b3f/consensus/ethash/consensus.go#L266-L272) **but the check is skipped for uncle's headers**. 
Reading tests
`tests\BlockchainTests\InvalidBlocks\bcUncleHeaderValidity\incorrectUncleTimestamp4.json`
`tests\BlockchainTests\InvalidBlocks\bcUncleHeaderValidity\incorrectUncleTimestamp5.json`
this comment specifies that

> Uncle has bizzare future timestamp but geth says its ok

Our Silkworm implementation does not implement the test yet and also the [verify_block_header](https://github.com/torquem-ch/silkworm/blob/master/core/silkworm/consensus/engine.hpp#L47) does not take into account if the header being verified is for an uncle or not.

Need to implement the same

AndreaLanfranchi:
Fixes #448

### DIFF
--- a/core/silkworm/consensus/base/engine.cpp
+++ b/core/silkworm/consensus/base/engine.cpp
@@ -26,7 +26,7 @@ namespace silkworm::consensus {
 ValidationResult ConsensusEngineBase::pre_validate_block(const silkworm::Block& block, silkworm::State& state) {
     const BlockHeader& header{block.header};
 
-    if (ValidationResult err{validate_block_header(header, state)}; err != ValidationResult::kOk) {
+    if (ValidationResult err{validate_block_header(header, state, /*is_ommer=*/false)}; err != ValidationResult::kOk) {
         return err;
     }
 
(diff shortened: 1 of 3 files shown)


core/silkworm/consensus/base/engine.cpp
```diff
@@ -81,7 +82,14 @@ ValidationResult ConsensusEngineBase::pre_validate_block(const silkworm::Block&
     return ValidationResult::kOk;
 }
 
-ValidationResult ConsensusEngineBase::validate_block_header(const BlockHeader& header, State& state) {
+ValidationResult ConsensusEngineBase::validate_block_header(const BlockHeader& header, State& state, bool is_ommer) {
```
yperbasis, line 85: I'd use something like "check_physical_time" instead of "is_ommer" for accuracy. To my mind the fact that ommers are allowed to have timestamps in the future is a quirk of geth rather than a logical thing to do. In other words, "is_ommer" is a poor moniker for the actual check happening here.

yperbasis: Also, "check_physical_time" highlights the fact that the function is no longer pure and depends not only on its arguments, but on the environment (machine's time).

AndreaLanfranchi: I mostly agree but imo `check_physical_time` does not depict correct meaning. 
I'm more inclined to use `with_future_timestamp_check` (which also adheres to compiler's syntax for warnings) and a good remark line which states the check is performed against current time of host

yperbasis: Right, "with_future_timestamp_check" sounds good.

AndreaLanfranchi: Done

yperbasis (approved):

PR #452, opened 2021-10-04
avoid extra string copy

greg7mdp:
add_leaf() is called with a Bytes parameter created in unpack_nibbles(). Why switch to a ByteView, which will cause another Bytes creation when assigning to key_ in add_leaf()'s body?


yperbasis:
LGTM. Can you please make the same change to `add_branch_node` for symmetry?

greg7mdp:
Sure, I will check in the requested change.

I'm a little confused by the function:

```
static Bytes node_ref(ByteView rlp) {
    if (rlp.length() < kHashLength) {
        return Bytes{rlp};
    }
    const ethash::hash256 hash{keccak256(rlp)};
    return wrap_hash(hash.bytes);
}
```

and its early return... it looks like it can return two different things as node_ref?
If the early return is the most common case it would make sense for it to take a Bytes as well.

yperbasis (approved):

PR #456, opened 2021-10-05
Storage changes

AndreaLanfranchi:
- move storage mode in its own file
- introduce prune mode
- decorate storage mode as deprecated


node/silkworm/db/storage.hpp
```diff
@@ -0,0 +1,184 @@
...
+#define SILKWORM_STORAGE_HPP_
+
+#include "mdbx.hpp"
+#include "tables.hpp"
+
+namespace silkworm::db {
+
+//! \brief Holds the storage mode set
+struct StorageMode {
+    bool Initialized;  // Whether db storage has been initialized
```
yperbasis, line 28: 1) Please use snake_case (e.g. `call_traces`) for struct data members for consistency with [Goggle's Style Guide](https://google.github.io/styleguide/cppguide.html#Variable_Names) and the rest of our code.
2) I strongly encourage to use default initializers for things that otherwise can be uninitialized. For example, if we have
```
struct Foo {
 bool a;
 bool b;
};
```
and later somebody writes `Foo x;` then happy debugging of inconsistent behaviour on different platforms 😖
```
struct Foo {
 bool a{false};
 bool b{false};
};
```
instead prevents such mistakes.
3. Given the above, are `Initialized` & `kDefaultStorageMode` necessary? What if we have something like
```
struct StorageMode {
    bool history{true};
    bool receipts{true};
    bool tx_index{true};
    bool call_traces{true};
    bool tevm{false};
}
```
4. The same goes for `PruneMode`.

yperbasis (changes requested):

node/silkworm/db/storage.hpp
AndreaLanfranchi: I would not focus too much on StorageMode as it has to be dropped (Erigon has switched to PruneMode)
Will look at PruneMode

PR #461, opened 2021-10-08
TEST_CASE("Incremental vs regeneration")

yperbasis:
Add a test case that compares incremental trie generation with regeneration in one big step.

Also:
* More fixes to AccountTrieCursor traversal
* Fix the case when there's a changed account between a can_skip_state node and the next node

Part of Issue #179.


greg7mdp:
BYW, I'm super happy to see the great progress being made in Silkworm!

yperbasis:
> Thanks for explaining, however I think that even in the use case you describe, it is likely more efficient (both time + space) to use a btree_set that maintains a sorted collection and never inserts duplicates.

Perhaps. For now I'm concentrating on the correctness of the algorithm. I still have to implement incremental intermediate hashes for storage. When the algo is fully implemented, I'm going to profile and optimize its performance. Will revisit PrefixSet then.

PR #465, opened 2021-10-12
Semantic and performance fussiness

AndreaLanfranchi:
Feeling niggling today.
Refactored `to_hex` and `from_hex` functions we heavily rely on.

to_hex : direct allocation of out string and use pointer to data.
Measurement on 100k iterations batches
- previous version average speed 115ms
- actual version average speed 86ms

from_hex: use lookup tables for decoding hex chars instead of math; direct allocation of byte array and use pointer to data
Measurement on 100k iterations batches
- previous version average speed 198.119ms
- actual version average speed 104.040ms

zeroless_view: use suggestion from [here](https://github.com/torquem-ch/silkworm/pull/427#discussion_r724242803)

### DIFF
--- a/core/silkworm/common/util_test.cpp
+++ b/core/silkworm/common/util_test.cpp
@@ -46,6 +46,8 @@ TEST_CASE("Split") {
 
 TEST_CASE("Hex") {
 
+    CHECK(decode_hex_digit('g').has_value() == false);
+
     auto parsed_bytes = from_hex("");
     CHECK((parsed_bytes.has_value() == true && parsed_bytes->empty()));
 


greg7mdp:
> No cause hex digit `0` (whose decimal value is 48) must return decimal 0
Oh man you are right thanks! Great change BTW!

AndreaLanfranchi:
> small change, but why not
> 
> ```
>         auto a = unhex_lut4(static_cast<uint8_t>(*src++));
>         auto b = unhex_lut(static_cast<uint8_t>(*src++));
> ```

Thank you. Done in e0175eb

PR #483, opened 2021-11-04
Downloader unwind method

mriccobene:
Complete the implementation of the downloader unwind method.
Test are still missing.

### DIFF
--- a/node/silkworm/downloader/internals/db_tx.hpp
+++ b/node/silkworm/downloader/internals/db_tx.hpp
@@ -313,17 +313,10 @@ class Db::ReadWriteAccess::Tx : public Db::ReadOnlyAccess::Tx {
     }
 
     void delete_canonical_hash(BlockNum b) {
+        auto hashes_table = db::open_cursor(txn, db::table::kCanonicalHashes);
         Bytes key = db::block_key(b);
         auto skey = db::to_slice(key);
-
-        auto hashes_table = db::open_cursor(txn, db::table::kCanonicalHashes);
-
-        bool throw_notfound = false;
-        auto result = hashes_table.find(skey, throw_notfound);
-        if (!result) return;
-
-        hashes_table.erase(false);  // false = no all duplicates
-        hashes_table.close();
+        (void)hashes_table.erase(skey);
     }
 };
 
--- a/node/silkworm/downloader/internals/db_tx.hpp
+++ b/node/silkworm/downloader/internals/db_tx.hpp
@@ -19,6 +19,7 @@
 
 #include <functional>
 
+#include <silkworm/common/assert.hpp>
 #include <silkworm/common/cast.hpp>
 #include <silkworm/common/endian.hpp>
 #include <silkworm/db/access_layer.hpp>
@@ -217,9 +218,7 @@ class Db::ReadOnlyAccess::Tx {
             ByteView key = db::from_slice(result.key);
             ByteView value = db::from_slice(result.value);
 
-            if (key.size() != sizeof(uint64_t) + kHashLength) {
-                throw std::logic_error("key in td table has to be 40 bytes long: " + result.key.as_hex_string());
-            }
+            SILKWORM_ASSERT(key.size() == sizeof(BlockNum) + kHashLength);
 
             Hash hash{key.substr(sizeof(BlockNum))};
             ByteView block_num = key.substr(0, sizeof(BlockNum));
--- a/node/silkworm/downloader/internals/db_tx.hpp
+++ b/node/silkworm/downloader/internals/db_tx.hpp
@@ -26,6 +26,7 @@
 #include <silkworm/db/tables.hpp>
 #include <silkworm/db/util.hpp>
 
+#include "cpp20_backport.hpp"
 #include "types.hpp"
 
 using namespace silkworm;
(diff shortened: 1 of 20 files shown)


node/silkworm/downloader/internals/db_tx.hpp
```diff
@@ -202,6 +203,49 @@ class Db::ReadOnlyAccess::Tx {
...
+                                                                          mdbx::cursor::move_result& result) -> bool {
+            ByteView key = db::from_slice(result.key);
+            ByteView value = db::from_slice(result.value);
+
+            if (key.size() != 40) {
+                throw std::logic_error("key in td table has to be 40 bytes long: " + result.key.as_hex_string());
+            }
+
+            Hash hash{key.substr(8)};
+            ByteView block_num = key.substr(0, 8);
```
AndreaLanfranchi, line 225: Semantically I'd prefer to consume key view early. I think is neater and reflects the composition of the key depicted above. 
```suggestion
            auto block_num{endian::load_big_u64(key.data())};
            key.remove_prefix(sizeof(block_num));
            Hash hash{key};
```

Eventually change below `max_block_num = block_num`

AndreaLanfranchi (changes requested):

node/silkworm/downloader/internals/db_tx_test.cpp
```diff
@@ -87,6 +87,52 @@ namespace db {
...
             REQUIRE(read_stage_progress == stage_progress);
         }
+
+        SECTION("write/read/delete canonical hash") {
+            CHECK_NOTHROW(tx.write_canonical_hash(header.number, header.hash()));
+            CHECK_NOTHROW(tx.delete_canonical_hash(header.number));
+
+            auto read_hash = tx.read_canonical_hash(header.number);
+
+            REQUIRE(read_hash == std::nullopt);
```
AndreaLanfranchi, line 97: In such tests where each step conditions the execution of the subsequents always prefer `REQUIRE` vs `CHECK`
The first stops the test if the condition is (not) met while the latter makes the test to continue and this may cause the following tests to produce misleading messages.
```suggestion
            REQUIRE_NOTHROW(tx.write_canonical_hash(header.number, header.hash()));
            REQUIRE_NOTHROW(tx.delete_canonical_hash(header.number));

            auto read_hash = tx.read_canonical_hash(header.number);

            REQUIRE(read_hash == std::nullopt);
```

AndreaLanfranchi (changes requested):

node/silkworm/downloader/internals/db_tx.hpp
```diff
@@ -202,6 +203,49 @@ class Db::ReadOnlyAccess::Tx {
...
+            }
+
+            return true;  // = continue loop
+        };
+
+        db::cursor_for_each(td_cursor, find_max, db::CursorMoveDirection::Reverse);
+
+        if (max_block_num == 0) {
+            max_hash = *read_canonical_hash(0);
+        }
```
AndreaLanfranchi, line 245: Accorrding to the logic of the `find_max` lambda above this snippet is unneded (the lambda traverses **all** the records of `kDifficulty` table down to block 0 included)

AndreaLanfranchi (changes requested):

node/silkworm/downloader/internals/db_tx.hpp
mriccobene: I think the current code is more readable, it clearly shows that the key is divided in two parts. Maybe I need to substitute 8 with sizeof(block_num)

node/silkworm/downloader/internals/db_tx.hpp
AndreaLanfranchi: Sorry ... I misread the snippet and realize now you're retrieving the **canonical** hash (not _a hash_ from difficulty).
Nevertheless it seems to me wrong cause reading [Erigon's implementation](https://github.com/ledgerwatch/erigon/blob/devel/eth/stagedsync/stage_headers.go#L329-L334) the code is
```
		if maxNum == 0 {
			maxNum = u.UnwindPoint
			if maxHash, err = rawdb.ReadCanonicalHash(tx, maxNum); err != nil {
				return err
			}
		}
```
which does not read the canonical hash for block 0, rather the canonical hash for the most recent unwind point.

node/silkworm/downloader/internals/db_tx.hpp
yperbasis: I agree with Mike. `sizeof(block_num)` instead of 8 probably helps readability, while whether `remove_prefix` does so is moot.

node/silkworm/downloader/internals/db_tx.hpp
mriccobene: The assigment `maxNum = u.UnwindPoint ` is done on the calling function `PersistedChain::remove_headers()` that checks `if max_block_num == 0` but my code end up to use the wrong hash in this condition. So I will remove this snippet here and I will fix the remove_headers() method. Thanks!

mriccobene: Implemented in commit 6d6ff8d

node/silkworm/downloader/internals/db_tx.hpp
```diff
@@ -202,6 +203,45 @@ class Db::ReadOnlyAccess::Tx {
...
+        BigInt max_td = 0;
+
+        auto td_cursor = db::open_cursor(txn, db::table::kDifficulty);
+
+        auto find_max = [bad_headers, &max_block_num, &max_hash, &max_td](mdbx::cursor&,
+                                                                          mdbx::cursor::move_result& result) -> bool {
+            ByteView key = db::from_slice(result.key);
+            ByteView value = db::from_slice(result.value);
+
+            if (key.size() != sizeof(uint64_t) + kHashLength) {
```
yperbasis, line 220: PR #488 introduces `SILKWORM_ASSERT`, which aborts the program on assert failure even in release mode. Here aborting the program in the case of the wrong key size is probably the only thing we can do, since it's unlikely that we can recover from corrupted database. So I'd replace exception throwing with
```
SILKWORM_ASSERT(key.size() == sizeof(BlockNum) + kHashLength);
```

yperbasis (approved):
LGTM. @AndreaLanfranchi have your concerns been addressed?

AndreaLanfranchi:
> LGTM. @AndreaLanfranchi have your concerns been addressed?

No other concerns atm ... would suggest @mriccobene a pass on changed files to clear Clang-Tidy suggestions by Clion (I know he uses it).

One small issue only. In `sentry_client.hpp`

```
class SentryClient : public rpc::Client<sentry::Sentry> {
  public:
    using base_t = rpc::Client<sentry::Sentry>;

    // ...
  
    void exec_remotely(SentryRpc& rpc);  // send a rpc request to the remote sentry

    // ...
};
```

`exec_remotely` is a non virtual function of `rpc::Client<sentry::Sentry>` 
As the implementation is 
```
void SentryClient::exec_remotely(SentryRpc& rpc) { base_t::exec_remotely(rpc); }
```

why hide it in the inherited class ?

node/silkworm/downloader/internals/db_tx.hpp
mriccobene: Sure! I added a commit with the proposed change. Also I will substitute logic_error with SILKWORM_ASSERT in other places, where it makes sense, in future PR

mriccobene:
@AndreaLanfranchi I will check Clang-Tidy suggestions in a future PR even if I and Clang-Tidy don't get along.

Also SentryClient will be completely rewritten in a following PR where I will introduce gRPC reconnect logic.

Please note that your requested changes has been implemented in past commits.

AndreaLanfranchi (approved):

PR #484, opened 2021-11-05
Signals

AndreaLanfranchi:
Introduce `SignalHandler` to enable (everywhere) check for process being signalled and allow neat shutdown

### DIFF
--- a/cmd/check_senders.cpp
+++ b/cmd/check_senders.cpp
@@ -14,10 +14,7 @@
    limitations under the License.
 */
 
-#include <atomic>
 #include <csignal>
-#include <filesystem>
-#include <queue>
 #include <string>
 #include <thread>
 
@@ -31,23 +28,13 @@
 #include <silkworm/concurrency/worker.hpp>
 #include <silkworm/crypto/ecdsa.hpp>
 #include <silkworm/db/access_layer.hpp>
-#include <silkworm/db/mdbx.hpp>
 #include <silkworm/db/stages.hpp>
-#include <silkworm/db/util.hpp>
 #include <silkworm/etl/collector.hpp>
 #include <silkworm/stagedsync/recovery/recovery_farm.hpp>
-#include <silkworm/types/block.hpp>
 
 using namespace silkworm;
 
 std::unique_ptr<stagedsync::recovery::RecoveryFarm> farm;
-
-void sig_handler(int) {
-    std::cout << std::endl << " Got interrupt. Stopping ..." << std::endl << std::endl;
-    if (farm) {
-        farm->stop();
-    }
-}
 
 struct app_options_t {
     std::string datadir{};  // Provided database path
@@ -61,6 +48,9 @@ struct app_options_t {
 };
 
 int main(int argc, char* argv[]) {
+
+    SignalHandler::init();
+
     namespace fs = std::filesystem;
     // Init command line parser
     CLI::App app("Senders recovery tool.");
@@ -115,9 +105,6 @@ int main(int argc, char* argv[]) {
         auto txn{env.start_write()};
 
         stagedsync::StageResult result{stagedsync::StageResult::kSuccess};
-
-        signal(SIGINT, sig_handler);
-        signal(SIGTERM, sig_handler);
 
         if (app_recover) {
             // Create farm instance and do work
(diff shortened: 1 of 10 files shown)


node/silkworm/common/signal_handler.cpp
```diff
@@ -0,0 +1,130 @@
...
+*/
+
+#include "signal_handler.hpp"
+
+#include <csignal>
+#include <iostream>
+
+namespace silkworm {
+
+uint32_t SignalHandler::sig_count_{0};
```
yperbasis, line 24: I'd use default initializers in the header, but it's a matter of preference.

node/silkworm/common/signal_handler.hpp
```diff
@@ -0,0 +1,39 @@
...
+//! \brief Handler with static storage for signals sig
+class SignalHandler {
+  public:
+    static void init();                                           // Enable the hooks
+    static void handle(int sig_code);                             // Handles incoming signal
+    [[nodiscard]] static bool signalled() { return signalled_; }  // Whether a signal has been intercepted
+    static void reset();                                          // Reset to un-signalled (see tests coverage)
+
+  private:
+    static uint32_t sig_count_;
```
yperbasis, line 34: Shouldn't these guys be atomic? Signal handling should be thread-safe (https://en.cppreference.com/w/cpp/utility/program/signal).

cmd/check_tx_lookup.cpp
```diff
@@ -131,15 +117,15 @@ int main(int argc, char* argv[]) {
                 SILKWORM_LOG(LogLevel::Info) << "Scanned blocks " << expected_block_number << std::endl;
             }
 
-            if (should_stop_) {
+            if (SignalHandler::signalled()) {
```
greg7mdp, line 120: should be `signaled` with one `l`

yperbasis: signalled is also correct: https://www.merriam-webster.com/dictionary/signal

cmd/check_senders.cpp
```diff
@@ -31,24 +28,14 @@
...
-#include <silkworm/types/block.hpp>
 
 using namespace silkworm;
 
 std::unique_ptr<stagedsync::recovery::RecoveryFarm> farm;
 
-void sig_handler(int) {
-    std::cout << std::endl << " Got interrupt. Stopping ..." << std::endl << std::endl;
-    if (farm) {
-        farm->stop();
```
greg7mdp, line 48: Is the call to `farm->stop();` not needed anymore?

greg7mdp: Nice change overall!

cmd/check_tx_lookup.cpp
greg7mdp: you are right, my bad. It is less common (about 10 times fewer results on google), but yes!

node/silkworm/common/signal_handler.cpp
AndreaLanfranchi: Right. (I had issues cause I forgot to inline the initializer)

node/silkworm/common/signal_handler.hpp
AndreaLanfranchi: Right.

cmd/check_senders.cpp
AndreaLanfranchi: Hmm. It was already removed in commit 2b0cd28
Maybe you were reviewing the initial commit ?

node/silkworm/common/signal_handler.cpp
AndreaLanfranchi: Solved vy c265f4f

cmd/check_senders.cpp
greg7mdp: yes, it was removed. But before when handling a signal we called `farm->stop();`, and we don't call it anymore in the new version. Just wondering if it is not needed anymore?

AndreaLanfranchi: Oh I see your point. You refer to method `stop()` for the farm class.
Well this is still needed : should any worker return an error in the recovery phase then the whole farm is stopped. This is a programmatic stop, not a signalled one.
See https://github.com/torquem-ch/silkworm/blob/master/node/silkworm/stagedsync/recovery/recovery_farm.cpp#L231-L303

yperbasis (approved):

PR #486, opened 2021-11-06
Signals - Don't inline initialization of static memeber

AndreaLanfranchi:
The inline initialization of static members causes warnings like 

```
LINK : warning C4744: 'static struct std::atomic<bool> silkworm::SignalHandler::signalled_' has different type in 'd:\github\silkworm\node\silkworm\stagedsync\stage_senders.cpp' and 'd:\github\silkworm\cmd\toolbox.cpp': '__declspec(align(1)) struct (1 bytes)' and 'struct (1 bytes)' [C:\Users\Andrea\CMakeBuilds\silkworm\Debug\cmd\toolbox.vcxproj]
```

Revert to initialization in own compile unit

### DIFF
--- a/node/silkworm/common/signal_handler.cpp
+++ b/node/silkworm/common/signal_handler.cpp
@@ -100,6 +100,9 @@ inline constexpr int kHandleableCodes[] {
         SIGTERM  // Termination request (kill/killall default)
 };
 
+std::atomic_uint32_t SignalHandler::sig_count_{0};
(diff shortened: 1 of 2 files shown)


greg7mdp:
Yes, makes sense, that way we are sure that there is only one instance of these variables and we don't rely on the linker to merge them.

yperbasis (approved):

PR #492, opened 2021-11-10
Move rlp::encode(vector) into encode_vector.hpp

yperbasis:
This allows to avoid forward declarations in `rlp/encode.hpp`.

Also delete copy-pasted `RLPVectorCoding.hpp`.


core/silkworm/rlp/encode.hpp
```diff
@@ -29,95 +29,48 @@
...
+void encode_header(Bytes& to, Header header);
+
+void encode(Bytes& to, const evmc::bytes32&);
+void encode(Bytes& to, ByteView);
+void encode(Bytes& to, uint64_t);
+void encode(Bytes& to, const intx::uint256&);
+
+template <size_t N>
+void encode(Bytes& to, gsl::span<const uint8_t, N> bytes) {
+    static_assert(N <= 55, "Complex RLP length encoding not supported");
```
greg7mdp, line 51: to.reserve(to.size() + N + 1); // avoid double memory allocation

core/silkworm/rlp/encode_vector.hpp
```diff
@@ -0,0 +1,49 @@
...
+    return length_of_length(payload_length) + payload_length;
+}
+
+template <class T>
+void encode(Bytes& to, const std::vector<T>& v) {
+    Header h{true, 0};
+    for (const T& x : v) {
+        h.payload_length += length(x);
+    }
+    encode_header(to, h);
```
greg7mdp, line 41: to.reserve(to.size() + h.payload_length()); // avoid multiple memory allocations

greg7mdp (commented):
Suggesting reserve() calls to avoid multiple separate memory allocations.

core/silkworm/rlp/encode_vector.hpp
greg7mdp: actually line should be before encode_header()

greg7mdp (commented):
not very good at this!

core/silkworm/consensus/base/engine.cpp
```diff
@@ -68,7 +69,8 @@ ValidationResult ConsensusEngineBase::pre_validate_block(const silkworm::Block&
             return ValidationResult::kNotAnOmmer;
         }
 
-        if (as_range::any_of(old_ommers, [&ommer](const BlockHeader& old_ommer) -> bool { return old_ommer == ommer; })) {
+        if (as_range::any_of(old_ommers,
+                             [&ommer](const BlockHeader& old_ommer) -> bool { return old_ommer == ommer; })) {
```
greg7mdp, line 73: I think:
```
if (as_range::find(old_ommers, ommer) != old_ommers.end()) {
```
is clearer as to the intent

greg7mdp (commented):
use clearer algorithm.

core/silkworm/consensus/base/engine.cpp
yperbasis: Nice! Done in https://github.com/torquem-ch/silkworm/pull/492/commits/149f6666bccf05b3d17aca753650780082624d71

core/silkworm/rlp/encode.hpp
yperbasis: Sure, done in https://github.com/torquem-ch/silkworm/pull/492/commits/c3f994b303b254e913c240ef43929b1ec94e59ff

core/silkworm/rlp/encode_vector.hpp
yperbasis: Good suggestion! Done in https://github.com/torquem-ch/silkworm/pull/492/commits/c3f994b303b254e913c240ef43929b1ec94e59ff after `encode_header()` because `h.payload_length` doesn't include the length of the RLP header itself.

core/silkworm/rlp/encode_vector.hpp
```diff
@@ -0,0 +1,50 @@
...
+    }
+    return length_of_length(payload_length) + payload_length;
+}
+
+template <class T>
+void encode(Bytes& to, const std::vector<T>& v) {
+    Header h{true, 0};
+    for (const T& x : v) {
+        h.payload_length += length(x);
+    }
```
AndreaLanfranchi, line 40: Shouldn't this be
```suggestion
    Header h{true, lenght(v)};
```
?

yperbasis: Nope because `length(v) == length_of_length(payload_length) + payload_length`

node/silkworm/downloader/packets/BlockBodiesPacket.hpp
```diff
@@ -30,10 +30,6 @@ struct BlockBodiesPacket66 {  // eth/66 version
 
 namespace rlp {
 
-    size_t length(const BlockBodiesPacket& from) noexcept;
-
-    void encode(Bytes& to, const BlockBodiesPacket& from);
-
```
mriccobene, line 36: It is great that we now can delete these functions. 
Can we also get rid of the decode one? 
I suppose you have not deleted it because we have problems defining 
`rlp::decode(ByteView& from, vector<T>& to) `
due to the fact that the decode overload set is really a set of a template function specializations. Why it is so? I looked at it in the past but I did not found a reason for this decision.

mriccobene (approved):

core/silkworm/rlp/encode_vector.hpp
yperbasis: I've refactored the code in https://github.com/torquem-ch/silkworm/pull/492/commits/f59e4201a677d17a500c76abbbff23cbcf77464c to reduce code duplicate between length & encode. Hopefully it's clearer now.

As to std:accumulate, between
```
        for (const T& x : v) {
            h.payload_length += length(x);
        }
```
and
```
        h.payload_length =
            std::accumulate(v.begin(), v.end(), size_t{0}, [](size_t sum, const T& x) { return sum + length(x); });
```
I find the former more readable.

node/silkworm/downloader/messages/OutboundNewBlockHashes.cpp
```diff
@@ -17,62 +17,62 @@ limitations under the License.
...
+                                if err != nil {
+                                        log.Error("propagateNewBlockHashes", "error", err)
+                                }
+                        }
+                default:
+                        //??
+                }
+        }
 }
  */
```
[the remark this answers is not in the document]
yperbasis: Done.

node/silkworm/downloader/packets/BlockBodiesPacket.hpp
yperbasis: Right, I've created Issue #493.

PR #498, opened 2021-11-15
Remove code duplication between rlp::decode(uint64_t) & decode(uint256)

yperbasis:
As a prerequisite, `from_big_compact_u64` & `from_big_compact_u256` are unified into template `from_big_compact`.

### DIFF
--- a/core/silkworm/common/endian.hpp
+++ b/core/silkworm/common/endian.hpp
@@ -162,7 +162,7 @@ inline void store_big_u64(uint8_t* bytes, const uint64_t value) {
 //! \return A string of bytes
 //! \remarks See Erigon TxIndex value
 //! \remarks A "compact" big endian form strips leftmost bytes valued to zero
-Bytes to_big_compact(const uint64_t value);
+Bytes to_big_compact(uint64_t value);
 
 //! \brief Transforms a uint256 stored in memory with native endianness to it's compacted big endian byte form
 //! \param [in] value : the value to be transformed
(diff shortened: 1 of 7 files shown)


core/silkworm/common/endian.hpp
```diff
@@ -171,19 +171,36 @@ Bytes to_big_compact(const uint64_t value);
...
+    if (data.empty()) {
+        return x;
+    }
+
+    if (data[0] == 0 && !allow_leading_zeros) {
+        return std::nullopt;
+    }
+
+    auto* ptr{reinterpret_cast<uint8_t*>(&x)};
+    std::memcpy(ptr + (sizeof(UnsignedInteger) - data.length()), &data[0], data.length());
```
greg7mdp, line 196: If allow_leadind_zeros is true, then I think this memcpy would be incorrect as the `ptr + (sizeof(UnsignedInteger) - data.length())` assumes that all the bytes from `data` are significant, right?

greg7mdp (commented):
Nice change! Always a good feeling to remove duplicated code and make the implementation more generic. Just have a small question about the from_big_compact function.

core/silkworm/common/endian.hpp
yperbasis: It should be correct even in that case. Let's say that `UnsignedInteger` is `uint32_t` and the data is from_hex("00AB"). Then after `memcpy(ptr + 2, &data[0], 2)` `x` will have bytes 0, 0, 0, 0xAB, which is the desired big-endian number. But I'll add a test case just to be on the safe side.

core/silkworm/common/endian.hpp
```diff
@@ -72,26 +72,28 @@ as compiler intrinsics to swap bytes in 16-bit, 32-bit, and 64-bit integers resp
...
 };
 #elif SILKWORM_BYTE_ORDER == SILKWORM_BIG_ENDIAN
-struct le {
-    static inline uint16_t uint16(uint16_t value) noexcept { return SILKWORM_BSWAP16(value); }
-    static inline uint32_t uint32(uint32_t value) noexcept { return SILKWORM_BSWAP32(value); }
-    static inline uint64_t uint64(uint64_t value) noexcept { return SILKWORM_BSWAP64(value); }
+struct LE {
+    static inline uint16_t load(uint16_t value) noexcept { return SILKWORM_BSWAP16(value); }
+    static inline uint32_t load(uint32_t value) noexcept { return SILKWORM_BSWAP32(value); }
+    static inline uint64_t load(uint64_t value) noexcept { return SILKWORM_BSWAP64(value); }
```
greg7mdp, line 91: no intx::uint256 version here?

core/silkworm/common/endian.hpp
```diff
@@ -72,26 +72,28 @@ as compiler intrinsics to swap bytes in 16-bit, 32-bit, and 64-bit integers resp
...
 #if SILKWORM_BYTE_ORDER == SILKWORM_LITTLE_ENDIAN
-struct le {
-    static inline uint16_t uint16(uint16_t value) noexcept { return value; }
-    static inline uint32_t uint32(uint32_t value) noexcept { return value; }
-    static inline uint64_t uint64(uint64_t value) noexcept { return value; }
+struct LE {
+    static inline uint16_t load(uint16_t value) noexcept { return value; }
+    static inline uint32_t load(uint32_t value) noexcept { return value; }
+    static inline uint64_t load(uint64_t value) noexcept { return value; }
+    static inline intx::uint256 load(const intx::uint256& value) noexcept { return value; }
```
greg7mdp, line 79: I don't think the const& parameter of the last function is needed since it returns a value anyways and will be inlined. 
Then these 4 could be: 
```
template <class UnsignedInt>
static constexpr UnsignedInt load(UnsignedInt value) noexcept { return value; }
```

core/silkworm/common/endian.hpp
yperbasis: No because I suspect that intx::uint256 might have some issues on big-endian platforms. In any case, currently there are some places in the Silkworm codebase where we assume that the platform is little-endian. IMHO supporting big-endian platforms is not a priority.

core/silkworm/common/endian.hpp
```diff
@@ -73,27 +73,34 @@ namespace silkworm::endian {
...
+    // In any case, Silkworm is currently untested on big-endian plaforms.
 };
 struct BE {
-    static inline uint16_t load(uint16_t value) noexcept { return value; }
-    static inline uint32_t load(uint32_t value) noexcept { return value; }
-    static inline uint64_t load(uint64_t value) noexcept { return value; }
+    static constexpr uint16_t load(uint16_t value) noexcept { return value; }
+    static constexpr uint32_t load(uint32_t value) noexcept { return value; }
+    static constexpr uint64_t load(uint64_t value) noexcept { return value; }
 };
```
greg7mdp, line 104: You can use the same template as above here as well. 
Also why use `struct` instead of `namespace` for LE and BE?

mriccobene (approved):

core/silkworm/common/endian.hpp
yperbasis: I'm afraid that intx is in fact mixed-endian on big-endian platforms, so I don't want to define the functions for it there.

Right, I've changed `struct` to `namespace`.

AndreaLanfranchi: intx has its own endian swapper in its own namespace

yperbasis: > intx has its own endian swapper in its own namespace

Which I believe is broken on big-endian platforms. In any case, IMHO supporting big-endian platforms is a low priority.

PR #500, opened 2021-11-16
Run consensus tests in parallel

yperbasis:
To enable parallel execution, I added a thread pool based on https://github.com/bshoshany/thread-pool. I had to modify it and use boost::thread instead of std::thread to support custom stack size. Default macOS stack size doesn't work for deep EVM calls.

After this change the running time of `cmd/consensus` on my macBook is reduced from 50 sec to 6 sec.


AndreaLanfranchi (commented):
We have StopWatch class in common to print human readable timings

AndreaLanfranchi (approved):

cmd/consensus.cpp
```diff
@@ -53,10 +52,12 @@ static const fs::path kBlockchainDir{"BlockchainTests"};
...
 static const fs::path kTransactionDir{"TransactionTests"};
 
-static const std::vector<fs::path> kExcludedTests{
-    // Very slow tests
+static const std::vector<fs::path> kSlowTests{
     kBlockchainDir / "GeneralStateTests" / "stTimeConsuming",
+    kBlockchainDir / "GeneralStateTests" / "VMTests" / "vmPerformance" / "loopMul.json",
+};
 
+static const std::vector<fs::path> kFailingTests{
```
chfast, line 60: I believe @gumb0 is working on fixing these tests upstream.

cmd/consensus.cpp
```diff
@@ -633,17 +627,26 @@ Status difficulty_test(const nlohmann::json& j, const std::optional<ChainConfig>
...
+    const auto begin{std::chrono::steady_clock::now()};
+
     CLI::App app{"Run Ethereum consensus tests"};
+
     std::string evm_path{};
     app.add_option("--evm", evm_path, "Path to EVMC-compliant VM");
     std::string tests_path{SILKWORM_CONSENSUS_TEST_DIR};
-    app.add_option("--tests", tests_path, "Path to consensus tests", true)->check(CLI::ExistingDirectory);
+    app.add_option("--tests", tests_path, "Path to consensus tests", /*defaulted=*/true)->check(CLI::ExistingDirectory);
+    unsigned num_threads{std::thread::hardware_concurrency()};
```
chfast, line 645: We should check what it returns in CI/docker. CMake had/has issue with similar context where it returns the number of hardware threads of a CPU even though it runs in a virtualized / containerized environment with number of CPU cores capped.

chfast (approved):

cmd/consensus.cpp
gumb0: Yes but only nonce tests.
I believe the next test release will include nonce larger than 2^64-1 being forbidden.
But also the new rule: nonce == 2^64-1 is fobidden, too, see updated [EIP-2681](https://eips.ethereum.org/EIPS/eip-2681).

PR #502, opened 2021-11-18
C++20 & gcc 11 compatibility

yperbasis:
Make Silkworm compatible with gcc 11 & C++20 (C++17 is still the default).


canepat (approved):
LGTM

Maybe it could be worth it naming the CMake files more explicitly, i.e. rename `toolchain.cmake` as `cpp17.cmake` and move common settings in a generic `toolchain.cmake` included by both `cpp17.cmake` and `cpp20.cmake`. So that also CI configurations can be made more precise and intuitive

PR #505, opened 2021-11-19
Draft of initial structure

AndreaLanfranchi:


yperbasis:
> One thing I have been wondering is why we have the `silkworm` hierarchy under `node` and `core`. I don't quite see the benefit.

It's touched upon in https://github.com/torquem-ch/silkworm#codemap. Basically we want to be able to compile the execution logic ("core") into WebAssembly (which doesn't support exceptions and multi-threading yet), while in the rest of the code ("node") we use exceptions and multi-threading.

greg7mdp:
> It's touched upon in https://github.com/torquem-ch/silkworm#codemap. Basically we want to be able to compile the execution logic ("core") into WebAssembly (which doesn't support exceptions and multi-threading yet), while in the rest of the code ("node") we use exceptions and multi-threading.

Yes, but do we really need the extra `silkworm` level?

yperbasis:
> Yes, but do we really need the extra `silkworm` level?

Oh, it's for include paths. The idea is that you can add `$SILKWORM_MAIN_DIR/core` & `$SILKWORM_MAIN_DIR/node` directories to your include path list and includes like `<silkworm/chain/difficulty.hpp>` will work fine.

greg7mdp:
> Oh, it's for include paths. The idea is that you can add `$SILKWORM_MAIN_DIR/core` & `$SILKWORM_MAIN_DIR/node` directories to your include path list and includes like `<silkworm/chain/difficulty.hpp>` will work fine.

I actually think it would be cleaner (and more expressive) to just have the top level in the include path, remove the silkworm level, and include things like `#include <node/chain/difficulty.hpp>`, because that way you'd know immediately if the file is in `node` or in `core`.

core/silkworm/chain/config_test.cpp
```diff
@@ -20,7 +20,7 @@
 
 namespace silkworm {
 
-TEST_CASE("Config lookup") {
+TEST_CASE("Settings lookup") {
```
yperbasis, line 23: IMHO "chain config" sounds better than "chain settings". Erigon also uses "chain config".

node/silkworm/common/log.hpp
```diff
@@ -17,60 +17,79 @@
...
+    Trace      // Trace calls to functions
+};
+
+//! \brief Holds logging configuration
+struct Settings {
+    bool log_std_out{false};    // Whether console logging goes to std::cout or std::cerr (default)
+    bool log_utc{false};        // Whether timestamps should be in UTC or imbue local timezone
+    bool log_nocolor{false};    // Whether to disable colorized output
+    bool log_threads{false};    // Whether to print thread ids in log lines
+    unsigned log_verbosity{4};  // Log verbosity level
```
yperbasis, line 44: Please use the enum value instead of the magic constant.

core/silkworm/chain/config_test.cpp
AndreaLanfranchi: Right ... I didn't notice that. I believe is a byproduct of a replacement.
Amended in 73d359e

node/silkworm/common/log.cpp
```diff
@@ -14,45 +14,117 @@
...
 
-std::ostream& null_stream() {
-    static struct null_buf : public std::streambuf {
-        int overflow(int c) override { return c; }
-    } null_buf;
-    static struct null_strm : public std::ostream {
-        null_strm() : std::ostream(&null_buf) {}
-    } null_strm;
-    return null_strm;
+void set_verbosity(Level level) { settings_.log_verbosity = static_cast<unsigned>(level); }
```
yperbasis, line 69: Currently this method is not thread-safe. I guess it's fine, but it should be documented.

node/silkworm/common/log.cpp
```diff
@@ -14,45 +14,117 @@
...
+    }
+#endif
+}
 
-std::ostream& log_::header_(LogLevel level) {
-    log_streams_ << kLogTags_[static_cast<int>(level)] << "["
-                 << absl::FormatTime("%m-%d|%H:%M:%E3S", absl::Now(), absl::LocalTimeZone()) << "]";
-    if (log_thread_enabled_) {
-        log_streams_ << " " << std::this_thread::get_id();
+void tee_file(std::filesystem::path path) {
```
yperbasis, line 61: Currently this method is not thread-safe. I guess it's fine, but it should be documented.

Also, it's better pass `path` by `const &`.

node/silkworm/common/log.cpp
AndreaLanfranchi: As far as I am concerned the verbosity is set at process startup and never changed hence no need to make it thread-safe.
Will add a comment

node/silkworm/common/log.cpp
AndreaLanfranchi: About thread safety see same comment for verbosity.

node/silkworm/common/log.hpp
```diff
@@ -17,60 +17,79 @@
...
 
-}  // namespace silkworm
+template <Level level>
+class LogBuffer : public BufferBase {
+  public:
+    LogBuffer() : BufferBase(level){};
+};
+
+using TraceChannel = LogBuffer<Level::Trace>;
+using DebugChannel = LogBuffer<Level::Debug>;
```
yperbasis, line 86: I would even drop "Channel" to simplify call sites to something like
```
log::Info() << "Initializing Light Cache for DAG epoch " << epoch_num;
```
instead of 
```
log::InfoChannel() << "Initializing Light Cache for DAG epoch " << epoch_num;
```

AndreaLanfranchi: I thought about that too but it causes an odd behavior on CLion.
Basically it (CLion) confuses the class with the value of LogLevel enum and improperly colorizes the line to "used" when it "might" be unused

yperbasis: Not a problem if we prefix the enum values with `k` per Google's style guide.

node/silkworm/common/settings.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <silkworm/common/directories.hpp>
+#include <silkworm/db/mdbx.hpp>
+#include <silkworm/db/storage.hpp>
+
+namespace silkworm {
+
+struct NodeSettings {
+    std::unique_ptr<DataDirectory> data_directory;    // Pointer to data folder
+    db::EnvConfig chaindata_config{};                 // Chaindata db config
+    uint64_t network_id{1};                           // Network/Chain id
```
yperbasis, line 33: `kMainnetConfig.chain_id` would be better for clarity

node/silkworm/db/tables.hpp
```diff
@@ -55,7 +55,7 @@ inline constexpr db::MapConfig kCallToIndex{"CallToIndex"};
 inline constexpr db::MapConfig kCallTraceSet{"CallTraceSet", mdbx::key_mode::usual, mdbx::value_mode::multi};
 inline constexpr db::MapConfig kClique{"Clique"};
 inline constexpr db::MapConfig kCode{"Code"};
-inline constexpr db::MapConfig kConfig{"Config"};
+inline constexpr db::MapConfig kConfig{"Settings"};
```
yperbasis, line 58: Is the DB table called "Settings" now? Shall we rename the variable accordingly?

yperbasis (approved):
Looks good to me overall.

node/silkworm/db/tables.hpp
AndreaLanfranchi: No. Damn replacer.

node/silkworm/common/log.hpp
AndreaLanfranchi: Done in 0186377

node/silkworm/common/log.cpp
AndreaLanfranchi: Done in 8bb730b

node/silkworm/common/settings.hpp
AndreaLanfranchi: Done in 6492076

node/silkworm/db/tables.hpp
AndreaLanfranchi: Amended bf6d478

node/silkworm/common/log.hpp
AndreaLanfranchi: Done in 4a51321 and 4b131a1

AndreaLanfranchi:
Still lot of work to be done but merging now so other PRs and branches can catch-up

PR #508, opened 2021-11-21
Use colour constants in cmd/consensus

yperbasis:


AndreaLanfranchi (commented):
At least on Windows though you must ensure output code page is set to UTF-8 otherwise you get the printout of escape sequence like this
```
←[32m49995 tests passed←[0m, 0 failed, 4 skipped in 3m:28.570s
```
This happens when the executable is ran in _normal_ cmd prompt window while if the user has installed the "Terminal" app the output is ok.

AndreaLanfranchi (approved):

PR #510, opened 2021-11-22
Structure step 2 - silkworm binary

AndreaLanfranchi:
* Given process will open several databases (chaindata, clique, nodes) rename some containers with their context.
* `db::table::create_all` method renamed -> `db::table::deploy_chaindata_tables` with strong check on flags compatibility
* silkworm binary now checks if db is initialized with genesis : if not **and** the chain id is known (we have the genesis file) db is automatically initialized with it
* consistency check amongst CLI arguments `--chain` and/or `--networkid` to match the `chain_config` persisted in db (if any)
* rename `stagedsync::TransactionManager` -> `db::RWTxn`


node/silkworm/db/tables.hpp
```diff
@@ -139,8 +137,8 @@ inline constexpr db::MapConfig kTables[]{
     kTxLookup,
 };
 
-// Create all tables that do not yet exist.
-void create_all(mdbx::txn& txn);
+//! \brief Ensures all defined tables are present in db with consistent flags. Should a table not exist it gets created
+void deploy_chaindata_tables(mdbx::txn& txn);
```
yperbasis, line 141: Hmm. Perhaps "create_chaindata_tables"? "deploy" is less clear IMHO.

AndreaLanfranchi: Well actually the method does both a check (if table exist it must have compatible flags) or a creation (if table does not exist). I cant find a single word to describe it. Maybe `check_or_create_chaindata_tables` ?

yperbasis: Well, IMHO simply "create" is fine and the checks are just details. For example, we typically call methods just something like "cleanup", even though `cleanup` is probably idempotent and has some checks internally against double cleanup. But I am not insisting.

cmd/silkworm.cpp
```diff
@@ -214,37 +215,85 @@ int main(int argc, char* argv[]) {
...
         chaindata_env.close();
 
     } catch (const CLI::ParseError& ex) {
         return cli.exit(ex);
+    } catch (const std::runtime_error& ex) {
+        log::Error() << ex.what();
+        return -1;
     } catch (const std::invalid_argument& ex) {
-        std::cerr << "Invalid argument :" << ex.what() << "\n" << std::endl;
+        std::cerr << "\tInvalid argument :" << ex.what() << "\n" << std::endl;
```
yperbasis, line 288: Why not `log::Error` ?

yperbasis (approved):

cmd/silkworm.cpp
AndreaLanfranchi: Cause this catch is related to the `try` block where log initialization occurs. We might not have it ready yet.

PR #512, opened 2021-11-28
Merge consensus engine + EIP-4399

yperbasis:
Implement [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) & [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399)

### DIFF
--- a/core/silkworm/chain/config.hpp
+++ b/core/silkworm/chain/config.hpp
@@ -22,6 +22,7 @@
 #include <optional>
 
 #include <evmc/evmc.h>
+#include <intx/intx.hpp>
 #include <nlohmann/json.hpp>
 
 #include <silkworm/common/util.hpp>
@@ -70,6 +71,9 @@ struct ChainConfig {
 
     // https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/arrow-glacier.md
     std::optional<uint64_t> arrow_glacier_block{std::nullopt};
+
+    // PoW to PoS switch; see EIP-3675
+    std::optional<intx::uint256> terminal_total_difficulty{std::nullopt};
 
     // Returns the revision level at given block number
     // In other words, on behalf of Json chain config data
--- a/core/silkworm/consensus/merge/engine.cpp
+++ b/core/silkworm/consensus/merge/engine.cpp
@@ -1,0 +1,54 @@
+/*
+   Copyright 2021 The Silkworm Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#include "engine.hpp"
+
+namespace silkworm::consensus {
+
+MergeEngine::MergeEngine(const ChainConfig& chain_config)
+    : terminal_total_difficulty_{*chain_config.terminal_total_difficulty},
+      ethash_engine_{chain_config},
+      pos_engine_{chain_config} {}
+
+ValidationResult MergeEngine::pre_validate_block(const Block&, BlockState&) {
+    // TODO (Andrew) implement
+    return ValidationResult::kUnknownConsensusEngine;
+}
+
+ValidationResult MergeEngine::validate_block_header(const BlockHeader&, BlockState&, bool) {
+    // TODO (Andrew) implement
+    return ValidationResult::kUnknownConsensusEngine;
+}
+
+ValidationResult MergeEngine::validate_seal(const BlockHeader& header) {
+    if (header.difficulty != 0) {
+        return ethash_engine_.validate_seal(header);
+    } else {
+        return pos_engine_.validate_seal(header);
+    }
+}
+
+void MergeEngine::finalize(IntraBlockState& state, const Block& block, evmc_revision revision) {
+    if (block.header.difficulty != 0) {
+        ethash_engine_.finalize(state, block, revision);
+    } else {
+        pos_engine_.finalize(state, block, revision);
+    }
+}
+
+evmc::address MergeEngine::get_beneficiary(const BlockHeader& header) { return header.beneficiary; }
+
+}  // namespace silkworm::consensus
--- a/core/silkworm/chain/config.cpp
+++ b/core/silkworm/chain/config.cpp
@@ -18,6 +18,8 @@
 
 namespace silkworm {
 
+constexpr const char* kTerminalTotalDifficulty{"terminalTotalDifficulty"};
+
 static inline void member_to_json(nlohmann::json& json, const std::string& key, const std::optional<uint64_t>& source) {
     if (source.has_value()) {
         json[key] = source.value();
@@ -59,6 +61,11 @@ nlohmann::json ChainConfig::to_json() const noexcept {
     member_to_json(ret, "muirGlacierBlock", muir_glacier_block);
     member_to_json(ret, "arrowGlacierBlock", arrow_glacier_block);
 
+    if (terminal_total_difficulty.has_value()) {
+        // TODO (Andrew) geth probably treats terminalTotalDifficulty as a JSON number
+        ret[kTerminalTotalDifficulty] = to_string(*terminal_total_difficulty);
+    }
+
     return ret;
 }
 
@@ -87,6 +94,11 @@ std::optional<ChainConfig> ChainConfig::from_json(const nlohmann::json& json) no
     read_json_config_member(json, "daoForkBlock", config.dao_block);
     read_json_config_member(json, "muirGlacierBlock", config.muir_glacier_block);
     read_json_config_member(json, "arrowGlacierBlock", config.arrow_glacier_block);
+
+    if (json.contains(kTerminalTotalDifficulty)) {
+        config.terminal_total_difficulty =
+            intx::from_string<intx::uint256>(json[kTerminalTotalDifficulty].get<std::string>());
+    }
 
     return config;
 }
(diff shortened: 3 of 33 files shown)


AndreaLanfranchi:
@greg7mdp 
probably this is more direct and removes redundant lookups

```
std::optional<BlockBody> InMemoryState::read_body(uint64_t block_number,
                                                  const evmc::bytes32& block_hash) const noexcept {
    const auto& l1{bodies_.find(block_number)};
    if (l1 != bodies_.end()) {
        const auto& l2{l1->second.find(block_hash)};
        if (l2 != l1->second.end()) {
            return l2->second;
        }
    }
    return std::nullopt;
}
```

Pushing the change

AndreaLanfranchi:
> why std::map and not `std::unordered_map`, or even better `absl::flat_hash_map`?

Is still work in progress and honestly I didn't think about container performance yet: I only removed the problematic usage of vector (besides I don't like absl containers at all - matter of taste)

greg7mdp:
> besides I don't like absl containers at all - matter of taste

I don't either, especially because it brings a huge library. This is why I created [phmap](https://github.com/greg7mdp/parallel-hashmap), which brings the speed and memory efficiency of the abseil containers into a small, header-only library. You should use it in silkworm :-).

AndreaLanfranchi:
> > besides I don't like absl containers at all - matter of taste
> 
> I don't either, especially because it brings a huge library. This is why I created [phmap](https://github.com/greg7mdp/parallel-hashmap), which brings the speed and memory efficiency of the abseil containers into a small, header-only library. You should use it in silkworm :-).

Nice to know. Will definitely give it a try. Thank you !

AndreaLanfranchi (approved):

core/silkworm/chain/config.hpp
```diff
@@ -72,6 +73,9 @@ struct ChainConfig {
     // https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/arrow-glacier.md
     std::optional<uint64_t> arrow_glacier_block{std::nullopt};
 
+    // PoW to PoS switch; see EIP-3675
+    std::optional<intx::uint256> terminal_total_difficulty{std::nullopt};
```
greg7mdp, line 77: Maybe a matter of preference, but the initializer `{std::nullopt}` is not needed, as by default `std::optional` is constructed empty (what else could it contain?).

core/silkworm/consensus/merge/engine.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+MergeEngine::MergeEngine(const ChainConfig& chain_config)
+    : terminal_total_difficulty_{*chain_config.terminal_total_difficulty},
+      ethash_engine_{chain_config},
+      pos_engine_{chain_config} {}
+
+ValidationResult MergeEngine::pre_validate_block(const Block& block, const BlockState& state) {
+    if (block.header.difficulty != 0) {
+        return ethash_engine_.pre_validate_block(block, state);
+    } else {
```
greg7mdp, line 29: `else` is not needed here

core/silkworm/consensus/merge/engine.cpp
```diff
@@ -0,0 +1,100 @@
...
+        const std::optional<intx::uint256> parent_total_difficulty{
+            state.total_difficulty(parent->number, header.parent_hash)};
+        if (parent_total_difficulty == std::nullopt) {
+            return ValidationResult::kUnknownParentTotalDifficulty;
+        }
+        if (parent_total_difficulty >= terminal_total_difficulty_) {
+            return ValidationResult::kPoWBlockAfterMerge;
+        }
+        return ethash_engine_.validate_block_header(header, state, with_future_timestamp_check);
+    } else {
```
greg7mdp, line 53: same, `else` not needed here. Maybe a matter of preference.

core/silkworm/consensus/merge/engine.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+bool MergeEngine::terminal_pow_block(const BlockHeader& header, const BlockState& state) const {
+    if (header.difficulty == 0) {
+        return false;  // PoS block
+    }
+
+    const std::optional<BlockHeader> parent{EngineBase::get_parent_header(state, header)};
+    if (parent == std::nullopt) {
```
greg7mdp, line 67: You guys sure you want these ` == std::nullopt` tests everywhere, instead of the simpler `if (!parent)`?

core/silkworm/consensus/merge/engine.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    return parent_total_difficulty < terminal_total_difficulty_ &&
+           *parent_total_difficulty + header.difficulty >= terminal_total_difficulty_;
+}
+
+ValidationResult MergeEngine::validate_seal(const BlockHeader& header) {
+    if (header.difficulty != 0) {
+        return ethash_engine_.validate_seal(header);
+    } else {
```
greg7mdp, line 85: `else` unnecessary

core/silkworm/state/in_memory_state.cpp
```diff
@@ -44,19 +44,17 @@ ByteView InMemoryState::read_code(const evmc::bytes32& code_hash) const noexcept
...
-        return {};
+    const auto it1{storage_.find(address)};
+    if (it1 != storage_.end()) {
+        const auto it2{it1->second.find(incarnation)};
+        if (it2 != it1->second.end()) {
+            const auto it3{it2->second.find(location)};
+            if (it3 != it2->second.end()) {
+                return it3->second;
+            }
+        }
```
greg7mdp, line 55: I personally like the version of the left better, but if we use this style why not have the declarations inside the if test?

```
    if (const auto it1{storage_.find(address)}; it1 != storage_.end()) {
        if (const auto it2{it1->second.find(incarnation)}; it2 != it1->second.end()) {
            if (const auto it3{it2->second.find(location)}; it3 != it2->second.end()) {
                return it3->second;
            }
        }
    }
```

core/silkworm/state/in_memory_state.cpp
```diff
@@ -69,69 +67,60 @@ uint64_t InMemoryState::previous_incarnation(const evmc::address& address) const
...
-
-    auto it{headers_[block_number].find(block_hash)};
-    if (it == headers_[block_number].end()) {
-        return std::nullopt;
+    const auto it1{headers_.find(block_number)};
+    if (it1 != headers_.end()) {
+        const auto it2{it1->second.find(block_hash)};
+        if (it2 != it1->second.end()) {
+            return it2->second;
+        }
```
greg7mdp, line 75: same, why not have the declarations inside the if test?

core/silkworm/state/in_memory_state.cpp
```diff
@@ -69,69 +67,60 @@ uint64_t InMemoryState::previous_incarnation(const evmc::address& address) const
...
-
-    auto it{bodies_[block_number].find(block_hash)};
-    if (it == bodies_[block_number].end()) {
-        return std::nullopt;
+    const auto it1{bodies_.find(block_number)};
+    if (it1 != bodies_.end()) {
+        const auto it2{it1->second.find(block_hash)};
+        if (it2 != it1->second.end()) {
+            return it2->second;
+        }
```
greg7mdp, line 87: same, why not have the declarations inside the if test?

core/silkworm/state/in_memory_state.cpp
```diff
@@ -69,69 +67,60 @@ uint64_t InMemoryState::previous_incarnation(const evmc::address& address) const
...
     }
-    return it->second;
+    return std::nullopt;
```
greg7mdp, line 89: I would prefer `return {}` rather than `return std::nullopt`.

core/silkworm/state/in_memory_state.cpp
```diff
@@ -69,69 +67,60 @@ uint64_t InMemoryState::previous_incarnation(const evmc::address& address) const
...
 std::optional<intx::uint256> InMemoryState::total_difficulty(uint64_t block_number,
                                                              const evmc::bytes32& block_hash) const noexcept {
-    if (block_number >= difficulty_.size()) {
-        return std::nullopt;
+    const auto it1{difficulty_.find(block_number)};
+    if (it1 != difficulty_.end()) {
+        const auto it2{it1->second.find(block_hash)};
+        if (it2 != it1->second.end()) {
+            return it2->second;
+        }
```
greg7mdp, line 99: same, why not have the declarations inside the if test?

core/silkworm/state/in_memory_state.hpp
```diff
@@ -102,15 +102,15 @@ class InMemoryState : public State {
...
     // block number -> hash -> body
-    std::vector<std::unordered_map<evmc::bytes32, BlockBody>> bodies_;
+    std::map<BlockNum, std::unordered_map<evmc::bytes32, BlockBody>> bodies_;
 
     // block number -> hash -> total difficulty
-    std::vector<std::unordered_map<evmc::bytes32, intx::uint256>> difficulty_;
+    std::map<BlockNum, std::unordered_map<evmc::bytes32, intx::uint256>> difficulty_;
 
-    std::vector<evmc::bytes32> canonical_hashes_;
+    std::map<BlockNum, evmc::bytes32> canonical_hashes_;
```
greg7mdp, line 113: `std::map` is a poor container choice here. There might be a performance drop vs the previous vector. A unordered map would be better.

greg7mdp (approved):
Despite my stylistic points, I keep being impressed by the very high quality of silkworm's code. This is truly a impressive project. Congrats guys!

core/silkworm/state/in_memory_state.hpp
yperbasis: `InMemoryState` is only used for tests, so its performance is not that important.

core/silkworm/state/in_memory_state.cpp
yperbasis: IMHO `std::nullopt` is clearer.

core/silkworm/consensus/merge/engine.cpp
yperbasis: It is unnecessary, but I prefer it for symmetry and arguably better readability.

PR #513, opened 2021-11-28
Structure 3

AndreaLanfranchi:
Implement network names for chain config


core/silkworm/chain/config.cpp
```diff
@@ -16,8 +16,19 @@
 
 #include "config.hpp"
 
+#include <functional>
+
+#include <silkworm/common/as_range.hpp>
+
 namespace silkworm {
 
+static const std::vector<std::pair<std::string, const ChainConfig*>> kKnownChainConfigs{
```
yperbasis, line 25: Why not `std::map`?

core/silkworm/chain/config.hpp
```diff
@@ -185,29 +186,11 @@ inline constexpr ChainConfig kGoerliConfig{
...
-inline uint64_t lookup_chain_id_by_name(const std::string& name) noexcept {
-    uint64_t ret{0};
-    if (iequals(name, "mainnet")) ret = kMainnetConfig.chain_id;
-    if (iequals(name, "ropsten")) ret = kRopstenConfig.chain_id;
-    if (iequals(name, "rinkeby")) ret = kRinkebyConfig.chain_id;
-    if (iequals(name, "goerli")) ret = kGoerliConfig.chain_id;
-    return ret;
-}
+//! \brief Looksup a chain config provided its network id or its common name
+const ChainConfig* lookup_chain_config(std::variant<uint64_t, std::string> identifier) noexcept;
```
yperbasis, line 190: IMHO it's simpler to have two overloaded functions `lookup_chain_config(uint64_t)` and `lookup_chain_config(string_view)` instead of using  `std::variant`.

core/silkworm/chain/config.cpp
AndreaLanfranchi: A little bit easier to deal with case insensitive comparator on the name

core/silkworm/chain/config.hpp
AndreaLanfranchi: Two overloads means two lambda functions on search. With this we can rationalize code.

yperbasis (approved):

PR #514, opened 2021-12-01
Erigon's pruning mode

AndreaLanfranchi:
* implements correct parsing and reciprocal exclusions of cli arguments
* implements read/write to db (yet convoluted - but keeps db compat)
* change node_settings

### DIFF
--- a/cmd/silkworm.cpp
+++ b/cmd/silkworm.cpp
@@ -289,7 +289,7 @@ int main(int argc, char* argv[]) {
                 }
                 db::write_prune_mode(*tx, *node_settings.prune_mode);
                 tx.commit();
-                node_settings.prune_mode = std::make_unique<db::PruneMode>(std::move(db::read_prune_mode(*tx)));
+                node_settings.prune_mode = std::make_unique<db::PruneMode>(db::read_prune_mode(*tx));
             }
             log::Message() << "Prune mode " << db_prune_mode.to_string();
         }
--- a/cmd/silkworm.cpp
+++ b/cmd/silkworm.cpp
@@ -291,7 +291,7 @@ int main(int argc, char* argv[]) {
                 tx.commit();
                 node_settings.prune_mode = std::make_unique<db::PruneMode>(db::read_prune_mode(*tx));
             }
-            log::Message() << "Prune mode " << db_prune_mode.to_string();
+            log::Message() << "Prune mode " << node_settings.prune_mode->to_string();
         }
 
         // Do sync stuff here
--- a/cmd/silkworm.cpp
+++ b/cmd/silkworm.cpp
@@ -14,7 +14,7 @@
     limitations under the License.
 */
 
-#include <map>
+#include <memory>
 #include <optional>
 #include <regex>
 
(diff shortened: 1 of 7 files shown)


cmd/silkworm.cpp
```diff
@@ -265,18 +278,18 @@ int main(int argc, char* argv[]) {
...
+                                             node_settings.prune_mode->to_string() + " got " +
                                              db_prune_mode.to_string());
                 }
-                db::write_prune_mode(*tx, node_settings.prune_mode);
+                db::write_prune_mode(*tx, *node_settings.prune_mode);
                 tx.commit();
-                db_prune_mode = db::read_prune_mode(*tx);
+                node_settings.prune_mode = std::make_unique<db::PruneMode>(db::read_prune_mode(*tx));
             }
             log::Message() << "Prune mode " << db_prune_mode.to_string();
```
yperbasis, line 294: Should probably be `<< node_settings.prune_mode->to_string();` instead

node/silkworm/db/storage.hpp
```diff
@@ -15,88 +15,92 @@
...
-};
 
-inline constexpr PruneMode kDefaultPruneMode{
-    true,          // initialized
-    std::nullopt,  // history
-    std::nullopt,  // receipts
-    std::nullopt,  // tx_index
-    std::nullopt   // call_traces
+  private:
+    const bool initialized_;         // Whether this class is initialized or default
```
yperbasis, line 94: I don't like initialized for C++ objects. IMHO if the constructor succeeds, the object is initialised. If the constructor doesn't succeed (i.e. the constructor throws because, for example, a class invariant is not satisfied), no object is constructed. I don't see a point in having a special in-between state "constructed-but-uninitialized".

node/silkworm/db/storage.hpp
```diff
@@ -15,88 +15,92 @@
...
+
+    void to_string(std::string& short_form, std::string& long_form, char prefix) const;
+
+    bool operator==(const BlockAmount& other) const {
+        return enabled_ == other.enabled_ && type_ == other.type_ && value_ == other.value_;
     }
+    bool operator!=(const BlockAmount& other) const { return !(this->operator==(other)); }
+
+  private:
+    const std::optional<BlockNum> value_;
```
yperbasis, line 63: `const` data member are a bit of a headache in C++: https://lesleylai.info/en/const-and-reference-member-variables. TBH, I wouldn't bother and just use the simplest approach
```
struct Point {
    int x = 0;
    int y = 0;
};

const Point immutable_point {42, 55};
```

node/silkworm/db/storage.hpp
```diff
@@ -15,88 +15,92 @@
...
-#define SILKWORM_STORAGE_HPP_
+#ifndef SILKWORM_DB_STORAGE_HPP_
+#define SILKWORM_DB_STORAGE_HPP_
 
 #include "mdbx.hpp"
 #include "tables.hpp"
 
 namespace silkworm::db {
 
+inline constexpr const BlockNum kFullImmutabilityThreshold{90'000};
```
yperbasis, line 26: Looks like a duplicate of now unused `kDefaultPruneThreshold`

yperbasis: And `constexpr const` is a bit of an overkill :)

node/silkworm/db/storage.cpp
```diff
@@ -20,126 +20,184 @@ limitations under the License.
...
-
-    /*
-     * TODO Eventually optimize with removal of keys which don't prune at all
-     * For compatibility reasons with Erigon we assume a value == UINT64_MAX means no pruning
-     */
-
-    // History
-    auto data{src.find(mdbx::slice{kPruneModeHistoryKey}, /*throw_notfound=*/false)};
+//! \brief Retrieves the proper BlockAmount prune threshold for given key
+static BlockAmount read_block_amount_for_key(mdbx::cursor_managed& source, const char* key) {
```
yperbasis, line 24: (Just nitpicking) It'd be cleaner to pass `source` as `mdbx::cursor&` rather than `mdbx::cursor_managed&` since this function doesn't care if the cursor as managed or not.

yperbasis (approved):

node/silkworm/db/storage.cpp
AndreaLanfranchi: Yes you're right
Doing

PR #523, opened 2021-12-14
Use .proto files from ledgerwatch/interfaces submodule

yperbasis:
And remove the .proto files copied to `silkworm/interfaces`. Rationale: consistency.


mriccobene (commented):
ok, but why did you choose erigon-lib and not https://github.com/ledgerwatch/interfaces ?

yperbasis:
> ok, but why did you choose erigon-lib and not https://github.com/ledgerwatch/interfaces ?

Right, I did not realize that interfaces has its own repo and is a subtree in erigon-lib.

mriccobene (approved):

PR #524, opened 2021-12-17
cmd/consensus depends on node now and can't be built if SILKWORM_CORE_ONLY

yperbasis:
after PR #500

### DIFF
--- a/cmd/CMakeLists.txt
+++ b/cmd/CMakeLists.txt
@@ -115,9 +115,6 @@ if(NOT SILKWORM_CORE_ONLY)
   # Benchmarks
   add_subdirectory(benchmark)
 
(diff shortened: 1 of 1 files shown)


chfast (approved):
This was previously undetected because you use `target_link_libraries(consensus PRIVATE silkworm_node)`. If `silkworm_node` is not a target CMake will just try to link with system library `-lsilkworm_node`.
To disable legacy CMake behavior you can define an alias `add_library(silkworm::silkworm_node ALIAS silkworm_node)` and then only use the alias for linking: `target_link_libraries(consensus PRIVATE silkworm::silkworm_node)`. The names with `::` must be a CMake target.

PR #527, opened 2021-12-30
Structure 4 - Initial working syncloop

AndreaLanfranchi:


cmake/compiler_settings.cmake
```diff
@@ -46,7 +46,7 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   add_compile_options(/wd4068) # Silence warning C4068: unknown pragma
   add_compile_options(/wd5030) # Silence warning C5030: unknown gnu/clang attribute
   add_compile_options(/W4)     # Display all other un-silenced warnings
-  add_compile_options(/bigobj) # Increase .obj sections, needed for hard coded pre-verified hashes
+  # add_compile_options(/bigobj) # Increase .obj sections, needed for hard coded pre-verified hashes
```
yperbasis, line 49: Why remove everything from `preverified_hashes_mainnet.cpp`? Doesn't seem necessary for this PR.

AndreaLanfranchi: This is still a draft and the commit will be reverted when ready for merge.
I had to do this cause with all preverified hashes it takes ages on my machine to build both the windows binary and the linux one.

node/silkworm/common/settings.hpp
```diff
@@ -29,14 +31,19 @@
...
     uint64_t network_id{kMainnetConfig.chain_id};    // Network/Chain id
+    std::optional<ChainConfig> chain_config;         // Chain config
     size_t batch_size{512_Mebi};                     // Batch size to use in stages
     size_t etl_buffer_size{256_Mebi};                // Buffer size for ETL operations
     std::string private_api_addr{"127.0.0.1:9090"};  // Default API listener
-    bool fake_pow{false};                            // Whether to check for PoW or not
+    std::string sentry_api_addr{};                   // Default address(es) of sentry
+    bool fake_pow{false};                            // Whether to verify Proof-of-Work
     std::unique_ptr<db::PruneMode> prune_mode;       // Prune mode
+    uint32_t sync_loop_throttle;                     // Minimum interval amongst sync cycle
```
yperbasis, line 45: Let's initialize to something, even if only 0, in order to prevent potentially unititlized value.

cmd/common.cpp
```diff
@@ -0,0 +1,309 @@
...
+    db::RWTxn tx(chaindata_env);
+
+    // Ensures all tables are present
+    db::table::check_or_create_chaindata_tables(*tx);
+    log::Message("Database schema", {"version", db::read_schema_version(*tx)->to_string()});
+
+    // Check db is initialized with chain config
+    {
+        node_settings.chain_config = db::read_chain_config(*tx);
+        while (!node_settings.chain_config.has_value()) {
```
yperbasis, line 255: Do we need `while` here? Perhaps `if` would do?

cmd/silkworm.cpp
```diff
@@ -14,291 +14,109 @@
...
-        ->default_val(std::to_string(static_cast<uint32_t>(log_settings.log_verbosity)));
-    log_opts.add_flag("--log.stdout", log_settings.log_std_out, "Outputs to std::out instead of std::err");
-    log_opts.add_flag("--log.nocolor", log_settings.log_nocolor, "Disable colors on log lines");
-    log_opts.add_flag("--log.utc", log_settings.log_utc, "Prints log timings in UTC");
-    log_opts.add_flag("--log.threads", log_settings.log_threads, "Prints thread ids");
-    log_opts.add_option("--log.file", log_settings.log_file, "Tee all log lines to given file name");
+using namespace silkworm;
 
-    cli.parse(argc, argv);
+static std::atomic<size_t> s_allocated_memory{0};
```
yperbasis, line 32: What about memory allocated not with `operator new`, but with `malloc` itself? And I don't like operator overloading, feels too hacky. Using something like `malloc_stats` would probably be cleaner. I'd suggest to scrap allocated memory counting for now and do it later in a separate PR.

cmd/silkworm.cpp
```diff
@@ -14,291 +14,109 @@
...
-    node_settings.data_directory = std::make_unique<DataDirectory>(datadir, /*create=*/true);
-    node_settings.chaindata_env_config.max_size = parse_size(chaindata_max_size).value();
-    node_settings.chaindata_env_config.growth_size = parse_size(chaindata_growth_size).value();
-    if (node_settings.chaindata_env_config.growth_size > node_settings.chaindata_env_config.max_size / 2) {
-        throw std::invalid_argument("--chaindata.growthsize too wide");
-    }
+void* operator new(size_t size) {
+    s_allocated_memory += size;
+    return malloc(size);
+}
```
greg7mdp, line 37: I agree with what @yperbasis  wrote. In addition The C++ standard specifies that operator new [throws std::bad_alloc or another exception derived from std::bad_alloc (since C++11) on failure to allocate memory](https://en.cppreference.com/w/cpp/memory/new/operator_new), in addition to other requirements such as retries and alignment. This overload doesn't implement the specification.

cmd/silkworm.cpp
```diff
@@ -14,291 +14,109 @@
...
+void operator delete[](void* ptr, size_t size) noexcept {
+    s_allocated_memory -= size;
+    free(ptr);
+}
 
-    node_settings.prune_mode = db::parse_prune_mode(prune_mode,  //
-                                                    olderHistory, olderReceipts, olderTxIndex, olderCallTraces,
-                                                    beforeHistory, beforeReceipts, beforeTxIndex, beforeCallTraces);
+void operator delete(void* ptr) noexcept {
+    s_allocated_memory -= sizeof(ptr);
```
greg7mdp, line 55: This is not semantically correct. The memory released is the memory pointed to by ptr, not the ptr itself, so `s_allocated_memory -= sizeof(ptr);` is not correct.

cmd/silkworm.cpp
AndreaLanfranchi: This is correct in principle but not in real life.
Modern OSes allow over-allocation and both `new` and `malloc` can return valid pointers even if there's not enough memory left hence `std::bad_alloc` might not throw and cannot be considered as a reliable sign of memory exhaustion. What **it does happen instead** is that if you try to **access** memory pointed by an over-allocated segment you get a segfault which is not catcheable (you can handle the signal but cannot resume the program).

cmd/silkworm.cpp
AndreaLanfranchi: This is a quick and dirty way to track the heap in a **portable** way. Is not meant to be definitive and is useful to me to track the sync loop I'm dealing with. Will gladly scrap it all when ready for review.
Besides **our** code does not make use of `malloc` (submodules do - but this is outside the scope of my actual nvestigations)

cmd/silkworm.cpp
AndreaLanfranchi: I honestly threw in some stuff there only to ensure build.
Without it, in fact, the project complains 
```
error: the program should also define ‘void operator delete [](void*)’ [-Werror=sized-deallocation]
   49 | void operator delete[](void* ptr, size_t size) noexcept {
      |      ^~~~~~~~
```
While for MSVC I could use `_msize` to get the segment size pointed by ptr I didn't pay much attention as, being this PR still a draft, I will polish knobs later on.

AndreaLanfranchi: Besides, to my knowledge, there is no direct way to know the size of a segment pointed by a raw ptr **unless you override memory allocations and prepend each segment with its own size**

node/silkworm/common/settings.hpp
AndreaLanfranchi: Right. Done

cmd/common.cpp
AndreaLanfranchi: Done

cmd/silkworm.cpp
AndreaLanfranchi: `malloc_stats` is also *nix only

AndreaLanfranchi: Removed by 1fef176

cmd/silkworm.cpp
AndreaLanfranchi: Removed by 1fef176

cmd/silkworm.cpp
AndreaLanfranchi: Removed by 1fef176

node/silkworm/etl/buffer.hpp
```diff
@@ -46,18 +46,18 @@ class Buffer {
 
     void put(Entry&& entry) {
         // Add a new entry to the buffer
-        size_ += entry.size();
+        size_ += entry.size() + sizeof(head_t);
```
greg7mdp, line 49: shouldn't the same change be added to the other `put` overload? `void put(const Entry& entry)`

AndreaLanfranchi:
@greg7mdp 

> Just curious, what is the reason for this change?

Is mainly aesthetic. When buffer flushes a file it reports the size of the flushed file which, in many cases, is a couple of megabytes over the chosen etl buffer size.

node/silkworm/common/log.hpp
```diff
@@ -53,13 +54,20 @@ void init(Settings& settings);
...
+bool test_verbosity(Level level);
+
 //! \brief Sets a file output for log teeing
 //! \note This function is not thread safe as it's meant to be used at start of process and never called again
 void tee_file(const std::filesystem::path& path);
 
 class BufferBase {
   public:
     explicit BufferBase(Level level);
+    explicit BufferBase(Level level, std::string_view msg, std::vector<std::string> args);
```
yperbasis, line 70: Pass args by `const &` ?

node/silkworm/common/log.cpp
```diff
@@ -105,6 +126,7 @@ void BufferBase::flush() {
         }
         *file_ << line << std::endl;
     }
+    out_lck.unlock();
```
yperbasis, line 129: Unnecessary (will be unlocked by the destructor)

node/silkworm/common/settings.hpp
```diff
@@ -29,14 +31,19 @@
...
     std::unique_ptr<DataDirectory> data_directory;   // Pointer to data folder
     db::EnvConfig chaindata_env_config{};            // Chaindata db config
     uint64_t network_id{kMainnetConfig.chain_id};    // Network/Chain id
+    std::optional<ChainConfig> chain_config;         // Chain config
     size_t batch_size{512_Mebi};                     // Batch size to use in stages
     size_t etl_buffer_size{256_Mebi};                // Buffer size for ETL operations
     std::string private_api_addr{"127.0.0.1:9090"};  // Default API listener
-    bool fake_pow{false};                            // Whether to check for PoW or not
+    std::string sentry_api_addr{};                   // Default address(es) of sentry
+    bool fake_pow{false};                            // Whether to verify Proof-of-Work
```
yperbasis, line 43: Has no effect at the moment. Perhaps we can remove `fake_pow` for now and add it later need be.

node/silkworm/db/mdbx.hpp
```diff
@@ -58,10 +58,24 @@ class RWTxn {
     mdbx::txn& operator*() { return external_txn_ ? *external_txn_ : managed_txn_; }
     mdbx::txn* operator->() { return external_txn_ ? external_txn_ : &managed_txn_; }
 
-    void commit() {
+    void commit(const bool renew = true) {
+        /*
+         * renew is required here due to RAII
+         * RWTxn txn(env);
+         * txn.commit();
+         * env.close();
```
yperbasis, line 66: Why call `env.close()` explicitly? If we rely on `env`'s destructor, `env` will be closed after `tnx`'s destructor.

node/silkworm/stagedsync/sync_loop.hpp
```diff
@@ -0,0 +1,47 @@
...
+#ifndef SILKWORM_STAGEDSYNC_SYNCLOOP_HPP_
+#define SILKWORM_STAGEDSYNC_SYNCLOOP_HPP_
+
+#include <silkworm/common/settings.hpp>
+#include <silkworm/common/stopwatch.hpp>
+#include <silkworm/concurrency/worker.hpp>
+#include <silkworm/db/mdbx.hpp>
+#include <silkworm/stagedsync/common.hpp>
+
+namespace silkworm::stagedysnc {
```
yperbasis, line 27: "stagedysnc" typo

node/silkworm/db/mdbx.hpp
AndreaLanfranchi: This is a comment only. It explains the rationale behind the implementation of `renew`
Also : should we know the object is no longer needed after the last commit is convenient NOT to open a new transaction for it to be aborted when the object falls out of scope.
Besides it might be necessary to call `env.close()` explicitly for example:
- to measure fsync timings (not yet implemented)
- to trap exceptions (see silkworm.cpp)

node/silkworm/common/settings.hpp
```diff
@@ -29,14 +31,19 @@
...
     std::unique_ptr<db::PruneMode> prune_mode;       // Prune mode
+    uint32_t sync_loop_throttle{0};                  // Minimum interval amongst sync cycle
```
yperbasis, line 45: What's the time unit?

node/silkworm/common/log.cpp
AndreaLanfranchi: Yeah right ... nonetheless it helped my readability.
Like tons of `else`s we have here and there
Will remove

node/silkworm/common/settings.hpp
AndreaLanfranchi: Seconds ... will amend naming

node/silkworm/stagedsync/sync_loop.hpp
AndreaLanfranchi: Solved by 9e8dd04

node/silkworm/common/log.cpp
AndreaLanfranchi: Solved by 9e8dd04

node/silkworm/common/log.hpp
AndreaLanfranchi: Solved by 9e8dd04

node/silkworm/stagedsync/sync_loop.cpp
```diff
@@ -0,0 +1,145 @@
...
+    mdbx::txn_managed external_txn;
+
+    StopWatch stop_watch;
+    Timer log_timer(
+        node_settings_->asio_context, node_settings_->sync_loop_log_interval_seconds * 1'000,
+        [&]() -> bool {
+            if (is_stopping()) {
+                return false;
+            }
+            std::string prefix = boost::str(boost::format(log_prefix_fmt) % (current_stage_ + 1) % stages_.size() %
```
yperbasis, line 50: (Minor) Would be nice to extract into a function since it's repeated at line 94.

AndreaLanfranchi: Create a function to replace a one liner call to a boost function ?
I'm puzzled

yperbasis: Yes, something like
```
std::string SyncLoop::format_prefix() const {
    static std::string log_prefix_fmt{"Stage %u/%u : %s"};
    return boost::str(boost::format(log_prefix_fmt) % (current_stage_ + 1) % stages_.size() %
                      stages_.at(current_stage_)->name());
}
```
It's a minor thing to remove code duplication between lines 50 & 94.

AndreaLanfranchi: Be my guest ... bd8a622

node/silkworm/common/log.cpp
```diff
@@ -83,6 +90,19 @@ BufferBase::BufferBase(Level level) : level_(level) {
     }
 }
 
+BufferBase::BufferBase(Level level, std::string_view msg, const std::vector<std::string>& args) : BufferBase(level) {
```
yperbasis, line 93: `vector<pair<string>>` would probably represent the intent better.

AndreaLanfranchi: Well ... if one side this ensures an even number of items to print it adds ugly braces to inline instantiaton.
Instead of
```
log::Error("Block Validation Error", {"block", std::to_string(block_num_), "err",
                                                      std::string(magic_enum::enum_name<ValidationResult>(res))});
```
we'd have to
```
log::Error("Block Validation Error", {{"block", std::to_string(block_num_)}, {"err",
                                                      std::string(magic_enum::enum_name<ValidationResult>(res))}});
```
Besides there is no strict need to have pairs as we can alwasy write
```
log::Error("Block Validation Error", {"block", std::to_string(block_num_), "err"}) << 
                                                      std::string(magic_enum::enum_name<ValidationResult>(res));
```

node/silkworm/stagedsync/stage_blockhashes.cpp
```diff
@@ -27,111 +28,131 @@
...
+    auto data{source.find(db::to_slice(header_key), /*throw_notfound=*/false)};
+    while (data.done) {
+        reached_block_num_ = {endian::load_big_u64(static_cast<uint8_t*>(data.key.iov_base))};
+        SILKWORM_ASSERT(reached_block_num_ == expected_block_number);
+        SILKWORM_ASSERT(data.value.length() == kHashLength);
+        collector_->collect(etl::Entry{Bytes(static_cast<uint8_t*>(data.value.iov_base), data.value.iov_len),
+                                       Bytes(static_cast<uint8_t*>(data.key.iov_base), data.key.iov_len)});
+        // Do we need to abort ?
+        if (!(expected_block_number % 1024) && SignalHandler::signalled()) {
+            throw std::runtime_error("Operation cancelled");
```
yperbasis, line 73: Perhaps return `kAborted` instead?

node/silkworm/stagedsync/stage_blockhashes.cpp
```diff
@@ -27,111 +28,131 @@
...
+     */
+
+    auto source{db::open_cursor(*txn, db::table::kCanonicalHashes)};
+    auto initial_key{db::block_key(to + 1)};
+    auto source_data{source.lower_bound(db::to_slice(initial_key), false)};
+
+    std::vector<Bytes> collected_keys;
+    if (source_data) {
+        db::cursor_for_each(source, [&collected_keys](::mdbx::cursor&, ::mdbx::cursor::move_result& _data) -> bool {
+            collected_keys.emplace_back(static_cast<uint8_t*>(_data.value.iov_base), _data.value.iov_len);
```
yperbasis, line 121:
```suggestion
            collected_keys.emplace_back(db::from_slice(_data.value));
```

node/silkworm/stagedsync/stagedsync.hpp
```diff
@@ -22,31 +22,94 @@
...
+    std::mutex progress_mtx_;  // Synchronizes access to progress stats
+    std::chrono::time_point<std::chrono::steady_clock> start_time_{std::chrono::steady_clock::now()};
+    std::chrono::time_point<std::chrono::steady_clock> lap_time_{std::chrono::steady_clock::now()};
+    size_t processed_blocks_{0};
+    size_t processed_transactions_{0};
+    size_t processed_mgas_{0};
+};
+
+inline constexpr size_t kDefaultBatchSize = 512_Mebi;          // TODO(Andrea) Replace with value from CLI
 inline constexpr size_t kDefaultRecoverySenderBatch = 50'000;  // This a number of transactions not number of bytes
```
yperbasis, line 95: `kDefaultRecoverySenderBatch` is unused now.

node/silkworm/stagedsync/stagedsync.hpp
```diff
@@ -22,31 +22,94 @@
...
+    StageResult forward(db::RWTxn& txn) final;
+    StageResult unwind(db::RWTxn& txn, BlockNum to) final;
+    StageResult prune(db::RWTxn& txn) final;
+    std::vector<std::string> get_log_progress() final;
+
+  private:
+    std::unique_ptr<recovery::RecoveryFarm> farm_{nullptr};
+};
+
+class Execution final : public IStage {
```
yperbasis, line 67: I'd suggest to use names like `ExecutionStage`, `SendersStage`, etc for clarity.

yperbasis: And I'd move this class declaration to `stage_execution.hpp` to follow the traditional pattern of `stage_execution.cpp` implementing methods from the corresponding header.

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -31,120 +30,174 @@
...
+                     << " while Bodies stage " << bodies_stage_progress;
+        return StageResult::kInvalidProgress;
+    } else if (previous_progress > senders_stage_progress) {
+        // Same as above but for senders
+        log::Error() << "Bad progress sequence. Execution stage progress " << previous_progress
+                     << " while Senders stage " << senders_stage_progress;
+        return StageResult::kInvalidProgress;
+    }
+
+    std::unique_lock progress_lock(progress_mtx_);
```
yperbasis, line 62: Why do we need a mutex? Execution is single-threaded, isn't it?

yperbasis (commented):
LGTM to me overall. Thanks a lot for this huge work to bring Silkworm closer to being a product!

Let's put some finishing touches, restore pre-verified headers, and merge.

node/silkworm/stagedsync/stage_execution.cpp
AndreaLanfranchi: The execution is single threaded but the call to `get_log_progress` comes from an async boost' timer which runs in the ASIO thread

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -31,120 +30,169 @@
...
     }
-
-    return res;
+    auto speed_blocks = processed_blocks_ / elapsed_seconds;
+    auto speed_transactions = processed_transactions_ / elapsed_seconds;
+    auto speed_mgas = processed_mgas_ / elapsed_seconds;
+    processed_blocks_ = 0;
+    processed_transactions_ = 0;
+    processed_mgas_ = 0;
+    progress_lock.unlock();
```
yperbasis, line 192: `progress_lock.unlock()` is unnecessary

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: I'd then put a comment into `IStage::get_log_progress` saying that implementations must be thread-safe.

yperbasis: Or, more precisely, that `get_log_progress` might be called not from the main staged sync thread.

yperbasis (approved):
Looks good to me overall. Though I would move classes like `stagedsync::Execution` to their own headers like `stage_execution.hpp` to follow the traditional pattern of `stage_execution.cpp` implementing methods from the corresponding header.

node/silkworm/common/stopwatch.hpp
```diff
@@ -28,59 +28,42 @@
...
 
 namespace silkworm {
-/// <summary>
-/// This class mimics the behavior of a stopwatch to measure timings of operations
-/// </summary>
+//! \brief This class mimics the behavior of a stopwatch to measure timings of operations
 class StopWatch {
   public:
     typedef std::chrono::time_point<std::chrono::high_resolution_clock> TimePoint;
     typedef std::chrono::nanoseconds Duration;
```
greg7mdp, line 35: Not related to this change... just wanted to suggest using `using` instead of `typedef`. The benefit is that the new type is right at the beginning and you don't have to hunt for it.

node/silkworm/stagedsync/stage_execution.cpp
AndreaLanfranchi: Well ... actually it is : it releases the lock **before** the function continues in building a vector with 4 to string conversions.
Yes is a matter of nanoseconds but is clear that the developer releases the lock ASAP

yperbasis: fair enough

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -31,120 +30,169 @@
...
+        {
+            std::unique_lock progress_lock(progress_mtx_);
+            lap_time_ = std::chrono::steady_clock::now();
         }
+
         while (true) {
-            std::optional<BlockWithHash> bh{db::read_block(txn, block_num, /*read_senders=*/true)};
-            if (!bh.has_value()) {
-                return StageResult::kBadChainSequence;
+            if (!(block_num_ % 64) && SignalHandler::signalled()) {
```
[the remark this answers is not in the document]
AndreaLanfranchi: Ok

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -31,120 +30,169 @@
...
-            if (block_num % 1000 == 0) {
-                log::Debug() << "Blocks <= " << block_num << " executed";
-            }
-
-            if (buffer.current_batch_size() >= batch_size || block_num >= max_block) {
+            buffer.insert_receipts(block_num_, receipts);
+            std::unique_lock progress_lock(progress_mtx_);
+            processed_blocks_++;
+            processed_transactions_ += block_with_hash->block.transactions.size();
+            processed_mgas_ += block_with_hash->block.header.gas_used / 1'000'000;
```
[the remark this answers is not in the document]
AndreaLanfranchi: I honestly find that showing, for example, `blocks/s=4.53` is a bit ugly and does not depict correctly what is going on. (Half a block might be 6 transactions or 90). I preferred to stay conservative and show only integers.
Besides properly formatting a float (with non variadic number of digits) requires printf or a lot of ioman helpers on a stringstream, or again a boost format.
If not something critical I'd leave as is ... nevertheless the rate is so variable that this should only indicate a gross value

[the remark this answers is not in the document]
yperbasis: It's not critical, but I'd still suggest the following: change processed_mgas_ to processed_gas_ and divide it by 1M only at the end. Thus you'll improve the accuracy without introducing double. Compare the accuracy of
```
700/1000 + 1500/1000 + 800/1000
```
with
```
(700 + 1500 + 800) / 1000
```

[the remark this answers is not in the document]
yperbasis: Also, if we're talking nanoseconds, division is a slow operation :)

node/silkworm/stagedsync/stage_execution.cpp
AndreaLanfranchi: Right.

node/silkworm/stagedsync/stage_execution.cpp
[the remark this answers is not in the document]
AndreaLanfranchi: Correct about the precision ... about divisions ... when the compiler detects a division by a constant _it should be smart enough_ to convert the division into a multiplication. And the division by 1M is constant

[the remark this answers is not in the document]
AndreaLanfranchi: [No division if known divisor at compile time](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:31,endLineNumber:9,positionColumn:31,positionLineNumber:9,selectionStartColumn:31,selectionStartLineNumber:9,startColumn:31,startLineNumber:9),source:'%23include+%3Ccstddef%3E%0A%23include+%3Ccstdint%3E%0A%0Auint64_t+divbyconst(uint64_t+arg)+%7B%0A%09return+arg+/+1000000%3B%0A%7D%0A%0Auint64_t+divbynotconst(uint64_t+dividend,+uint64_t+divisor)+%7B%0A++++return+dividend+/+divisor%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:51.680539266943384,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:g112,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+11.2+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),k:100,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+11.2+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:48.319460733056616,l:'3',n:'0',o:'',t:'0')),l:'2',m:100,n:'0',o:'',t:'0')),version:4)

node/silkworm/common/log.cpp
```diff
@@ -68,6 +71,10 @@ static inline std::pair<const char*, const char*> get_channel_settings(Level lev
 }
 
 BufferBase::BufferBase(Level level) : level_(level) {
+    if (level > settings_.log_verbosity) {
+        return;
+    }
```
canepat, line 76: This is an improvement for sure. I think the same early return should be added also into the `append` method to avoid unnecessary work, i.e. appending to the `std::stringstream` instance.

AndreaLanfranchi: Will amend ... nevertheless I wish we could focus a bit more on overall design of this PR instead of searching performance bounds for logging which, btw, is really not a critical part (we don't log hundreds times per second).

node/silkworm/stagedsync/stage_execution.cpp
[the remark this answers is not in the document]
yperbasis: Interesting, I didn't know that.

node/silkworm/common/stopwatch.hpp
AndreaLanfranchi: Thanks. Done

node/silkworm/common/log.cpp
AndreaLanfranchi: Done

PR #538, opened 2022-01-13
mdbx::slice::data() instead of iov_base

yperbasis:
`iov_base` looks like an implementation detail we shouldn't rely upon

### DIFF
--- a/cmd/check_blockhashes.cpp
+++ b/cmd/check_blockhashes.cpp
@@ -60,13 +60,14 @@ int main(int argc, char* argv[]) {
             ByteView hash_data_view{db::from_slice(canonical_hashes_data.value)};  // Canonical Hash
             auto block_hashes_data{blockhashes_table.find(canonical_hashes_data.value, /*throw_notfound*/ false)};
             if (!block_hashes_data) {
-                uint64_t hash_block_number{endian::load_big_u64(canonical_hashes_data.key.byte_ptr())};
+                uint64_t hash_block_number{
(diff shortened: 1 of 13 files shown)


AndreaLanfranchi:
Why this ?
I understand the removal of `byte_ptr` but why remove `iov_base/iov_len` if both `data()` and `length()` return the very same values ?

yperbasis:
Semantically `mdbx::slice` is very close to `std::string_view`. `std::string_view`,  `std::string`, and `std::vector` all have the `data()` member function, while `iov_base` looks more niche, and `MDBX_val` being `iovec` seems like an implementation detail to me.

AndreaLanfranchi:
Well ... the difference is `mdbx::slice::data()` returns a `void*` to data (by iovec) while the others you mentioned return a `T*` to data where T is the undelying type. Not much difference as long as we static cast always to bytes but `iovec` is more "generic" and depicts a raw memory ptr.

yperbasis:
It's about names and the API. In the C++ world `data()` is more commonly used and widely understood than `iov_base` (at least that's my impression). When a new reader sees `data()` in our code, he is more likely to get the meaning compared to him seeing `iov_base`.

AndreaLanfranchi:
Ok for me then.

AndreaLanfranchi (approved):

greg7mdp:
I like this change as well. `iov_base` looks like accessing an internal variable, while `data()` resembles a proper api.

PR #543, opened 2022-01-18
[Optimization] Reduce memory allocation in execute_transaction

yperbasis:
This is a small opimization reducing memory allocation in `ExecutionProcessor::execute_transaction`. The idea is use logs (which might have some capacity already) from the provided out parameter `receipt` in `IntraBlockState`. The total time of executing 14M main net blocks is reduced from 33.8h to 33.4h.

### DIFF
--- a/core/silkworm/execution/processor.cpp
+++ b/core/silkworm/execution/processor.cpp
@@ -16,7 +16,6 @@
 
 #include "processor.hpp"
 
-#include <algorithm>
 #include <cassert>
 
 #include <silkworm/chain/dao.hpp>
--- a/core/silkworm/execution/processor.cpp
+++ b/core/silkworm/execution/processor.cpp
@@ -149,8 +149,8 @@ uint64_t ExecutionProcessor::refund_gas(const Transaction& txn, uint64_t gas_lef
 }
 
 ValidationResult ExecutionProcessor::execute_block_no_post_validation(std::vector<Receipt>& receipts) noexcept {
-    receipts.clear();
-    receipts.reserve(evm_.block().transactions.size());
+    const size_t n{evm_.block().transactions.size()};
+    receipts.resize(n);
 
     uint64_t block_num{evm_.block().header.number};
     if (block_num == evm_.config().dao_block) {
@@ -158,12 +158,13 @@ ValidationResult ExecutionProcessor::execute_block_no_post_validation(std::vecto
     }
 
     cumulative_gas_used_ = 0;
-    for (const Transaction& txn : evm_.block().transactions) {
+    for (size_t i{0}; i < n; ++i) {
+        const Transaction& txn{evm_.block().transactions[i]};
         const ValidationResult err{validate_transaction(txn)};
         if (err != ValidationResult::kOk) {
             return err;
         }
-        receipts.push_back(execute_transaction(txn));
+        receipts[i] = execute_transaction(txn);
     }
 
     consensus_engine_.finalize(state_, evm_.block(), evm_.revision());
--- a/core/silkworm/execution/processor.cpp
+++ b/core/silkworm/execution/processor.cpp
@@ -149,17 +149,18 @@ uint64_t ExecutionProcessor::refund_gas(const Transaction& txn, uint64_t gas_lef
 }
 
 ValidationResult ExecutionProcessor::execute_block_no_post_validation(std::vector<Receipt>& receipts) noexcept {
-    const size_t n{evm_.block().transactions.size()};
-    receipts.resize(n);
+    const Block& block{evm_.block()};
 
-    uint64_t block_num{evm_.block().header.number};
-    if (block_num == evm_.config().dao_block) {
+    if (block.header.number == evm_.config().dao_block) {
         dao::transfer_balances(state_);
     }
 
     cumulative_gas_used_ = 0;
+
+    const size_t n{block.transactions.size()};
+    receipts.resize(n);
     for (size_t i{0}; i < n; ++i) {
-        const Transaction& txn{evm_.block().transactions[i]};
+        const Transaction& txn{block.transactions[i]};
         const ValidationResult err{validate_transaction(txn)};
         if (err != ValidationResult::kOk) {
             return err;
@@ -167,7 +168,7 @@ ValidationResult ExecutionProcessor::execute_block_no_post_validation(std::vecto
         receipts[i] = execute_transaction(txn);
     }
 
-    consensus_engine_.finalize(state_, evm_.block(), evm_.revision());
+    consensus_engine_.finalize(state_, block, evm_.revision());
 
     return ValidationResult::kOk;
 }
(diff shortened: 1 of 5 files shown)


core/silkworm/execution/processor.cpp
```diff
@@ -149,21 +149,22 @@ uint64_t ExecutionProcessor::refund_gas(const Transaction& txn, uint64_t gas_lef
...
-    for (const Transaction& txn : evm_.block().transactions) {
+    for (size_t i{0}; i < n; ++i) {
+        const Transaction& txn{evm_.block().transactions[i]};
         const ValidationResult err{validate_transaction(txn)};
         if (err != ValidationResult::kOk) {
             return err;
         }
-        receipts.push_back(execute_transaction(txn));
+        receipts[i] = execute_transaction(txn);
     }
```
AndreaLanfranchi, line 168: Why not
```suggestion
    receipts.resize(evm_.block().transactions.size());
    auto receipts_it{receipts.begin()};
    
    uint64_t block_num{evm_.block().header.number};
    if (block_num == evm_.config().dao_block) {
        dao::transfer_balances(state_);
    }
    
    cumulative_gas_used_ = 0;
    for (const auto& txn : evm_.block().transactions) {
        if (auto err{validate_transaction(txn)} ; err != ValidationResult::kOk) {
            return err;
        }
        *receipts_it = execute_transaction(txn);
        ++receipts_it;
    }
```

yperbasis: Well, I find the index version simpler and more readable than the iterator version, but that's a matter of taste.

AndreaLanfranchi: Imho is the opposite ... however is only a suggestion

core/silkworm/execution/processor.cpp
```diff
@@ -146,24 +148,26 @@ uint64_t ExecutionProcessor::refund_gas(const Transaction& txn, uint64_t gas_lef
...
+
+    receipts.resize(block.transactions.size());
+    auto receipt_it{receipts.begin()};
+    for (const auto& txn : block.transactions) {
         const ValidationResult err{validate_transaction(txn)};
         if (err != ValidationResult::kOk) {
             return err;
         }
-        receipts.push_back(execute_transaction(txn));
+        execute_transaction(txn, *receipt_it);
```
AndreaLanfranchi, line 165: Shouldn't we check, after return, for `txn.success == true` ?

core/silkworm/execution/processor.cpp
```diff
@@ -113,13 +117,11 @@ Receipt ExecutionProcessor::execute_transaction(const Transaction& txn) noexcept
...
 
-    return {
-        txn.type,                       // type
-        vm_res.status == EVMC_SUCCESS,  // success
-        cumulative_gas_used_,           // cumulative_gas_used
-        logs_bloom(state_.logs()),      // bloom
-        state_.logs(),                  // logs
-    };
+    receipt.type = txn.type;
+    receipt.success = vm_res.status == EVMC_SUCCESS;
```
AndreaLanfranchi, line 120: Any chance `vm_res.status != EVMC_SUCCESS` ?
If so ... can we exit the function early (right after `evm_.execute`) ?

core/silkworm/execution/processor.cpp
```diff
@@ -66,9 +67,12 @@ ValidationResult ExecutionProcessor::validate_transaction(const Transaction& txn
...
-Receipt ExecutionProcessor::execute_transaction(const Transaction& txn) noexcept {
+void ExecutionProcessor::execute_transaction(const Transaction& txn, Receipt& receipt) noexcept {
     assert(validate_transaction(txn) == ValidationResult::kOk);
 
+    // Optimization: since receipt.logs might have some capacity, let's reuse it.
+    std::swap(receipt.logs, state_.logs());
+
     state_.clear_journal_and_substate();
 
     assert(txn.from.has_value());
```
AndreaLanfranchi, line 77: Maybe this should be a `SILKWORM_ASSERT` or a std::runtime error if no value

core/silkworm/execution/processor.cpp
```diff
@@ -66,9 +67,12 @@ ValidationResult ExecutionProcessor::validate_transaction(const Transaction& txn
     return ValidationResult::kOk;
 }
 
-Receipt ExecutionProcessor::execute_transaction(const Transaction& txn) noexcept {
+void ExecutionProcessor::execute_transaction(const Transaction& txn, Receipt& receipt) noexcept {
     assert(validate_transaction(txn) == ValidationResult::kOk);
```
AndreaLanfranchi, line 70: Assert vould be voided in build release ... SILKWORM_ASSERT or runtime error ?

core/silkworm/execution/processor.cpp
yperbasis: Nope: blocks with unsuccessful transactions are still valid. E.g. https://etherscan.io/tx/0xeb1272b31fb9ef4a876c3a386748d0b3fcf76d7a3aac280336fdbecffa027f99

core/silkworm/execution/processor.cpp
yperbasis: Ditto: blocks with unsuccessful transactions are still valid.

core/silkworm/execution/processor.cpp
```diff
@@ -66,9 +66,12 @@ ValidationResult ExecutionProcessor::validate_transaction(const Transaction& txn
...
 
+    // Optimization: since receipt.logs might have some capacity, let's reuse it.
+    std::swap(receipt.logs, state_.logs());
```
AndreaLanfranchi, line 73: `std::swap` is defined in vector so maybe semnatically is better `receipt.logs.swap(state_.logs());`

core/silkworm/execution/processor.cpp
yperbasis: `validate_transaction` is already called by `execute_block_no_post_validation` prior to `execute_transaction`. `execute_transaction` on its own is called only by tests; execution stage always goes through `execute_block_no_post_validation`. I don't want to use `SILKWORM_ASSERT` here since that would mean calling `validate_transaction` twice in production.

core/silkworm/execution/processor.cpp
AndreaLanfranchi: Ok

core/silkworm/execution/processor.cpp
AndreaLanfranchi: Ok

core/silkworm/execution/processor.cpp
AndreaLanfranchi: Ok

core/silkworm/execution/processor.cpp
yperbasis: I prefer `std::swap(a, b)` to highlight the symmetry of the operation, but it's a minor aesthetical point.

AndreaLanfranchi: Fair enough

core/silkworm/execution/processor.cpp
yperbasis: `validate_transaction` already checks that `txn.from` is not empty, so here assert is more like documentation rather than a runtime check.

AndreaLanfranchi: Ok

AndreaLanfranchi (approved):

PR #545, opened 2022-01-21
Fix downloader issues

mriccobene:


node/silkworm/downloader/internals/chain_elements.hpp
```diff
@@ -78,7 +78,7 @@ struct Link {
 
     void remove_child(std::shared_ptr<Link> child) {
```
greg7mdp, line 79: It might not be an issue here, but it is not recommended to take a std::shared_ptr as a value, because every call causes a reference increment/decrement. Better take a `const std::shared_ptr<Link>&`, or even a `const Link&`

yperbasis: Yes, I agree. `child` doesn't leak so there's not need for `shared_ptr` here, `const &` should be just fine.

mriccobene: ok, I will do it in the next commit

canepat (changes requested):

node/silkworm/downloader/internals/preverified_hashes.hpp
```diff
@@ -20,8 +20,8 @@
 #include <map>
 #include <set>
 
-#include <silkworm/chain/identity.hpp>
-#include <silkworm/common/base.hpp>
+#include "silkworm/chain/identity.hpp"
+#include "silkworm/common/base.hpp"
```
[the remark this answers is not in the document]
mriccobene: done

node/silkworm/downloader/internals/preverified_hashes_mainnet.cpp
```diff
@@ -16,7 +16,7 @@
 
 #include "silkworm/common/base.hpp"
 
-#include "preverified_hashes.hpp"
+#include "silkworm/downloader/internals/preverified_hashes.hpp"
```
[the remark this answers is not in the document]
mriccobene: done

node/silkworm/downloader/internals/working_chain.cpp
```diff
@@ -55,17 +59,134 @@ bool WorkingChain::in_sync() const {
...
+           std::to_string(persisted_link_queue_.size()) + " persisting/ed" + (invariant ? "), " : " !),") +
+           std::to_string(anchors_.size()) + "/" + std::to_string(anchor_queue_.size()) + " anchors, " +
+           std::to_string(highest_in_db_) + " highest block in db";
+
+    //auto elapsed_minutes = duration_cast<minutes>(system_clock::now() - last_execution).count();
+    //if (elapsed_minutes >= 5) {
+    //    output += "\n";
+    //    output += dump_chain_bundles();
+    //    last_execution = system_clock::now();
+    //}
```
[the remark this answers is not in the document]
mriccobene: removed

node/silkworm/downloader/internals/working_chain.cpp
```diff
@@ -55,17 +59,134 @@ bool WorkingChain::in_sync() const {
...
+
+    return output;
+}
+
+void WorkingChain::assure_no_orphaned_links() const {
+    using std::to_string;
```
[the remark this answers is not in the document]
mriccobene: code removed

node/silkworm/downloader/internals/working_chain.cpp
```diff
@@ -55,17 +59,134 @@ bool WorkingChain::in_sync() const {
...
+    if (!orphaned_links.empty())
+        log::Trace() << dump_orphaned_links(orphaned_links);
+
+    assert(orphaned_links.empty());
+}
+
+std::set<BlockNum> WorkingChain::search_orphaned_links() const {
+    std::set<BlockNum> orphaned_links;
+
+    for (auto& l: links_) {
```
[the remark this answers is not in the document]
mriccobene: code removed

node/silkworm/downloader/internals/working_chain.cpp
```diff
@@ -55,17 +59,134 @@ bool WorkingChain::in_sync() const {
...
+        auto link = l.second;
```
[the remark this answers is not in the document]
mriccobene: code removed

node/silkworm/downloader/internals/chain_elements.hpp
mriccobene: done

node/silkworm/downloader/internals/chain_elements.hpp
```diff
@@ -76,9 +76,9 @@ struct Link {
         persisted = persisted_;
     }
 
-    void remove_child(std::shared_ptr<Link> child) {
+    void remove_child(const std::shared_ptr<Link>& child) {
```
yperbasis, line 79: There's no need for `shared_ptr` here at all because `child` doesn't outlive this function, does it? So no need to take and then release ownership.
The signature can be simplified to
```
void remove_child(const Link& child)
```

mriccobene: Sure, you are right: there is no need to take and then release ownership. The link outlive this function because the caller manipulate shared_ptr. All the code around use shared_ptr because links are created and inserted in some queues, passed around and so on.
So now the caller must do:
`anchor.remove_child(link_ptr); `
while with your suggestion it must do:
`anchor.remove_child(*(link_ptr)); `
What do you think is it better?

yperbasis: I think that it's better (faster, simpler) to pass child by `const &`  to `remove_child` (and to other methods that don't require ownership). The call site will, correspondingly, be `anchor.remove_child(*link_ptr)` if `link_ptr` is a `shared_ptr`.

mriccobene: done

node/silkworm/downloader/header_downloader.cpp
```diff
@@ -54,22 +54,26 @@ void HeaderDownloader::execution_loop() {
...
+            log::Trace() << "HeaderDownloader processing message " << *in_message;
+        }
 
         // process the message (command pattern)
         message->execute();
 
         auto out_message = std::dynamic_pointer_cast<OutboundMessage>(message);
         if (out_message) {
-            log::Info() << "HeaderDownloader sent message " << *out_message;
+            log::Trace() << "HeaderDownloader sent message " << *out_message;
```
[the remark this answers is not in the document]
mriccobene: done

cmd/download_headers.cpp
```diff
@@ -50,7 +50,10 @@ int main(int argc, char* argv[]) {
 
     CLI11_PARSE(app, argc, argv);
 
-    log::set_verbosity(log::Level::kTrace);
+    log::Settings settings;
+    settings.log_threads = true;
+    log::init(settings);
+    log::set_verbosity(log::Level::kInfo);
```
AndreaLanfranchi, line 56: This line is not needed. log::Settings ctor already defaults the verbosity to kInfo

cmd/download_headers.cpp
```diff
@@ -50,7 +50,10 @@ int main(int argc, char* argv[]) {
...
     log::tee_file(std::filesystem::path("downloader.log"));
```
AndreaLanfranchi, line 57: I'd suggest to to set the file name into settings directly and let `log::init(settings)` do the rest.
```
log::Settings settings;
settings.log_file = "downloader.log";
log::init(settings);
```

core/silkworm/common/lru_cache.hpp
```diff
@@ -80,6 +81,12 @@ class lru_cache {
         }
     }
 
+    std::optional<value_t> get_a_copy(const key_t& key) {
+        auto val = get(key);
+        if (val == nullptr) return std::nullopt;
```
AndreaLanfranchi, line 86: Let's avoid 1 line ifs

cmd/download_headers.cpp
AndreaLanfranchi: BTW ... you don't set a CLI argument to set verbosity level : as a result, being defaulted to Info, all traces are not printed.

core/silkworm/common/lru_cache.hpp
```diff
@@ -80,6 +81,12 @@ class lru_cache {
         }
     }
 
+    std::optional<value_t> get_a_copy(const key_t& key) {
```
AndreaLanfranchi, line 84: Probably better `get_as_copy` instead of `get_a_copy`

AndreaLanfranchi:
In `thread_safe_queue` I find a bit of a nonsense in here 
```
    template <typename Duration>
    bool timed_wait_and_pop(T& popped_value, Duration const& wait_duration) {
        std::unique_lock lock(mutex_);
        if (!condition_variable_.wait_for(lock, wait_duration, [this] { return !queue_.empty(); })) {
            return false;
        }
        popped_value = queue_.front();
        queue_.pop_front();
        return true;
    }
```
As this is a "timed wait" I interpret it as "wait until some other thread pushes something in the queue". But this cannot happen as the very moment we enter here we lock the mutex hence any `push()` cannot happen as we're under the same mutex. Therefore the stop predicate in the function does not make any sense to me: basically if we enter here and the queue is empty the only possibility to exit is either the queue is not empty, or the cv to timeout. It cannot happen that something gets into the queue while waiting for the cv.

I'd have created a second mutex for the cv awaiter and rewrite as 
```
    template <typename Duration>
    bool timed_wait_and_pop(T& popped_value, Duration const& wait_duration) {
        while(!try_pop(popped_value)) {
            std::unique_lock lock(cvmutex_);
            if(condition_variable_.template wait_for(lock, wait_duration) == std::cv_status::timeout) {
                return false;
            }
        }
        return true;
    }
```
This allows other threads to push items and the cv to notify. We also save a lambda which capture the whole `this`

AndreaLanfranchi:
In `HeaderDownloader::execution_loop()` why the early continue ?
```
    while (!is_stopping() && !sentry_.is_stopping()) {
        // pop a message from the queue
        std::shared_ptr<Message> message;
        bool present = messages_.timed_wait_and_pop(message, 1000ms);
        if (!present) continue;  // timeout, needed to check exiting_

        // process the message (command pattern)
        message->execute();
    }
```
I find this simpler and immediate to understand
```
    while (!is_stopping() && !sentry_.is_stopping()) {
        std::shared_ptr<Message> message;
        if(messages_.timed_wait_and_pop(message, 1s)) {
            message->execute(); // Process the message from the queue (if any)
        }
    }
```

canepat:
> In `thread_safe_queue` I find a bit of a nonsense in here
> 
> ```
>     template <typename Duration>
>     bool timed_wait_and_pop(T& popped_value, Duration const& wait_duration) {
>         std::unique_lock lock(mutex_);
>         if (!condition_variable_.wait_for(lock, wait_duration, [this] { return !queue_.empty(); })) {
>             return false;
>         }
>         popped_value = queue_.front();
>         queue_.pop_front();
>         return true;
>     }
> ```
> 
> As this is a "timed wait" I interpret it as "wait until some other thread pushes something in the queue". But this cannot happen as the very moment we enter here we lock the mutex hence any `push()` cannot happen as we're under the same mutex. 

AFAIK the code seems correct as it is because the `std::condition_variable::wait_for` method performs three actions _atomically_: releases lock, blocks the current executing thread and adds it to the the waiting threads. When the current thread is unblocked (either because of a notify or timeout expired), it will reacquire the lock before exiting the call

mriccobene:
> In `thread_safe_queue` I find a bit of a nonsense in here
> 
> ```
>     template <typename Duration>
>     bool timed_wait_and_pop(T& popped_value, Duration const& wait_duration) {
>         std::unique_lock lock(mutex_);
>         if (!condition_variable_.wait_for(lock, wait_duration, [this] { return !queue_.empty(); })) {
>             return false;
>         }
>         popped_value = queue_.front();
>         queue_.pop_front();
>         return true;
>     }
> ```
> 
> As this is a "timed wait" I interpret it as "wait until some other thread pushes something in the queue". But this cannot happen as the very moment we enter here we lock the mutex hence any `push()` cannot happen as we're under the same mutex. Therefore the stop predicate in the function does not make any sense to me: basically if we enter here and the queue is empty the only possibility to exit is either the queue is not empty, or the cv to timeout. It cannot happen that something gets into the queue while waiting for the cv.

I think you are wrong, it is the work of the condition_variable to unlock and permit push.

AndreaLanfranchi:
You're right sorry ... forgot the unlock by the cv **and** the fact waiting with predicate returns the result of the predicate, not of the waiting. My fault : big mistake.

cmd/download_headers.cpp
mriccobene: done

cmd/download_headers.cpp
mriccobene: I just introduced CLI argument for log verbosity level, as you suggested

core/silkworm/common/lru_cache.hpp
mriccobene: done

core/silkworm/common/lru_cache.hpp
mriccobene: done

canepat (approved):

PR #547, opened 2022-01-23
Update evmone

yperbasis:
Also use `intx::bswap` instead of `SILKWORM_BSWAP` macros.


chfast:
If you want to check performance, you should include https://github.com/ethereum/evmone/pull/418. The https://github.com/ethereum/evmone/pull/416 makes Baseline slower.

yperbasis:
> If you want to check performance, you should include [ethereum/evmone#418](https://github.com/ethereum/evmone/pull/418). The [ethereum/evmone#416](https://github.com/ethereum/evmone/pull/416) makes Baseline slower.

Right, I've checked Silkworm execution performance with evmone at https://github.com/ethereum/evmone/commit/2f17da7a76e96931f67558843e0bbb150f53ddb0. The evmone update improved the performance only very slightly, from 33h:23m to 33h:14m.

chfast:
> Right, I've checked Silkworm execution performance with evmone at [ethereum/evmone@2f17da7](https://github.com/ethereum/evmone/commit/2f17da7a76e96931f67558843e0bbb150f53ddb0). The evmone update improved the performance only very slightly, from 33h:23m to 33h:14m.

Keep in mind that only Baseline is being optimized. So if you are using mostly Advanced you may not see big difference. Even regressions would not be a surprise.

yperbasis:
> Keep in mind that only Baseline is being optimized. So if you are using mostly Advanced you may not see big difference. Even regressions would not be a surprise.

I see. Anyway, I'll check Silkworm performance with https://github.com/ethereum/evmone/pull/418 as well in the next few days.

core/silkworm/common/endian.hpp
```diff
@@ -80,9 +67,9 @@ namespace le {
     }
 }  // namespace le
 namespace be {
-    static inline uint16_t load(uint16_t value) noexcept { return SILKWORM_BSWAP16(value); }
-    static inline uint32_t load(uint32_t value) noexcept { return SILKWORM_BSWAP32(value); }
-    static inline uint64_t load(uint64_t value) noexcept { return SILKWORM_BSWAP64(value); }
+    static inline uint16_t load(uint16_t value) noexcept { return intx::bswap(value); }
+    static inline uint32_t load(uint32_t value) noexcept { return intx::bswap(value); }
+    static inline uint64_t load(uint64_t value) noexcept { return intx::bswap(value); }
```
AndreaLanfranchi, line 72: Does intx already implement instrinsics ?

yperbasis: Yes, it does: https://github.com/chfast/intx/blob/v0.7.1/include/intx/intx.hpp#L547

AndreaLanfranchi (approved):

chfast (commented):
1. What is the stack space situation?
2. Do you use Baseline anywhere?

yperbasis:
> 1. What is the stack space situation?
> 2. Do you use Baseline anywhere?

2. Baseline is used in consensus tests and also when executing initial deployment code. Advanced is used when executing already deployed smart contracts.
1. With this evmone update I had to bump the stack size in consensus tests – otherwise linux-clang-10-address-sanitizer was failing with stack overflow.

PR #549, opened 2022-01-27
[Optimization] Reuse cursors to reduce memory reallocation

yperbasis:
With this optimization main net execution time goes down from 33h:26m to 28h:56m.

### DIFF
--- a/node/silkworm/db/access_layer.cpp
+++ b/node/silkworm/db/access_layer.cpp
@@ -373,9 +373,10 @@ evmc::bytes32 read_storage(mdbx::txn& txn, const evmc::address& address, uint64_
                                     ? historical_storage(txn, address, incarnation, location, block_num.value())
                                     : std::nullopt};
     if (!val.has_value()) {
-        auto src{db::open_cursor(txn, table::kPlainState)};
+        thread_local auto state_cursor{db::open_cursor(txn, table::kPlainState)};
+        state_cursor.renew(txn);
         auto key{storage_prefix(address, incarnation)};
-        val = find_value_suffix(src, key, location);
+        val = find_value_suffix(state_cursor, key, location);
     }
 
     if (!val.has_value()) {
--- a/node/silkworm/db/access_layer.cpp
+++ b/node/silkworm/db/access_layer.cpp
@@ -25,7 +25,8 @@
 namespace silkworm::db {
 
 std::optional<VersionBase> read_schema_version(mdbx::txn& txn) noexcept {
-    auto src{db::open_cursor(txn, table::kDatabaseInfo)};
+    thread_local mdbx::cursor_managed src;
+    src.bind(txn, db::open_map(txn, db::table::kDatabaseInfo));
     if (!src.seek(mdbx::slice{kDbSchemaVersionKey})) {
         return std::nullopt;
     }
@@ -55,13 +56,14 @@ void write_schema_version(mdbx::txn& txn, const VersionBase& schema_version) {
     endian::store_big_u32(&value[0], schema_version.Major);
     endian::store_big_u32(&value[4], schema_version.Minor);
     endian::store_big_u32(&value[8], schema_version.Patch);
-    auto src{db::open_cursor(txn, table::kDatabaseInfo)};
+    thread_local mdbx::cursor_managed src;
+    src.bind(txn, db::open_map(txn, db::table::kDatabaseInfo));
     src.upsert(mdbx::slice{kDbSchemaVersionKey}, to_slice(value));
 }
 
 std::optional<BlockHeader> read_header(mdbx::txn& txn, BlockNum block_number, const uint8_t (&hash)[kHashLength]) {
-    thread_local auto src{db::open_cursor(txn, table::kHeaders)};
-    src.renew(txn);
+    thread_local mdbx::cursor_managed src;
+    src.bind(txn, db::open_map(txn, db::table::kHeaders));
     auto key{block_key(block_number, hash)};
     auto data{src.find(to_slice(key), false)};
     if (!data) {
@@ -79,8 +81,8 @@ void write_header(mdbx::txn& txn, const BlockHeader& header, bool with_header_nu
     rlp::encode(value, header);
     auto header_hash{header.hash()};
     auto key{db::block_key(header.number, header_hash.bytes)};
-    thread_local auto target{db::open_cursor(txn, table::kHeaders)};
-    target.renew(txn);
+    thread_local mdbx::cursor_managed target;
+    target.bind(txn, db::open_map(txn, table::kHeaders));
     target.upsert(to_slice(key), to_slice(value));
     if (with_header_numbers) {
         write_header_number(txn, header_hash.bytes, header.number);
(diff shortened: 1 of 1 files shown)


AndreaLanfranchi (commented):
I am very skeptical about this optimization.
Basically it saves the reallocation (and deallocation) of `MDBX_cursor_couple` (which is good) but it exposes to a risk of reading wrong data if I read MDBX's implementation correctly.

Actually `renew` tries to bind the same `dbi` of the cursor into the privded transaction but dbis are mapped to bucket names at `env` level on behalf of the order they were firstly accessed

```

        {
            auto chaindata_env{silkworm::db::open_env(node_settings.chaindata_env_config)};
            auto txn{chaindata_env.start_read()};
            auto cursor_ps{db::open_cursor(txn, db::table::kPlainState)};
            cursor_ps = db::open_cursor(txn,db::table::kPlainState);
            std::cout << cursor_ps.map().dbi << std::endl;
        }

        {
            auto chaindata_env{silkworm::db::open_env(node_settings.chaindata_env_config)};
            auto txn{chaindata_env.start_read()};
            auto cursor1{db::open_cursor(txn, db::table::kPlainContractHash)};
            auto cursor2{db::open_cursor(txn, db::table::kHashedAccounts)};
            auto cursor3{db::open_cursor(txn, db::table::kCallTraceSet)};
            auto cursor_ps{db::open_cursor(txn, db::table::kPlainState)};
            std::cout << cursor_ps.map().dbi << std::endl;
        }
```

Produces prints 2 and 5 respectively.

So ... as here we don't know which env the transaction is bound to (nor if the env has been previously closed and reopened) we definitely break code isolation and we risk to renew the binding to a dbi which has nothing to do with original named bucket

I find more reasonable instead to provide a proper set of overloads to read_header from cursor (instead of from txn) and keep cursor opened (e.g. in buffer) at higher level code.

greg7mdp:
does `thread_local` make any difference here? My understanding is that it is useful only for static variables, as local variables are stored in the stack which is already thread specific. https://en.cppreference.com/w/cpp/language/storage_duration

yperbasis:
> does `thread_local` make any difference here? My understanding is that it is useful only for static variables, as local variables are stored in the stack which is already thread specific. https://en.cppreference.com/w/cpp/language/storage_duration

`thread_local` here means thread storage duration. That is similar to static storage duration, but with the instances being independent between threads, which is what we want (`static` would mean an instance shared between threads and potential data races).

greg7mdp:
> thread_local here means thread storage duration. That is similar to static storage duration

I don't understand the benefit of using thread_local for these variables? Do we want the initialization to occur only once? This doesn't seem correct to me.

AndreaLanfranchi:
> > thread_local here means thread storage duration. That is similar to static storage duration
> 
> I don't understand the benefit of using thread_local for these variables? Do we want the initialization to occur only once? This doesn't seem correct to me.

The struct holding cursor is not light-weight (actually not a simple variable rather a complex object) and I understand @yperbasis's attempt to reduce memory re-allocations.

greg7mdp:
> I don't understand the benefit of using thread_local for these variables? Do we want the initialization to occur only once? This doesn't seem correct to me.

OK, never mind, I get it, you want to avoid calling `db::open_cursor()` every time. I don't understand mdbx enough to have an opinion of whether it is a good idea :-).

greg7mdp:
> The struct holding cursor is not light-weight (actually not a simple variable rather a complex object) and I understand @yperbasis's attempt to reduce memory re-allocations.

Thanks! Maybe instead of these APIs being called with mdbx::txn&, they could be called with a struct containing opened cursors to the relevant tables?

yperbasis:
@AndreaLanfranchi 's concern is absolutely valid (and his thorough comment is great!). I'll think about how it can be addressed.

On another note, this PR (or something similar) is a good optimization opportunity: I've just measured the time to execute all 14M main net blocks and with this PR it drops from 33h:23m to 28h:52m. Have to double check, but this avenue is looking promising.

AndreaLanfranchi:
@yperbasis I believe a viable path could be to replace the usage of `renew` with a `bind`.
If we do 
```
std::optional<BlockBody> read_body(mdbx::txn& txn, const Bytes& key, bool read_senders) {

    thread_local mdbx::cursor_managed src{};
    src.bind(txn, db::open_map(txn, db::table::kBlockBodies));
    
    [...]
}
```
We obtain that 
- the heavy lift malloc occurs only once (constructor of cursor_managed calls `mdbx_cursor_create` which is where `mdbx_calloc(1, sizeof(MDBX_cursor_couple))` occurs;
- `open_map` is relatively light-weight cause if the named bucket had already previous opened it returns immeditaly the dbi
- `bind` assigns the cursor struct to given map dbi within the scope of the transaction

Actually `cursor.renew()` is a bind
```
int mdbx_cursor_renew(MDBX_txn *txn, MDBX_cursor *mc) {
  return likely(mc) ? mdbx_cursor_bind(txn, mc, mc->mc_dbi) : MDBX_EINVAL;
}
```

AndreaLanfranchi:
FYI ... just tested the mod and node test still get a success ... not measured the perfomance though. I nevertheless expect results very similar to what you already have achieved.

AndreaLanfranchi:
What is "undesirable" is that basically each API hash it's own cursor storage so perhaps implementing a cursor pool might be a further improvement.

AndreaLanfranchi:
Just sketched this 
```
::mdbx::cursor_managed& open_pooled_cursor(::mdbx::txn& tx, const MapConfig& config){
    auto map{open_map(tx, config)};

    thread_local std::vector<mdbx::cursor_managed> cursors_pool{};
    const auto predicate= [](const mdbx::cursor_managed& cursor) -> bool {
        return !mdbx_cursor_txn(&(*cursor)); // Not bound to a live transaction
    };

    auto it{std::find_if(cursors_pool.begin(), cursors_pool.end(), predicate)};
    if(it != cursors_pool.end()) {
        it->bind(tx, map);
        return *it;
    }
    cursors_pool.emplace_back();
    cursors_pool.back().bind(tx,map);
    return cursors_pool.back();
}
```

And eventually change in access_layer the needed occurrencies of
```
auto src{db::open_cursor(txn, <whatever>)};
```
to
```
auto& src{db::open_pooled_cursor(txn, <whatever>)};
```

The mechanic works but have not measured performance

AndreaLanfranchi:
Further addendum ... using this technique though is important to call `cursor.close()` after every use : we cannot rely on RAII destructor and two (or more) subsequent openings (from pool) of the same named bucket cause two (or more) cursors to be created.

greg7mdp:
> Further addendum ... using this technique though is important to call `cursor.close()` after every use 

To enforce the close() using RAII, maybe we could use a class like:

```
class PooledCursor
{
public:
    PooledCursor(::mdbx::txn& tx,  <whatever>) {
        cursor_ = db::open_cursor(txn, <whatever>};
    }

    ~PooledCursor() {
        cursor_.close();
    }

    ::mdbx::cursor_managed& operator*() const noexcept { return cursor_; }
    ::mdbx::cursor_managed* operator->() const noexcept { return &cursor_; }

private:
    ::mdbx::cursor_managed& cursor_;
}
```

and use it like:

```
PooledCursor cursor(txn, <whatever>);
cursor->upsert(...)
```

AndreaLanfranchi:
@greg7mdp actually `cursor_managed` is already a RAII wrapper around `cursor` which does exactly what you described.
My point is related to cursors in a pool: we want to `close` them (not destruct them) so they stay in the pool as "unlinked" from any transaction so the pool can reuse it (with `bind`) instead of creating a new one.
Using the pool in fact we get a **a reference to the cursor** and even if the reference falls out of scope the original object remain in the last state it was used.

greg7mdp:
> My point is related to cursors in a pool: we want to `close` them (not destruct them)

Yes, that was what the PooledCursor destructor I propose is doing:

```
    ~PooledCursor() {
        cursor_.close();
    }
```

AndreaLanfranchi:
Either I don't follow or you propose something like this

```
class PooledCursor
{
public:
    PooledCursor(::mdbx::txn& tx,  <whatever>) {
        cursor_ = db::open_pooled_cursor(txn, <whatever>}; // <---- Note the "pooled"
    }

    ~PooledCursor() {
        cursor_.close();
    }

    ::mdbx::cursor_managed& operator*() const noexcept { return cursor_; }
    ::mdbx::cursor_managed* operator->() const noexcept { return &cursor_; }

private:
    ::mdbx::cursor_managed& cursor_;
}
```
And instead of 
```
auto& src{db::open_pooled_cursor(txn, <whatever>)};
src.find(....);
[...]
src.close();
```
Use 
```
PooledCursor src(txn, <whatever>);
*src.find(...);
 // do other stuff and let it fall out scope
```
Well this is also possible but imho adds a bit of overhead and eventually we need to change all the code using cursor to leverage `*` or `->` operators (massive change). Imho a simple crs.close() is a one liner change.

greg7mdp:
Yes, this is exactly what I was suggesting. I don't see it adding *any* overhead at all, and yes there would be changes in the code (changing `.` to `->`), but it would guarantee that close() is always called on all return paths, without having to add it explicitly.

AndreaLanfranchi (approved):
Does it provide similar results ?

yperbasis:
I've switched to `bind` per @AndreaLanfranchi 's suggestion.

yperbasis:
> Does it provide similar results ?

Yes, `bind` brings the same performance gain.

PR #550, opened 2022-01-27
Fix unnecessary std::string creation/destruction

greg7mdp:
This is trying to avoid creating a std::string in `to_big_compact()`, which is destroyed immediately after. The issue is that, with this change,   `to_big_compact()`  returns a `ByteView` which points to the static buffer of inside the function. As the result is usually used immediately after the call, I don't think it it a significant issue, but you guys might disagree of course :-).

### DIFF
--- a/core/silkworm/common/endian.cpp
+++ b/core/silkworm/common/endian.cpp
@@ -19,23 +19,23 @@
 #include <silkworm/common/util.hpp>
 
 #if defined(__wasm__)
-#define SILKWORM_THREAD_LOCAL
+#define SILKWORM_THREAD_LOCAL static
 #else
 #define SILKWORM_THREAD_LOCAL thread_local
 #endif
 
 namespace silkworm::endian {
 
-Bytes to_big_compact(const uint64_t value) {
+ByteView to_big_compact(const uint64_t value) {
     SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(uint64_t)];
     store_big_u64(&full_be[0], value);
-    return Bytes{zeroless_view(full_be)};
+    return zeroless_view(full_be);
 }
 
-Bytes to_big_compact(const intx::uint256& value) {
+ByteView to_big_compact(const intx::uint256& value) {
     SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(intx::uint256)];
     intx::be::store(full_be, value);
-    return Bytes{zeroless_view(full_be)};
+    return zeroless_view(full_be);
 }
 
 }  // namespace silkworm::endian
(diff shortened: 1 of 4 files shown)


AndreaLanfranchi (approved):
Makes perfect sense to me.
Just checked all usages of to_big_compact consume data into `Bytes` hence this change saves one copy contructor.

Nice catch

AndreaLanfranchi:
@yperbasis thoughts ?

AndreaLanfranchi (commented):
As an additional change I'd add an early exit for values which appear to be zero.
```
ByteView to_big_compact(const uint64_t value) {
    if(!value) {
      return {};
    }
    SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(uint64_t)];
    store_big_u64(&full_be[0], value);
    return zeroless_view(full_be);
}

ByteView to_big_compact(const intx::uint256& value) {
    if(!value) {
      return {};
    }
    SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(intx::uint256)];
    intx::be::store(full_be, value);
    return zeroless_view(full_be);
}
```

yperbasis:
> As an additional change I'd add an early exit for values which appear to be zero.
> 
> ```
> ByteView to_big_compact(const uint64_t value) {
>     if(!value) {
>       return {};
>     }
>     SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(uint64_t)];
>     store_big_u64(&full_be[0], value);
>     return zeroless_view(full_be);
> }
> 
> ByteView to_big_compact(const intx::uint256& value) {
>     if(!value) {
>       return {};
>     }
>     SILKWORM_THREAD_LOCAL uint8_t full_be[sizeof(intx::uint256)];
>     intx::be::store(full_be, value);
>     return zeroless_view(full_be);
> }
> ```

I wouldn't bother with this particular change `if(!value)` because:
a) It adds some code.
b) Modern compilers are often clever and modern hardware is very complicated, so one can never be sure whether a code change is actually an optimization w/o measurement.
c) The gain if probably very small if any.

AndreaLanfranchi:
> I wouldn't bother with this particular change `if(!value)` because: a) It adds some code. b) Modern compilers are often clever and modern hardware is very complicated, so one can never be sure whether a code change is actually an optimization w/o measurement. c) The gain is probably very small if any.

I don't tink is irrelevant as :
- is not a code a compiler can optimize (no constants here) imho
- saves a bswap and algorithm search if not needed
- clearly states the developer has examined the logic

Anyway ... if measurements needed I don't have the bandwidth now.

yperbasis (approved):

greg7mdp:
Thanks for the comments! I updated the comments, and added the early exit suggested by Andrew.

PR #551, opened 2022-01-28
Fix account decoding

AndreaLanfranchi:
- fix decode account from encoded rlp : 1) Check payload is either empty or > 1 (otherwise first `encoded_payload[pos++]` gives undefined behavior); 2) remove duplicate checks and reduce to a loop; 
- fix decode incarnation from encoded rlp : implementation did not include all checks of the previous
- move `decode_account_from_storage` to `Account::from_encoded_storage`
- move `extract_incarnation` to `Account::incarnation_from_encoded_storage`
- addedd missing tests for wrong payloads

### DIFF
--- a/core/silkworm/types/account.cpp
+++ b/core/silkworm/types/account.cpp
@@ -88,7 +88,7 @@ std::pair<Account, rlp::DecodingResult> Account::from_encoded_storage(ByteView e
     Account a{};
     if (encoded_payload.empty()) {
         return {a, rlp::DecodingResult::kOk};
-    } else if (encoded_payload.length() == 1) {
+    } else if (encoded_payload[0] && encoded_payload.length() == 1) {
         // Must be at least 2 bytes : field_set + len of payload
         return {a, rlp::DecodingResult::kInputTooShort};
     }
@@ -141,7 +141,7 @@ std::pair<Account, rlp::DecodingResult> Account::from_encoded_storage(ByteView e
 std::pair<uint64_t, rlp::DecodingResult> Account::incarnation_from_encoded_storage(ByteView encoded_payload) noexcept {
     if (encoded_payload.empty()) {
         return {0, rlp::DecodingResult::kOk};
-    } else if (encoded_payload.length() == 1) {
+    } else if (encoded_payload[0] && encoded_payload.length() == 1) {
         // Must be at least 2 bytes : field_set + len of payload
         return {0, rlp::DecodingResult::kInputTooShort};
     }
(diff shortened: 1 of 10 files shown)


core/silkworm/types/account.cpp
```diff
@@ -84,85 +84,96 @@ size_t Account::encoding_length_for_storage() const {
...
-}
-
-std::pair<Account, rlp::DecodingResult> decode_account_from_storage(ByteView encoded) noexcept {
+std::pair<Account, rlp::DecodingResult> Account::from_encoded_storage(ByteView encoded_payload) noexcept {
     Account a{};
-    if (encoded.empty()) {
+    if (encoded_payload.empty()) {
         return {a, rlp::DecodingResult::kOk};
+    } else if (encoded_payload.length() == 1) {
+        // Must be at least 2 bytes : field_set + len of payload
```
yperbasis, line 92: That's not right: `encode_for_storage` on an empty account will return a string of length 1 with a single 0 byte.

AndreaLanfranchi: Ok skimmed over it. The check has to be hardened so if `encoded_payload[0] != 0` then there MUST be something afterwards. Doing now

yperbasis: The simplest thing is to remove this check altogether. The length is already checked in the main loop.

AndreaLanfranchi: Done 8425b9b

core/silkworm/types/account.cpp
```diff
@@ -84,85 +84,109 @@ size_t Account::encoding_length_for_storage() const {
...
+                case 1: {
+                    const std::optional<uint64_t> nonce{
+                        endian::from_big_compact<uint64_t>(encoded_payload.substr(pos, len))};
+                    if (nonce == std::nullopt) {
+                        return {a, rlp::DecodingResult::kLeadingZero};
+                    }
+                    a.nonce = *nonce;
+                } break;
+                case 2:
+                    std::memcpy(&as_bytes(a.balance)[32 - len], &encoded_payload[pos], len);
```
yperbasis, line 129: Let's use `endian::from_big_compact<intx::uint256>` for consistency

AndreaLanfranchi: Sure ... I have question though.
Actually `from_big_compact` returns std::nullopt in two cases:

- either size of target is < size of data string (overflow)
- or there are leading zeroes and are not allowed

How we could efficiently make a distinction ?
Wouldn't it be better to transform `from_big_compact` into a parser into the rlp namespace ?
E.g.

```
template <typename UnsignedInteger>
static rlp::DecodingResult parse_big_compact(ByteView data, UnsignedInteger& target, bool allow_leading_zeros = false) {
 [...]
 target = be:load(x)
 return rlp::DecodingResult::kOk
)
```

yperbasis: Perhaps. I don't have a strong view on that. An argument against moving `parse_big_compact` into RLP is that it's not particularly RLP-specific, but rather a useful low-level function. If you decide to move it into RLP, please move it alongside `to_big_compact` and in a separate PR.

yperbasis (approved):

PR #552, opened 2022-02-01
Add logging unit tests and macros

canepat:
This PR introduces small backward-compatible additions to the logging package:
- macros (motivation: remove any overhead for level-filtered log traces and prepare for Silkrpc integration)
- unit tests

### DIFF
--- a/node/silkworm/common/log.hpp
+++ b/node/silkworm/common/log.hpp
@@ -90,8 +90,8 @@ class BufferBase {
 template <Level level>
 class LogBuffer : public BufferBase {
   public:
-    explicit LogBuffer() : BufferBase(level){};
-    explicit LogBuffer(std::string_view msg, std::vector<std::string> args = {}) : BufferBase(level, msg, args){};
+    explicit LogBuffer() : BufferBase(level) {}
+    explicit LogBuffer(std::string_view msg, std::vector<std::string> args = {}) : BufferBase(level, msg, args) {}
 };
 
 using Trace = LogBuffer<Level::kTrace>;
@@ -103,5 +103,18 @@ using Critical = LogBuffer<Level::kCritical>;
 using Message = LogBuffer<Level::kNone>;
 
 }  // namespace silkworm::log
+
+#define LOG(level_)                               \
+    if (!silkworm::log::test_verbosity(level_)) { \
+    } else                                        \
+        silkworm::log::LogBuffer<level_>()
+
+#define SILK_TRACE LOG(silkworm::log::Level::kTrace)
+#define SILK_DEBUG LOG(silkworm::log::Level::kDebug)
+#define SILK_INFO LOG(silkworm::log::Level::kInfo)
+#define SILK_WARN LOG(silkworm::log::Level::kWarning)
+#define SILK_ERROR LOG(silkworm::log::Level::kError)
+#define SILK_CRIT LOG(silkworm::log::Level::kCritical)
+#define SILK_LOG LOG(silkworm::log::Level::kNone)
 
 #endif  // !SILKWORM_COMMON_LOG_HPP_
--- a/node/silkworm/common/log.hpp
+++ b/node/silkworm/common/log.hpp
@@ -104,17 +104,17 @@ using Message = LogBuffer<Level::kNone>;
 
 }  // namespace silkworm::log
 
-#define LOG(level_)                               \
+#define SILK_LOGBUFFER(level_)                    \
     if (!silkworm::log::test_verbosity(level_)) { \
     } else                                        \
         silkworm::log::LogBuffer<level_>()
 
-#define SILK_TRACE LOG(silkworm::log::Level::kTrace)
-#define SILK_DEBUG LOG(silkworm::log::Level::kDebug)
-#define SILK_INFO LOG(silkworm::log::Level::kInfo)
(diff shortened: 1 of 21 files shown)


AndreaLanfranchi:
Not fully reviewed yet but my very first opinion is to avoid macros like a plague.

- loglevel is a runtime user settable value
- effective code is not what you read in the editor
- I believe branch prediction is good enough to void any overhead
- logging is not a "critical" cpu intensive part
- we can easily skip blocks of traces (which is actually what we're trying to hide) using enclosing them into `if` blocks

node/silkworm/common/log.hpp
```diff
@@ -104,4 +104,17 @@ using Message = LogBuffer<Level::kNone>;
 
 }  // namespace silkworm::log
 
+#define LOG(level_)                               \
```
AndreaLanfranchi, line 107: LOG is too similar to the keyword for a logarithmic function. (Which I now realize is also for the `log` namespace)
Maybe change to SILK_LOGGING_LEVEL ?

canepat:
> ... avoid macros like a plague.

Generally speaking, I couldn't agree more. I am proposing an exception to such rule just for logging purposes because it can reach overhead avoidance maintaining code clarity with a tiny addition on top of the existing Silkworm log facility. Maybe another way could exist using templates, but this solution seemed immediate without extra-effort.

> * effective code is not what you read in the editor

This is one major drawback, that's why I dislike macros as well and always avoid them except for logging

> * I believe branch prediction is good enough to void any overhead

TBH I really don't know if branch prediction can avoid also parameter manipulation happening before append, e.g.:
```
SILK_TRACE << "hash: " << hash(tx) << ...;
```
as opposed to
```
log::Trace() << "hash: " << hash(tx) << ...;
```
I took a look at generated assembly code for both cases and in the first case no instruction is present for the level-filtered branch: IMHO the branch prediction is well positioned to optimize such case, not so sure for the second one

> * logging is not a "critical" cpu intensive part

It could be If usage of log traces becomes more extensive as it is in Silkrpc and will probably be also in Silkworm node

> * we can easily skip blocks of traces (which is actually what we're trying to hide) using enclosing them into `if` blocks

IMHO this well-established technique makes code much more cluttered and unreadable. I used it for many years in production code except to regret it: these `if`blocks around logs make function bodies longer and are really distracting when reading code later on. Log macros just use the same trick but saving readability

node/silkworm/common/log.hpp
yperbasis: I agree that at least the name of this macro should start with `SILK_` or `SILKWORM_`. More rationale at https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros

canepat: You're right! Maybe `SILK_LOG_LINE` fits better? Just because it's not a logging level, it's a log line (it receives the `level_` as parameter)

AndreaLanfranchi:
> I took a look at generated assembly code for both cases and in the first case no instruction is present for the level-filtered branch: IMHO the branch prediction is well positioned to optimize such case, not so sure for the second one

Seem to me instructions [are generated as well](https://godbolt.org/z/44567hq75) if level is not known at compile time.

> It could be If usage of log traces becomes more extensive as it is in Silkrpc and will probably be also in Silkworm node

That is a smell imho. While I agree tracing is useful while debugging on release (if we aim for performance) all tracing should be enabled only if not defined NDEBUG

Last but not least ... this technique does not allow `log::Trace("Prefix..", {"arg1", "value1", [...]})` which produces erigon's like well formatted outputs.

AndreaLanfranchi:
One last comment ... if this **is not** a replacement of legacy syntax ... can go in for me.

node/silkworm/common/log.hpp
AndreaLanfranchi: Well actually is not a log line either: it eventually prepares the buffer to receive logging data.
My vote is for `SILK_LOGBUFFER(silkworm::log::Level::xxx)`  which reminds sort of a template.
But your call on this

canepat:
> > I took a look at generated assembly code for both cases and in the first case no instruction is present for the level-filtered branch: IMHO the branch prediction is well positioned to optimize such case, not so sure for the second one
> 
> Seem to me instructions [are generated as well](https://godbolt.org/z/44567hq75) if level is not known at compile time.

It is sufficient to add `-O1` compiler option to obtain [no instruction](https://godbolt.org/z/jdv5befzh) in the `if` branch

> 
> > It could be If usage of log traces becomes more extensive as it is in Silkrpc and will probably be also in Silkworm node
> 
> That is a smell imho. While I agree tracing is useful while debugging on release (if we aim for performance) all tracing should be enabled only if not defined NDEBUG
 
If I understand correctly, you are proposing to enable logs only in DEBUG builds. I think this requires some discussion within the team in terms of what this means for managing the Silkworm node in production

> Last but not least ... this technique does not allow `log::Trace("Prefix..", {"arg1", "value1", [...]})` which produces erigon's like well formatted outputs.

This must be definitively taken into account because it introduces in the discussion the usage of the Silkworm log facility for the application console. Again this should be discussed within the team

AndreaLanfranchi (approved):

yperbasis (approved):

PR #553, opened 2022-02-02
Plug HashState into syncloop

AndreaLanfranchi:
Completed unwind of Execution.
HashState stage completely rewritten:

- removed double scan on AccountChangeSet
- amended errors in unwind `to` parameter : when unwinding from X to X-5 previous version was in reality unwinding to X-6
- addresses are hashed only once : a map keep track of already hashed addresses
- reduced code
- removed usage of etl collector for incremental mode. Rationale : as incremental happens after first sync in 2GB Ram we can fit (at current figures) up to 200k block changes  in <2GB RAM


core/silkworm/common/util.hpp
```diff
@@ -88,4 +108,4 @@ std::vector<std::string> split(std::string_view source, std::string_view delimit
 
 }  // namespace silkworm
 
-#endif  // SILKWORM_COMMON_UTIL_HPP_
+#endif  // SILKWORM_COMMON_UTIL_H
```
yperbasis, line 111: Looks like an inadvertent change

AndreaLanfranchi: Right. Corrected

node/silkworm/db/tables.hpp
```diff
@@ -57,37 +67,123 @@ inline constexpr db::MapConfig kCallToIndex{"CallToIndex"};
...
-inline constexpr db::MapConfig kPlainContractCode{"PlainCodeHash"};
+
+//! \details Store contract code hash for given contract address + incarnation
+//! \def "Incarnation" how many times given account was SelfDestruct'ed.
+//! \struct
+//! \verbatim
+//!   key   : contract address (20 bytes) + incarnation (u64 BE)
+//!   value : code hash (32 bytes)
+//! \endverbatim
+inline constexpr db::MapConfig kPlainContractHash{"PlainCodeHash"};
```
yperbasis, line 124: Perhaps rename to `kPlainCodeHash` for consistency with the string and `kHashedCodeHash`.

cmd/common.cpp
```diff
@@ -31,7 +31,7 @@ struct HumanSizeParserValidator : public CLI::Validator {
     template <typename T>
     explicit HumanSizeParserValidator(T min, std::optional<T> max = std::nullopt) {
         std::stringstream out;
-        out << " in [" << min << " - " << (max.has_value() ? max.value() : "∞") << "]";
+        out << " in [" << min << " - " << (max.has_value() ? max.value() : "") << "]";
```
yperbasis, line 34: "inf" rather than "" instead of "∞" ?

core/silkworm/common/util.cpp
```diff
@@ -128,17 +128,28 @@ ByteView zeroless_view(ByteView data) {
...
         *dest++ = kHexDigits[b & 0x0f];  // Lo
     }
     return out;
 }
 
+std::string abridge(std::string_view input, size_t length) {
+    if (input.length() <= length) {
+        return std::string(input);
+    }
+    return std::string(input.substr(0, length)) + "…";
```
yperbasis, line 150:
```suggestion
    return std::string(input.substr(0, length - 1)) + "…";
```

node/silkworm/db/buffer.cpp
```diff
@@ -132,21 +131,17 @@ void Buffer::write_to_state_table() {
...
+            for (const auto& [incarnation, contract_storage] : it->second) {
                 Bytes prefix{storage_prefix(address, incarnation)};
 
-                const auto& contract_storage{contract.second};
-
                 // sort before inserting into the DB
-                storage_keys.clear();
-                for (const auto& x : contract_storage) {
-                    storage_keys.push_back(x.first);
+                std::vector<evmc::bytes32> locations;
```
yperbasis, line 138: This is probably a pessimization because you allocate and deallocate memory for `locations` for each loop iteration instead of growing capacity were it declared outside the loop.

node/silkworm/db/tables.hpp
AndreaLanfranchi: Yes.

cmd/common.cpp
AndreaLanfranchi: Ok.

node/silkworm/db/buffer.cpp
AndreaLanfranchi: Yes you're right I forgot to keep the outside loop declaration and use clear on each iteration.
But thinking more I believe the sorting of locations is completely unneeded : we're not appending to table but using `upsert_storage_value` which makes use of `find_value_suffix(state_cursor, storage_prefix, location)` to point to the exact dup'ed record (if any) and we upsert into table.
Hence I beleive the code 
```
                // sort before inserting into the DB
                std::vector<evmc::bytes32> locations;
                for (const auto& [location, _] : contract_storage) {
                    locations.push_back(location);
                }
                std::sort(locations.begin(), locations.end());
                for (const auto& location : locations) {
                    upsert_storage_value(state_table, prefix, location, contract_storage.at(location));
                }
```
should be simplified to 
```
                Bytes prefix{storage_prefix(address, incarnation)};
                for (const auto& [location, value] : contract_storage) {
                    upsert_storage_value(state_table, prefix, location, value);
                }
```
which saves a copy, a sort, a double lookup in contract_storage map

AndreaLanfranchi: I'm also not sure processing addresses in a sorted way brings any benefit.
```
    // Extract sorted index of unique addresses before inserting into the DB
    absl::btree_set<evmc::address> addresses;
    for (auto& x : accounts_) {
        addresses.insert(x.first);
    }
    for (auto& x : storage_) {
        addresses.insert(x.first);
    }
```
This adds overhead and forces lookups on both `accounts_` and `storage_` maps relying on the assumption "adjacent addresses" can be adjacent also on db which is not true.

AndreaLanfranchi: It _might_ bring some benefit if we allocate the `storage_prefix` key outside the loop and do a `memcpy` only when address changes (instead of building the key for each record in `storage_` map) ... but I believe storage changes are way less than EOA Accounts changes per block

AndreaLanfranchi: Also noticed `block_account_changes_` records also changes for identical accounts. 
See https://github.com/torquem-ch/silkworm/blob/master/node/silkworm/db/buffer.cpp#L54-L63

node/silkworm/db/buffer.cpp
```diff
@@ -158,36 +146,37 @@ void Buffer::write_to_db() {
...
-        for (const auto& account_entry : block_entry.second) {
-            data = ByteView{account_entry.first};
-            data.append(account_entry.second);
-            auto data_slice{to_slice(data)};
-            account_change_table.put(to_slice(change_key), &data_slice, MDBX_APPENDDUP);
+    Bytes change_key(8, '\0');
+    for (const auto& [block_num, account_changes] : block_account_changes_) {
+        endian::store_big_u64(change_key.data(), block_num);
+        for (const auto& [address, storage_encoded] : account_changes) {
+            Bytes change_value(kAddressLength + storage_encoded.length(), '\0');
```
yperbasis, line 169: Better declare `change_value` outside of the loops and resize it inside to avoid memory reallocation.

core/silkworm/common/util.cpp
yperbasis: ok, fine

node/silkworm/db/buffer.cpp
yperbasis: > It _might_ bring some benefit if we allocate the `storage_prefix` key outside the loop and do a `memcpy` only when address changes (instead of building the key for each record in `storage_` map) ... but I believe storage changes are way less than EOA Accounts changes per block

That's why we have to measure and it'd better be in separate PR. But anyway, I won't insist.

node/silkworm/db/util.cpp
```diff
@@ -103,15 +110,29 @@ std::optional<ByteView> find_value_suffix(mdbx::cursor& table, ByteView key, Byt
...
-    if (find_value_suffix(state_cursor, storage_prefix, location)) {
-        state_cursor.erase();
-    }
-    value = zeroless_view(value);
-    if (!value.empty()) {
-        Bytes data{location};
-        data.append(value);
-        state_cursor.upsert(to_slice(storage_prefix), to_slice(data));
+void upsert_storage_value(mdbx::cursor& state_cursor, ByteView storage_prefix, ByteView location, ByteView new_value) {
+    static const auto build_db_value = [](const ByteView& loc, const ByteView& val) -> Bytes {
```
yperbasis, line 114: The old code was simpler.

node/silkworm/stagedsync/common.hpp
```diff
@@ -103,11 +109,22 @@ class IStage {
     //! \brief This function implementation MUST be thread safe as is called asynchronously from ASIO thread
     [[nodiscard]] virtual std::vector<std::string> get_log_progress() = 0;
 
+    //! \brief Returns the key name of the stage instance
     [[nodiscard]] const char* name() const { return stage_name_; }
 
+    //! \brief Sets a stop request for instance;
+    virtual void stop() { stopping_.store(true); }
```
yperbasis, line 116: Suggestion: extract `Stoppable` common superclass from `ActiveComponent` & `IStage`.

node/silkworm/db/buffer.cpp
AndreaLanfranchi: See [54bce04](https://github.com/torquem-ch/silkworm/pull/553/commits/54bce04d93244532c481fd82b77845d0973e1ed2)
Only one allocation. No resizing ... only slice (view) is adjusted in length

node/silkworm/db/util.cpp
AndreaLanfranchi: I know **but was also less efficient** cause it forced an erase in any case (which causes MDBX to free a page and, eventually on update, reallocate a new one).
Here instead if old value matches new value no operation is performed and if not there is no additional seek of key for upsertion as it is performed on current record.

node/silkworm/db/buffer.cpp
AndreaLanfranchi: And what about this ? https://github.com/torquem-ch/silkworm/pull/553#discussion_r798842253

node/silkworm/db/buffer.cpp
```diff
@@ -158,36 +147,39 @@ void Buffer::write_to_db() {
...
-            data.append(account_entry.second);
-            auto data_slice{to_slice(data)};
-            account_change_table.put(to_slice(change_key), &data_slice, MDBX_APPENDDUP);
+    Bytes change_key(8, '\0');
+    Bytes change_value(kAddressLength + 128 /* see comment*/,
+                       '\0');  // Max size of encoded value is 85. We allocate - once - some byte more for safety
+                               // and avoid reallocation or resizing in the loop
+    for (const auto& [block_num, account_changes] : block_account_changes_) {
+        endian::store_big_u64(change_key.data(), block_num);
+        for (const auto& [address, storage_encoded] : account_changes) {
```
yperbasis, line 172: Better rename `storage_encoded` to something like `account_encoded` to avoid confusion with the storage.

AndreaLanfranchi: [74c3949](https://github.com/torquem-ch/silkworm/pull/553/commits/74c3949f712accece08dbc42f6387e0e979bb4a3)

node/silkworm/db/buffer.cpp
yperbasis: Again, we have to measure.

node/silkworm/stagedsync/common.hpp
AndreaLanfranchi: [817d851](https://github.com/torquem-ch/silkworm/pull/553/commits/817d85137c1adc23340c0d4e9857c261ef75bb1a)

node/silkworm/stagedsync/recovery/recovery_farm.hpp
```diff
@@ -50,6 +50,9 @@ class RecoveryFarm {
         }
     }
 
+    //! \brief Whether running tasks should stop
+    bool is_stopping() { return is_stopping_.load(); }
```
yperbasis, line 54: `RecoveryFarm` could also inherit `Stoppable`.

yperbasis (approved):

node/silkworm/stagedsync/recovery/recovery_farm.hpp
AndreaLanfranchi: Done

PR #558, opened 2022-02-04
Fix historical_block in cmd/check_changes & scan_txs

yperbasis:


AndreaLanfranchi:
I believe the same issue occurs in cmd/scan_txs

yperbasis:
> I believe the same issue occurs in cmd/scan_txs

Good catch; fixed in d28c48094533d00990d200d59aebc70a34412830.

AndreaLanfranchi (approved):

PR #561, opened 2022-02-13
Revisit some PR 553 performance changes

yperbasis:
Some performance changes introduced in PR #553 don't bring performance gains unfortunately, while complicating the code. In my testing the total execution time with & w/o these changes was the same (about 30 hours).

Also, the changes in PR 553 seem to cause random execution failures on Paolo's Ryzen machine.


greg7mdp:
> I'm measuring whether some performance changes introduced in PR #553 actually brought performance gains or not.

If I remember correctly, I think it will make more difference when the upstream calls are updated as well.

AndreaLanfranchi:
Are you able to compare database size and freelist pages after execution with and without this PR ?
`toolbox --datadir <whatever> tables`

Besides I do have improvements in speed too. (IntelNUC 32GB RAM 2TB NVMe)

yperbasis:
> Are you able to compare database size and freelist pages after execution with and without this PR ? `toolbox --datadir <whatever> tables`
> 
> Besides I do have improvements in speed too. (IntelNUC 32GB RAM 2TB NVMe)

OK, I'll try to compare database size and freelist pages, but it'll take a few days.

What was your OS, compiler, and the speed improvement you observed?

My test machine in Google Cloud is Ubuntu 21.10 with Intel Xeon CPU @ 2.80GHz, 64GB RAM and 3TB NVMe. Compiler: gnu 11.2.0.

PR #568, opened 2022-02-16
Intermediate hashes: split account & storage changes

yperbasis:
This is a major performance fix for incremental intermediated hashes. `PrefixSet` should be traversed in order – otherwise its performance is terrible. The problem was that account & storage changes use incompatible formats: the storage prefixes are _packed_ account address + incarnation + unpacked location, and the account prefixes are _unpacked_ account addresses.

Kudos to Johen for identifying this issue 🙏


jomuel (approved):
lgtm

PR #570, opened 2022-02-20
Don't use default data directory in tests

yperbasis:
It's dangerous to tamper with the default data directory (e.g. ~/Library/Silkworm), which could contain your real instance, while running unit tests.

### DIFF
--- a/node/silkworm/common/directories_test.cpp
+++ b/node/silkworm/common/directories_test.cpp
@@ -24,6 +24,19 @@ namespace silkworm {
 
 TEST_CASE("DataDirectory") {
     {
+        // Open and create a storage path
+        TemporaryDirectory tmp_dir0;
+        DataDirectory data_dir(/*base_path=*/tmp_dir0.path(), /*create=*/true);
+        REQUIRE(data_dir.exists());
+        REQUIRE_NOTHROW(data_dir.deploy());
+        REQUIRE_THROWS(data_dir.clear());
+
+        // Eventually delete the created paths
+        std::filesystem::remove_all(data_dir.path());
+        REQUIRE(data_dir.exists() == false);
+    }
+
+    {
         // Open datadir from current process running path
         DataDirectory data_dir{std::filesystem::path(), false};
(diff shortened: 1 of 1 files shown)


node/silkworm/common/directories_test.cpp
```diff
@@ -23,18 +23,6 @@
 namespace silkworm {
 
 TEST_CASE("DataDirectory") {
```
AndreaLanfranchi, line 25: I would not remove the test rather initialize it from a temporary directory.
We get the same result not to tamper with default storage path and keep the test
```suggestion
TEST_CASE("DataDirectory") {
    {
        // Open and create a storage path
        TemporaryDirectory tmp_dir0;
        DataDirectory data_dir(/*base_path=*/tmp_dir0.path(), /*create=*/true);
        REQUIRE(data_dir.exists());
        REQUIRE_NOTHROW(data_dir.deploy());
        REQUIRE_THROWS(data_dir.clear());

        // Eventually delete the created paths
        std::filesystem::remove_all(data_dir.path());
        REQUIRE(data_dir.exists() == false);
    }

```

AndreaLanfranchi (changes requested):

AndreaLanfranchi:
Besides ... the more I think about it the more I'm convinced the automagic selection of a default storage path is an error. I'd much prefer users to deal with **mandatory** `--datadir` CLI argument so they are aware of where their real data is.

node/silkworm/common/directories_test.cpp
AndreaLanfranchi: Or ... another option is to keep the test as it is and continue the test only if `data_dir.is_pristine() == true` (which means there is nothing in it). In such case the test for initialization from default storage path succeeds while subsequent deletions get skipped.

yperbasis: I've switched to your temporary directory suggestion just to be on the safe side and avoid doing anything in potentially sensitive directories.

AndreaLanfranchi (approved):

PR #571, opened 2022-02-20
Execution 6 akula style

AndreaLanfranchi:
In this PR :

- revisited `Execution` stage to behave as much as possible as Akula : the boundaries of _a batch of blocks_ is determined on behalf of processed gas instead of accrued changes to write; History batch size is separated from State batch size so we end up writing history more frequently than state and, by consequence we keep state in memory for longer hence producing more cache hits; blocks are read in _batches_ of 10240 each time instead of reading them one by one.
- introduce an amendment in workers synchronization for workers in `Senders` stage (two separate mutexes were incorrectly used)


cmd/toolbox.cpp
```diff
@@ -616,7 +616,7 @@ void do_compact(db::EnvConfig& config, const std::string& work_dir, bool replace
 }
 
 void do_copy(db::EnvConfig& src_config, const std::string& target_dir, bool create, bool noempty,
-             std::vector<std::string>& names, std::vector<std::string>& xnames, bool dry) {
+             std::vector<std::string>& names, std::vector<std::string>& xnames) {
```
greg7mdp, line 619: Can `std::vector<std::string>& names` and `std::vector<std::string>& xnames` be const parameters?

node/silkworm/common/log.cpp
```diff
@@ -30,6 +30,7 @@ namespace silkworm::log {
 static Settings settings_{};
 static std::mutex out_mtx{};
 static std::unique_ptr<std::fstream> file_{nullptr};
+thread_local std::string thread_name_{};
```
greg7mdp, line 33: Why not declare this inside the `get_thread_name` function?

node/silkworm/db/access_layer.cpp
```diff
@@ -194,17 +210,14 @@ bool read_block(mdbx::txn& txn, BlockNum block_number, bool read_senders, BlockW
...
-    data = headers_cursor.find(to_slice(key), false);
-    if (!data) {
+    auto raw_header{read_header_raw(txn, key)};
+    if (raw_header.empty()) {
         return false;
     }
-
-    ByteView data_view(from_slice(data.value));
-    rlp::success_or_throw(rlp::decode(data_view, bh.block.header));
+    ByteView raw_header_view{raw_header.data(), raw_header.length()};
```
greg7mdp, line 219: ByteView has a constructor for Bytes, so I think this could be written:
`ByteView raw_header_view{raw_header);`

node/silkworm/db/access_layer.cpp
```diff
@@ -260,25 +261,50 @@ void write_body(mdbx::txn& txn, const BlockBody& body, const uint8_t (&hash)[kHa
...
+    if (out.empty()) {
+        return;
+    }
+    auto data_view{read_senders_raw(txn, key)};
+    if (!data_view.empty()) {
+        SILKWORM_ASSERT(data_view.length() % kAddressLength == 0);
+        SILKWORM_ASSERT(data_view.length() / kAddressLength == out.size());
+        auto addresses = reinterpret_cast<const evmc::address*>(data_view.data());
+        size_t idx{0};
+        for (auto& transaction : out) {
```
greg7mdp, line 297: const auto& ?

greg7mdp (commented):
Nice! Silkworm is really an outstanding project and you guys are doing amazing work. I wish I worked more on it so I could do meaningful reviews rather than just some basic reamrks. Keep up the good work please!

node/silkworm/common/stopwatch.cpp
```diff
@@ -57,7 +57,7 @@ StopWatch::Duration StopWatch::lap_duration() noexcept {
 }
 
 StopWatch::Duration StopWatch::since_start(const TimePoint& origin) noexcept {
-    if (!started_) {
+    if (start_time_ == TimePoint()) {
```
yperbasis, line 60: Would it make sense to start StopWatch on construction?

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -129,24 +193,34 @@ StageResult Execution::execute_batch(db::RWTxn& txn, BlockNum max_block_num, Blo
 
             // TODO(Andrea) implement pruning
             buffer.insert_receipts(block_num_, receipts);
+
+            // Stats
             std::unique_lock progress_lock(progress_mtx_);
-            ++processed_blocks_;
-            processed_transactions_ += block_with_hash.block.transactions.size();
-            processed_gas_ += block_with_hash.block.header.gas_used;
+            processed_blocks_++;
```
yperbasis, line 199: (Minor) Google Code Style recommends pre-increment: https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -92,34 +91,99 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
+std::queue<Block> Execution::prefetch_blocks(db::RWTxn& txn, BlockNum from, BlockNum to, size_t max_blocks) {
+    static std::unique_ptr<StopWatch> sw;
+    bool should_trace{log::test_verbosity(log::Level::kTrace)};
+    if (should_trace) {
+        sw = std::make_unique<StopWatch>();
+        sw->start();
+    }
+
+    std::queue<Block> ret{};
+    BlockNum reached_block_num{0};
```
yperbasis, line 103: Could be moved inside the loop.

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -92,34 +91,99 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
 
-        BlockWithHash block_with_hash;
+        std::queue<Block> prefetched_blocks{prefetch_blocks(txn, block_num_, max_block_num, 10240)};
 
         while (true) {
+            if (prefetched_blocks.empty()) {
+                if (is_stopping()) {
+                    return StageResult::kAborted;
+                }
+                prefetched_blocks = prefetch_blocks(txn, block_num_, max_block_num, 10240);
```
yperbasis, line 169: Better extract a constant (10240 is also used a few lines above).

node/silkworm/db/buffer.cpp
```diff
@@ -57,22 +53,30 @@ void Buffer::update_account(const evmc::address& address, std::optional<Account>
...
 
     if (equal) {
         return;
     }
-
-    if (accounts_.insert_or_assign(address, current).second) {
-        bump_batch_size(kAddressLength, current ? current->encoding_length_for_storage() : 0);
+    auto it{accounts_.find(address)};
+    if (it != accounts_.end()) {
+        accounts_[address] = current;
```
yperbasis, line 69: Why combination of `find` & `operator[]` instead of `insert_or_assign`?

node/silkworm/db/buffer.cpp
```diff
@@ -96,19 +101,188 @@ void Buffer::update_storage(const evmc::address& address, uint64_t incarnation,
...
+        if (block_storage_changes_[block_number_][address][incarnation]
+                .insert_or_assign(location, initial_val)
+                .second) {
+            batch_history_size_ += kPlainStoragePrefixLength + kHashLength + initial_val.size();
         }
     }
 
     if (storage_[address][incarnation].insert_or_assign(location, current).second) {
-        bump_batch_size(kPlainStoragePrefixLength, kHashLength + zeroless_view(current).size());
+        batch_state_size_ += kPlainStoragePrefixLength + kHashLength + kHashLength;
```
yperbasis, line 113: `kHashLength` instead of `zeroless_view(current).size())` overestimates batch size. A lot of storage values are relatively small numbers, nowhere close to 2^256.

node/silkworm/db/buffer.cpp
greg7mdp: The new version doesn't insert if address was not already in accounts_. Don't know if this was the intent.

Also why do 2 lookups and 2 constructor calls in the new version, instead of:

```
    auto it{accounts_.find(address)};
    if (it != accounts_.end()) {
        *it = current;
```

node/silkworm/common/log.cpp
AndreaLanfranchi: Cause it can be valued also by `set_thread_name`

node/silkworm/db/access_layer.cpp
AndreaLanfranchi: Yes of course ... I think is a matter of taste to use initialization list vs ctor. 
Will change

node/silkworm/common/stopwatch.cpp
AndreaLanfranchi: Yes but with defaulted parameter in constructor. Say you want to create a Stopwatch far from where it is effectively used and pass it around by reference in several functions then the recorded timings wouldn't be accurate.

node/silkworm/stagedsync/stage_execution.cpp
AndreaLanfranchi: Ok

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -92,34 +91,99 @@ StageResult Execution::forward(db::RWTxn& txn) {
     return is_stopping() ? StageResult::kAborted : StageResult::kSuccess;
 }
 
+std::queue<Block> Execution::prefetch_blocks(db::RWTxn& txn, BlockNum from, BlockNum to, size_t max_blocks) {
+    static std::unique_ptr<StopWatch> sw;
```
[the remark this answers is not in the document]
AndreaLanfranchi: Ok

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -92,34 +91,99 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
+    bool should_trace{log::test_verbosity(log::Level::kTrace)};
+    if (should_trace) {
+        sw = std::make_unique<StopWatch>();
+        sw->start();
+    }
+
+    std::queue<Block> ret{};
+    BlockNum reached_block_num{0};
+    auto hashes_table{db::open_cursor(*txn, db::table::kCanonicalHashes)};
```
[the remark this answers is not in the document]
AndreaLanfranchi: We already have it. Is called `read_block`.  Problem with that one is that it opens a cursor and looks up a canonical hash for every block request. Here instead `CanonicalHashes` table is opened only once (every 10240 blocks) and read sequentially by cursor (10240 savings in rebinding cursor and in lookups).

node/silkworm/stagedsync/stage_execution.cpp
AndreaLanfranchi: Ok

node/silkworm/db/buffer.cpp
AndreaLanfranchi: I find the `insert_or_assign` semantic hard to digest. Checking the bool value returned due to an update (instead of new insertion) causes something like this `!accounts_.insert_or_assign(address, current).second` which always read, to me, like a failure in updating the map. But will change

node/silkworm/db/buffer.cpp
AndreaLanfranchi: I don't agree ... we're measuring the occupation in RAM of `storage_` which is not zerolessed until is flushed to database.
Actual RAM occupation is its full size, not its zerolessed view size.

node/silkworm/db/buffer.cpp
AndreaLanfranchi: Correction ... won't change as insert_or_assign can't give me the previous value in case of update and to keep the size accounting as correct as possibile when an update occurs I substract the size of previous value and add the size of current value.

yperbasis (approved):

node/silkworm/db/buffer.cpp
AndreaLanfranchi: @greg7mdp 
you're right in using the iterator but your snippet does not use it properly.
Should be `it->second = current`

node/silkworm/db/buffer.cpp
AndreaLanfranchi: BTW ... this comment made me realize was doing the wrong accounting on in memory `accounts_`
Thanks for heads up

node/silkworm/db/buffer.cpp
greg7mdp: yes indeed!

PR #572, opened 2022-02-21
Remove batch size from db::Buffer

yperbasis:
After PR #571 batch size is used only for trace logging, while adding considerable code complexity.

### DIFF
--- a/node/silkworm/db/buffer.cpp
+++ b/node/silkworm/db/buffer.cpp
@@ -17,7 +17,6 @@
 #include "buffer.hpp"
 
 #include <algorithm>
-#include <iostream>
 
 #include <absl/container/btree_set.h>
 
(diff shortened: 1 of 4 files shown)


AndreaLanfranchi:
While I agree on the rationale of this PR I think is premature to get rid of size accounting basically for a single reason: relying on gas accounting only may lead to huge in memory states if we execute large chains with low gas per block. Also actual settings work well for machines with 32GB+ RAM while on 16GBRAM or less excess memory consumption could worsen performances due to swapping and/or less space available for MDBX cache.

yperbasis:
OK, I'll keep it as draft for now.

yperbasis:
@AndreaLanfranchi I agree with, but then the proper way is to use the batch size instead of gas counting. Artem used gas counting in Akula not because it was somehow better than batch sizes, but due to some Akula code technicalities.

AndreaLanfranchi:
@yperbasis there are a lot of considerations IMHO that should be taken into account. 

- while I agree gas accounting by Akula is a workaraound it showed me some interesting points of view
- yes the main factor of speed actually achieved by execution is due to the split of threshold on history and state with the latter kept for long in memory and as a result producing more hits

But while I agree we should achieve control over RAM occupation and flush it on threshold, try picture this : imagine a 30M block chain with billions of transactions with an overall number of unique accounts, say, below 50k (I know is a very edgy case but ... you never know). Under this circumnstance you'd never reach a `state size` threshold and even on very first sync you'll end up with an execution stage which does _all-or-nothing_ in a single DB transaction. And considering `Execution` is where the heavy lift occurs then is highly desirable to make it restartable (from very last checkpoint).

I believe a balance amongst gas accounted and size accounted is a good compromise. 
Or a third way is to **borrow** another idea from Akula (actually drafted but not completely implemented): implement a timed commit. (i.e. for example on execution stage commit every 1 hour regardless any accrued size in gas or bytes).

AndreaLanfranchi:
I understand I am thinking beyond the boundaries of mainnet and testnets ... but I believe a little more effort and suffering the pain o little more complex code now will save us many head aches in the future.

yperbasis:
To my mind the simplest solution is to base commits on the batch size. If that proves a problem in the future, we can revisit that. I guess I'm cautious about introducing code complexity to fight future problems before actually experiencing them.

In any case, we both agree that batch size is useful, so I'm closing this PR.

PR #573, opened 2022-02-21
Small fixes

AndreaLanfranchi:
Minor fixes

- lock guard all HashState stats access
- remove UTF-8 ellipsis from abridged text and replace with triple ASCII dot
- better duration formatting : for durations >= 1 minute we get a formatting like `[xd] [xxh] [xxm] [xxs]`; for durations < 1m we get seconds (or lower milliseconds or microseconds) with a precision of three digits. Examples (from tests) :
```
    CHECK(sw1.format(255h + 12min + 14s) == "10d 15h 12m 14s");
    CHECK(sw1.format(240h) == "10d");
    CHECK(sw1.format(240h + 14s) == "10d 14s");
    CHECK(sw1.format(7min + 12s + 120ms) == "7m 12s");
    CHECK(sw1.format(1ms) == "1ms");
    CHECK(sw1.format(1200ms) == "1.200s");
    CHECK(sw1.format(1010us) == "1.010ms");
    CHECK(sw1.format(20us) == "20us");
```

### DIFF
--- a/node/silkworm/etl/collector.hpp
+++ b/node/silkworm/etl/collector.hpp
@@ -76,7 +76,7 @@ class Collector {
 
     //! \brief Returns the hex representation of current load key (for progress tracking)
     [[nodiscard]] std::string get_load_key() const {
-        std::unique_lock l{mutex_};
+        std::atomic_thread_fence(std::memory_order_acquire);
         return loading_key_;
     }
 
@@ -86,8 +86,8 @@ class Collector {
     void flush_buffer();  // Write buffer to file
 
     void set_loading_key(ByteView key) {
-        std::unique_lock l{mutex_};
         loading_key_ = to_hex(key, true);
+        std::atomic_thread_fence(std::memory_order_release);
     }
 
     bool work_path_managed_;
@@ -109,7 +109,6 @@ class Collector {
 
     std::vector<std::unique_ptr<FileProvider>> file_providers_;  // Collection of file providers
     size_t size_{0};                                             // Total collected size
-    mutable std::mutex mutex_{};                                 // To sync loading_key_
     std::string loading_key_{};                                  // Actual load key (for log purposes)
 };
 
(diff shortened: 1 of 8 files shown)


node/silkworm/etl/collector.hpp
```diff
@@ -76,7 +76,7 @@ class Collector {
 
     //! \brief Returns the hex representation of current load key (for progress tracking)
     [[nodiscard]] std::string get_load_key() const {
-        std::unique_lock l{mutex_};
```
yperbasis, line 79: This is not correct. Mutexes  [do synchronize memory](https://stackoverflow.com/questions/15313296/release-acquire-semantics-wrt-stdmutex). [atomic_thread_fence](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence) is useful for atomic variables, which `loading_key_` is not. The previous code with mutex was fine.

yperbasis (changes requested):

node/silkworm/etl/collector.hpp
AndreaLanfranchi: > The previous code with mutex was fine.

Well actually it wasn't : see https://erigon5900.weblogix.it/er.log
image
image
image

AndreaLanfranchi: atomic_thread_fence is useful for NON atomic variables (like string)

AndreaLanfranchi: Oh well missed that.

node/silkworm/stagedsync/stage_hashstate.cpp
```diff
@@ -350,6 +360,7 @@ StageResult HashState::hash_from_plaincode(db::RWTxn& txn) {
         if (!is_stopping()) {
             if (!collector_->empty()) {
                 source = db::open_cursor(*txn, db::table::kHashedCodeHash);
+                std::unique_lock lck(log_mtx_);
```
yperbasis, line 363: `collector_->load` could be a long operation, better perform it outside the lock.

AndreaLanfranchi:
Too many changes ... rewrite all !!!!

PR #576, opened 2022-02-24
Hashstate take2

AndreaLanfranchi:
- ensure logging vars are lock guarded
- ancillary function to check whether stop or continue
- ancillary function to very block numbering sequence
- code reuse


yperbasis:
No need to do it in this PR, but IMHO we should move things like `class HashState` into their own headers (e.g. `stage_hashstate.hpp`) out from `stagedsync.hpp`.

node/silkworm/stagedsync/stage_hashstate.cpp
```diff
@@ -26,9 +26,7 @@
 namespace silkworm::stagedsync {
 
 StageResult HashState::forward(db::RWTxn& txn) {
-    if (is_stopping()) {
-        return StageResult::kAborted;
-    }
+    continue_or_throw();
```
yperbasis, line 29: Hmm. I'm not sure about throwing instead of returning an error code. IMHO the previous version was better. It's not that I'm against exceptions, but we should be consistent. If we rely on exceptions, then `HashState::forward` should be `void`, not return `StageResult`.

node/silkworm/stagedsync/common.hpp
```diff
@@ -112,10 +112,16 @@ class IStage : public Stoppable {
     //! \brief Returns the key name of the stage instance
     [[nodiscard]] const char* name() const { return stage_name_; }
 
+    //! \brief Forces an exception if stage has been requested to stop
+    void continue_or_throw() { success_or_throw(is_stopping() ? StageResult::kAborted : StageResult::kSuccess); }
```
yperbasis, line 116: Perhaps `throw_if_stopping`?

yperbasis (changes requested):
`HashState::get_log_progress` should be protected by `log_mtx_`

(Minor) Also, I find throwing error codes and then catching exceptions and returning error codes a bit inelegant. Perhaps a more elegant solution could be found with the help of [gsl::finally](https://docs.microsoft.com/en-us/cpp/code-quality/c26448?view=msvc-170). But it's not that big a deal I guess.

AndreaLanfranchi:
> `HashState::get_log_progress` should be protected by `log_mtx_`

Doh ... forgot the most important point.

> (Minor) Also, I find throwing error codes and then catching exceptions and returning error codes a bit inelegant. Perhaps a more elegant solution could be found with the help of [gsl::finally](https://docs.microsoft.com/en-us/cpp/code-quality/c26448?view=msvc-170). But it's not that big a deal I guess.

Actually is for me reasonable to have error logs printed correctly (on behalf of exception type) instead of repeating log lines before every return for error.
gsl::finally is more like defer in go ... only one code block

node/silkworm/stagedsync/common.hpp
AndreaLanfranchi: Done

yperbasis (approved):

PR #580, opened 2022-02-24
expected in from_big_compact

yperbasis:
Use [el::expected](https://github.com/TartanLlama/expected) instead of `std::optional` in `from_big_compact` in order to differentiate between Overflow & LeadingZero errors. el::expected is an implementation of [std::expected](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p0323r12.html), which has been accepted into C++23.

Also merge `DecodingResult::kNonCanonicalSingleByte` into `kNonCanonicalSize`.


AndreaLanfranchi:
Only an opinion. Seem excessively convoluted and complicated way instead of using a simpler out parameter and return code.

greg7mdp:
> Only an opinion. Seem excessively convoluted and complicated way instead of using a simpler out parameter and return code.

yes, I was also thinking why not just return a `std::pair<UnsignedInteger, BigCompactError>` and use:

`auto [x, err] = from_big_compact(...)`

yperbasis:
> Only an opinion. Seem excessively convoluted and complicated way instead of using a simpler out parameter and return code.

You probably have a point: we already use error codes & output parameters, so why introduce a new error handling mechanism. I'll prepare an alternative PR.

yperbasis:
Closing in favour of PR #582.

PR #582, opened 2022-02-25
DecodingResult in from_big_compact

yperbasis:


core/silkworm/common/endian.hpp
```diff
@@ -185,31 +186,33 @@ ByteView to_big_compact(uint64_t value);
...
+//! \brief Parses unsigned integer from a compacted big endian byte form.
+//! \param [in] data : byte view of compacted value. Length must be <= sizeof(UnsignedInteger);
+//! otherwise kOverflow is returned.
+//! \param [out] out: the corresponding integer with native endianness.
+//! \return kOk or kOverflow or kLeadingZero.
+//! \remarks A "compact" big endian form strips leftmost bytes valued to zero;
+//! if the input is not compact kLeadingZero is returned.
 template <typename UnsignedInteger>
-static std::optional<UnsignedInteger> from_big_compact(ByteView data, bool allow_leading_zeros = false) {
+static DecodingResult from_big_compact(ByteView data, UnsignedInteger& out) {
```
AndreaLanfranchi, line 197: Why remove the possibility to "allow_leading_zeroes" ? I can't recall the rationale but I guess there was a reason for it to be there.

AndreaLanfranchi: Also shouldn't we specialize the template ?
```
template <typename UnsignedInteger,
          typename std::enable_if<std::is_integral<UnsignedInteger>::value && std::is_unsigned<UnsignedInteger>::value, T>::type* = nullptr>
```

core/silkworm/common/endian_test.cpp
```diff
@@ -121,8 +121,9 @@ TEST_CASE("Block as key and compact form") {
         auto block_number_compact_bytes{to_big_compact(block_number)};
         CHECK(to_hex(block_number_compact_bytes) == "5485ffde");
         // Convert back and check
-        auto block_number_from_compact{from_big_compact<uint64_t>(block_number_compact_bytes)};
-        CHECK(block_number_from_compact == block_number);
+        uint64_t out64;
```
AndreaLanfranchi, line 124: Shouldn't be initialized regardless ?

core/silkworm/common/endian.hpp
```diff
@@ -185,31 +186,33 @@ ByteView to_big_compact(uint64_t value);
...
     if (data.length() > sizeof(UnsignedInteger)) {
-        return std::nullopt;
+        return DecodingResult::kOverflow;
     }
 
-    UnsignedInteger x{0};
-
+    out = 0;
```
AndreaLanfranchi, line 202: Imho this is questionable ... should be duty of higher caller to ensure out is validly initialized.
Any error condition should NOT touch out.

core/silkworm/common/endian.hpp
yperbasis: Unfortunately `is_integral` is not true for `intx`. Perhaps @chfast can enable it.

yperbasis: > Why remove the possibility to "allow_leading_zeroes" ? I can't recall the rationale but I guess there was a reason for it to be there.

I dropped allow_leading_zeroes for simplicity. It was used only in one place in cmd/consensus, where I replaced it with `std::stoull` base=16).

core/silkworm/common/endian_test.cpp
yperbasis: Done in https://github.com/torquem-ch/silkworm/pull/582/commits/d0892f5083faca4051ef738e2d3876a928f7e602

core/silkworm/common/endian.hpp
chfast: > Unfortunately `is_integral` is not true for `intx`. Perhaps @chfast can enable it.

The `std::is_intergral` is not user-extendable. https://en.cppreference.com/w/cpp/types/is_integral

AndreaLanfranchi (approved):

core/silkworm/common/endian.hpp
```diff
@@ -185,31 +186,30 @@ ByteView to_big_compact(uint64_t value);
...
-    if (data.empty()) {
-        return x;
+    if (!data.empty() && data[0] == 0) {
+        return DecodingResult::kLeadingZero;
     }
 
-    if (data[0] == 0 && !allow_leading_zeros) {
-        return std::nullopt;
-    }
+    out = 0;
```
AndreaLanfranchi, line 206: Just a note ... This sets to zero all bytes of out.
Probably a `memset` to zero for all bytes NOT overwritten by subsequent `memcpy` depicts better the logic of the function but I understand it adds some overhead in computing.

PR #588, opened 2022-02-27
[WIP] Small clean-ups

yperbasis:
This is a small code clean-up (extract `kMinGasLimit` & `kMaxGasLimit` constants). See also [EIP-4803](https://eips.ethereum.org/EIPS/eip-4803): Limit transaction gas to a maximum of 2^63-1.

Also a small clean-up in `calculate_ripemd_160`: per [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs) out params should come after inputs.

### DIFF
--- a/core/silkworm/chain/protocol_param.hpp
+++ b/core/silkworm/chain/protocol_param.hpp
@@ -1,5 +1,5 @@
 /*
-   Copyright 2020-2021 The Silkworm Authors
+   Copyright 2020-2022 The Silkworm Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -52,8 +52,10 @@ namespace fee {
 
 namespace param {
 
-    // https://eips.ethereum.org/EIPS/eip-170
-    inline constexpr size_t kMaxCodeSize{0x6000};
+    inline constexpr size_t kMaxCodeSize{0x6000};  // EIP-170
+
+    inline constexpr uint64_t kMinGasLimit{5000};
+    inline constexpr uint64_t kMaxGasLimit{INT64_MAX};  // EIP-4803
 
     inline constexpr uint64_t kBlockRewardFrontier{5 * kEther};
     inline constexpr uint64_t kBlockRewardByzantium{3 * kEther};
--- a/core/silkworm/consensus/base/engine.cpp
+++ b/core/silkworm/consensus/base/engine.cpp
@@ -102,13 +102,7 @@ ValidationResult EngineBase::validate_block_header(const BlockHeader& header, co
         return ValidationResult::kGasAboveLimit;
     }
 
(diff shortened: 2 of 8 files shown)


core/silkworm/chain/protocol_param.hpp
```diff
@@ -52,8 +52,10 @@ namespace fee {
 
 namespace param {
 
-    // https://eips.ethereum.org/EIPS/eip-170
-    inline constexpr size_t kMaxCodeSize{0x6000};
+    inline constexpr size_t kMaxCodeSize{0x6000};  // EIP-170
+
+    inline constexpr uint64_t kMinGasLimit{5000};
+    inline constexpr uint64_t kMaxGasLimit{INT64_MAX};  // EIP-4803
```
AndreaLanfranchi, line 58: Do we really need to re-define an already existent constant ? Can't we alias it with `using` ?

core/silkworm/consensus/base/engine.cpp
```diff
@@ -102,13 +102,7 @@ ValidationResult EngineBase::validate_block_header(const BlockHeader& header, co
...
     }
 
-    if (header.gas_limit < 5000) {
-        return ValidationResult::kInvalidGasLimit;
-    }
-
-    // https://github.com/ethereum/go-ethereum/blob/v1.9.25/consensus/ethash/consensus.go#L267
-    // https://eips.ethereum.org/EIPS/eip-1985
-    if (header.gas_limit > INT64_MAX) {
+    if (header.gas_limit < param::kMinGasLimit || header.gas_limit > param::kMaxGasLimit) {
```
AndreaLanfranchi, line 105: Please explain me ... EIP-4803 talks about **transaction gas limit** here instead we check for block gas being over limit. It is obvious that overall block gas limit exceeds `kMaxGasLimit` if any of its transactions exceeds the limit but shouldn't the check be performed at transaction level ?

AndreaLanfranchi: Rethinking ... probably previous version was correct already with proper reference EIP-1985
I believe the meaning of 4803 is to early reject transactions at tx-pool level: if it's already in a block is the block gas check which holds (EIP-1985) not EIP-4803

> Backwards Compatibility
While this is a breaking change, no actual effect should be visible.

>Before [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) it was possible to include transactions with gasPrice = 0 and thus the gasLimit * gasPrice <= accountBalance calculation could have allowed for arbitrarily large values of gasLimit. However, the rule that the transaction list cannot exceed the block gas limit, and the strict rules about how the block gas limit can change, prevented arbitrarily large values of gasLimit to be in the historical state.

yperbasis: My bad, I somehow confused block and transaction gas limits.

PR #589, opened 2022-02-27
Cursor pool

AndreaLanfranchi:
###  This is a proof-of-concept PR for evaluation purposes.
It rewrites the goals expressed in #543 with these major differences:

- cursors have no longer `thread_static` storage lifetime within the scope of the functions where they're (re)used
- cursors can be pulled from a per-thread pool and, after use, returned to the pool
- before re-use the pool checks whether the cursors is bound to any live mdbx transaction : in such case the cursor is firstly closed and then recreated (only if the willing-to-reuse transaction has a different id from the original transaction which has generated the cursor)

Work relies on hints by @greg7mdp 

All of this solves the issues @mriccobene [described here](https://discord.com/channels/687972960811745322/733288466837799002/946745204705206282).
It is my opinion though that having two transactions in the same thread is conceptually wrong : to be determined if it's due to unsync'ed parallelism or deliberate.

Besides this approach saves some memory as in previous version we had cursors opened forever even if used only once (see for example the ones to pull network config at the very beginning of the process).

I'm launchin an execution batch to measure loss of performance (which intuitively should be)


node/silkworm/db/mdbx.cpp
```diff
@@ -160,6 +163,37 @@ ::mdbx::cursor_managed open_cursor(::mdbx::txn& tx, const MapConfig& config) {
...
+PooledCursor::PooledCursor(::mdbx::txn& tx, const MapConfig& config) {
+    cursor_ = cursors_pool_.acquire();
+    if (!cursor_) {
+        cursor_ = std::make_unique<::mdbx::cursor_managed>();
+    }
+    bind(tx, config);
+}
+
+PooledCursor::~PooledCursor() {
+    if (cursor_->operator bool()) {
```
greg7mdp, line 175: why not `if (cursor_)`

node/silkworm/common/object_pool.hpp
```diff
@@ -0,0 +1,55 @@
...
+        }
+        return {nullptr};
+    }
+
+    [[nodiscard]] bool empty() const { return pool_.empty(); }
+
+    [[nodiscard]] size_t size() const { return pool_.size(); }
+
+  private:
+    std::stack<std::unique_ptr<T>> pool_;
```
greg7mdp, line 50: I would just use std::vector, where `back()` returns a non-const reference, so you acquire() could return the unique_ptr.

node/silkworm/db/access_layer.cpp
```diff
@@ -24,14 +24,13 @@
 
 namespace silkworm::db {
 
-std::optional<VersionBase> read_schema_version(mdbx::txn& txn) noexcept {
-    thread_local mdbx::cursor_managed src;
-    src.bind(txn, db::open_map(txn, db::table::kDatabaseInfo));
-    if (!src.seek(mdbx::slice{kDbSchemaVersionKey})) {
+std::optional<VersionBase> read_schema_version(mdbx::txn& txn) {
+    PooledCursor src(txn, db::table::kDatabaseInfo);
```
greg7mdp, line 28: shouldn't the cursor pool still be thread_local?

node/silkworm/db/mdbx.cpp
AndreaLanfranchi: Well `if (cursor_)` returns whether the unique_pointer has a value.
Propably you meant `if(*cursor_)` which is basically a more concise syntax of what I wrote: I find it more readable in the sense that invokes the boolean operator on unique_ptr held value

node/silkworm/common/object_pool.hpp
AndreaLanfranchi: But that would mean the vector returns a ref to the unique_ptr.
Here instead the unique_ptr is moved away from it's container and is not guaranteed it returns into the pool

node/silkworm/db/mdbx.cpp
greg7mdp: I don't think so. `if(*cursor_)` would crash is the unique_ptr was not set.
`if (cursor_)` converst the cursor to bool (using cursor_->operator bool()) and test that boolean, so it is exactly equivalent to what you wrote.

node/silkworm/common/object_pool.hpp
greg7mdp: you would do: `auto uptr = std::move(pool_.back());`, so grab the unique_ptr from the pool.

node/silkworm/db/mdbx.cpp
AndreaLanfranchi: Hmm ... Operator bool on a unique otr checks if there's an associated managed object. And in pooled cursor there always is. What I need instead is to reach the bool operator of the managed object.

node/silkworm/common/object_pool.hpp
AndreaLanfranchi: But why use a vector when there is so nice FIFO stack ?

node/silkworm/db/mdbx.cpp
greg7mdp: Oh, you are right, I read too fast and not noticed that you used `cursor_->` and not `cursor_.`.
So yes of course I think `if (*cursor_)` is much more readable.

node/silkworm/common/object_pool.hpp
greg7mdp: You mean LIFO, right? As far as I know virtually nobody uses std::stack. Why deal with a different class when `std::vector` is a stack when you `push_back` and `pop_back`, and has a better api.

AndreaLanfranchi: Yeah LIFO. Sorry.
Well honestly I think I don't need to bother with contiguous memory. The underlying container is a deque which provides better perf imho especially when the pool gets large.

greg7mdp: > provides better perf imho especially when the pool gets large

It would have to grow quite large to make a difference. I guess it is a matter of preference at this point, and there is nothing wrong with your version :-).

AndreaLanfranchi: Well i thought object pool to be as abstract as possible so you never know how many elements. Anyhow keep pushing your comments: always extremely appreciated

yperbasis:
I'm going to test performance impact on my Google Cloud machine.

mriccobene (commented):
I appreciate the ingenuity of the solution but I wonder if a simpler solution cannot be used.

**We are assuming that reusing a cursor gives us a performance gain** (thanks @yperbasis  for making the measurements, it would be worth comparing also the old solution which did not reuse cursors but created them every time).

With this assumption, we consider that:
- the access layer functions are all flat functions that take the transaction as their first parameter; this suggests from an object oriented point of view to create a Transaction class and make all the functions methods of this class by encapsulating the implementation
- each transaction can only be used in the thread in which it is created

So we can think to pair a `managed_txn` and a `managed_cursor` in the implementation of the class and add all the functions as methods.

So for example the function:

```
Bytes read_header_raw(mdbx::txn& txn, ByteView key) {
    PooledCursor src(txn, db::table::kHeaders);
    //...
}
```

becomes parts of the class:

```
class db::Transaction {
    mdbx::txn_managed txn_;
    mdbx::cursor_managed crs_;
public:	
    Bytes read_header_raw(ByteView key) {
        crs_.bind(txn_, db::open_map(txn_, db::table::kHeaders));
        //...
    }

    // other method here: write_header, read_total_difficulty, ...
}
```

Sure this save cursors only for long standing transactions where many methods are called on a single instance. But this can be an acceptable trade off between simplicity and cost. After all, transactions that are used only once also have a transaction creation cost that could be much higher than that of the cursor creation.

AndreaLanfranchi:
> **We are assuming that reusing a cursor gives us a performance gain** (thanks @yperbasis for making the measurements, it would be worth comparing also the old solution which did not reuse cursors but created them every time).

If I recall correctly @yperbasis has reported that a test on his AWS machine depicted a saving on execution of roughly 40 minutes over a total execution time of about 16 hours.

> becomes parts of the class:
> 
> ```
> class db::Transaction {
>     mdbx::txn_managed txn_;
>     mdbx::cursor_managed crs_;
> public:	
>     Bytes read_header_raw(ByteView key) {
>         crs_.bind(txn_, db::open_map(txn_, db::table::kHeaders));
>         //...
>     }
> 
>     // other method here: write_header, read_total_difficulty, ...
> }
> ```
> 
> Sure this save cursors only for long standing transactions where many methods are called on a single instance. But this can be an acceptable trade off between simplicity and cost. After all, transactions that are used only once also have a transaction creation cost that could be much higher than that of the cursor creation.

While this might be technically and semantically impeccable it'd require a lot of code refactoring which, with all due respect, I'd gladly avoid. Also, and please forgive my stubborness, this somehow hides the dust under the rug which : "do we have two or more transactions live at the same time in one single thread ?". It seems to me the problem you encounter derives from this condition and will be honestly glad to be proved wrong before going through a huge rewriting.

mriccobene:
>  this somehow hides the dust under the rug which means: "do we have two or more transactions live at the same time in one single thread ?". It seems to me the problem you encounter derives from this condition and will be honestly glad to be proved wrong before going through a huge rewriting.

If this is a solution to a problem of the downloader please wait, I'm investigating.

yperbasis:
A couple of files from `interfaces/remote` erroneously made it into changed files.

AndreaLanfranchi:
> A couple of files from `interfaces/remote` erroneously made it into changed files.

Dont' understand on my local builds they keep getting generated

AndreaLanfranchi:
Anyway ... I dont' think this PR will go through as **for sure** will give worse performances compared to actual master.

node/silkworm/db/mdbx.cpp
yperbasis: I concur with Greg: `if (*cursor_)` is more readable.

yperbasis:
I've just tested and it seems that this PR actually improves performance. On my Google cloud test machine I got:
```
cursor-pool commit.37003edf    Execution         done=1d 7h 56m 32s
master      commit.d9e36b80    Execution         done=1d 9h 27m 9s
```
Maybe it's a glitch in the data centre, but at the very least this PR in unlikely to cause performance degradation.

So my take would be to sort out unnecessarily changed generated files and to proceed with this PR.

yperbasis (approved):

AndreaLanfranchi:
Wait ... still don't understand where new proto files come from

yperbasis (approved):

node/silkworm/db/mdbx.cpp
```diff
@@ -160,6 +162,37 @@ ::mdbx::cursor_managed open_cursor(::mdbx::txn& tx, const MapConfig& config) {
...
+
+void PooledCursor::bind(::mdbx::txn& tx, const MapConfig& config) {
+    assert(cursor_);
+    const auto& cm{*cursor_};
+    // Check cursor is bound to a live transaction
+    if (auto cm_tx{mdbx_cursor_txn(&(*cm))}; cm_tx) {
+        // If current transaction id does not match cursor's transaction close it
+        // and recreate a new one
+        if (tx.id() != mdbx_txn_id(cm_tx)) {
+            cursor_.reset(new ::mdbx::cursor_managed());  // RAII implement cursor closure
```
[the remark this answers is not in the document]
AndreaLanfranchi: Well actually is not exactly so ... MDBX transactions track opened cursors and void their handle when the transaction itself is aborted or committed. By consequence is most likely the first check to fail (if (auto cm_tx{mdbx_cursor_txn(&(*cm))}; cm_tx) {` and the cursor is re-usable. Besides having multiple MDBX transactions opened in the same thread is considered bad practice hence, our strive to keep transactions at a minimum, would not require an indexing of cursors by transaction id.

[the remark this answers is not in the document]
greg7mdp: When you declare a PooledCursor in a function, and the function exits, the destructor of PooledCursor moves the cursor to the cache, but the transaction is not aborted or committed, right? 

> Besides having multiple MDBX transactions opened in the same thread is considered bad practice

If it is the case, does it mean that the cursor_pool_ will contain only one cursor?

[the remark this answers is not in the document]
AndreaLanfranchi: > When you declare a PooledCursor in a function, and the function exits, the destructor of PooledCursor moves the cursor to the cache, but the transaction is not aborted or committed, right?

Right

> If it is the case, does it mean that the cursor_pool_ will contain only one cursor?

Not necessarily. Consider this : function A opens a cursor (e.g. to traverse a table) and in its internal loop calls one or more functions (let's call them B) from access layer in a serialized way. At the end of function A there will be two cursors in the cache. At next call of function A all cursors in the cache will be reused.
Bottom line we've reduced to the very bare minimum the amount of `malloc`'s required by cursor creation.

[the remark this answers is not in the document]
AndreaLanfranchi: Obviously all of this implies PooledCursor can be used also outside of access_layer.cpp's scope

greg7mdp:
Sounds good. So typically the number of cursors in the pool will be the max recursion depth, right? I still think that a `vector<unique_ptr<mdbx::cursor_managed>>` would work as well as ObjectPool (which seems like an unnecessary abstraction to me).

I don't really understand how cursors are used, so probably it is not the case, but shouldn't functions A and B be able to use the same cursor? Or maybe there would be no benefit to this. It seems a good idea to avoid allocating cursors every time anyways.

greg7mdp:
Actually, why have a `unique_ptr`? the mdbx cursor is itself a pointer I think `MDBX_cursor *handle_{nullptr}`.
So the pool could simply be `vector<mdbx::cursor_managed>` and you wouldn't have to change all the `operator.()` to `operator->()`.

AndreaLanfranchi:
> Sounds good, thanks dfor explaining @AndreaLanfranchi . So typically the number of cursors in the pool will be the max recursion depth, right? I still think that a `vector<unique_ptr<mdbx::cursor_managed>>` would work as well as ObjectPool (which seems like an unnecessary abstraction to me).
 
Again not necessarily : there are function which require to open two or more cursors so cache size is not necessarily related to recursion depth. 

> I don't really understand how cursors are used, so probably it is not the case, but shouldn't functions A and B be able to use the same cursor? Or maybe there would be no benefit to this. It seems a good idea to avoid allocating cursors every time anyways.

A cursor is a pointer to data in the table: when you reuse the cursor the pointer position is invalidated and the cursor is marked as "not-set" (with an improper API property called EOF) and you have to perform a "seek" on the internal btree to reposition it.. 
Now picture this : function A opens a cursor to orderly read all keys of table (initial set and then all `next` up to the end of data). Within the loop other function(s) are called to retrieve related data ... dependant functions cannot "share" the same cursor as it would invalidate the loop in function A (you'd have to memorize the key you were at and reposition after the return of sub-functions - which is extremely inefficient).

greg7mdp:
Thanks @AndreaLanfranchi for the great explanation. Still we don't need to create `unique_ptr` to cursors, which are already pointers, right?

AndreaLanfranchi:
> Actually, why have a `unique_ptr`? the mdbx cursor is itself a pointer I think `MDBX_cursor *handle_{nullptr}`. So the pool could simply be `vector<mdbx::cursor_managed>` and you wouldn't have to change all the `operator.()` to `operator->()`.

Cause, unless u use some hacks, you can't take over memory owned by a vector. You can get a reference to the memory where the object is stored but you can't move away an element from the vector without invoking the element's dtor.
With unique_ptr instead I'm physically moving the ownership of the object from the cache to the consumer ... and it may not return to the cache.

greg7mdp:
But you could just use `cursor` instead of `cursor_managed`, and then it would work fine (no close on destruction)

AndreaLanfranchi:
> But you could just use `cursor` instead of `cursor_managed`, and then it would work fine (no close on destruction)

Yes that's true ... cursor does not destroy the internal handle but also it does not expose it. So when a cursor is destroyed, by the means of C++ bindings, I also loose the knowledge of where `MDBX_cursor struct` (the handle) is.

greg7mdp:
You could just do something like this in silkworm:

```
class cursor_unmanaged : public mdbx::cursor {
public:
    // no destructor

    void close() { ::mdbx_cursor_close(handle_); handle_ = nullptr; }

};
```

and then you can use all the nice functionality from mdbx::cursor but bypassing the destructor from cursor_managed.
And when destructing the thread_local static holding the `std::vector<cursor_unmanaged>` you can close all the cursors if needed by calling close().

AndreaLanfranchi:
But once again we're back to a vector with unmanageable ownership. Maybe I'm a bit tired but I don't get the point. If you could provide a mock by code example maybe I can se were we try to aim

PR #595, opened 2022-03-01
fputs instead of logger in SignalHandler::handle

yperbasis:
Also this PR temporarily disables consensus tests in linux-gcc-11-thread-sanitizer (deferred to Issue #597).


AndreaLanfranchi (commented):
Should silkworm be executed headless with logging to file we don't have knowledge of signal beign captured

yperbasis:
> Should silkworm be executed headless with logging to file we don't have knowledge of signal beign captured

Right. But typically the user will intentionally send a signal to stop the process. IMHO it's not a big deal if the stopping messages aren't copied to a log file.

AndreaLanfranchi:
> > Should silkworm be executed headless with logging to file we don't have knowledge of signal beign captured
> 
> Right. But typically the user will intentionally send a signal to stop the process. IMHO it's not a big deal if the stopping messages aren't copied to a log file.

Well we've seen in Ubuntu for example that some sort of memory resource protection signals Erigon process to stop.
Don't recall exactly the signal code though.

yperbasis:
In most cases stdout/stderr will be redirected to a file on Linux, so shouldn't be a problem. Frankly I don't know why we have this option of teeing logger output to a file. Seems like an unnecessary feature to me.

AndreaLanfranchi:
> In most cases stdout/stderr will be redirected to a file on Linux, so shouldn't be a problem. Frankly I don't know why we have this option of teeing logger output to a file. Seems like an unnecessary feature to me.

There are several reasons :

1. Redirection teeing does not work well on Windows
2. In case of coloured logging the lines on file are purged of coloring codes
3. On request by Paolo the feature was introduced to provide a "simpler" way to capture logging without forcing the user to know (sometime cryptic) redirection instructions
4. This is the base to implement, if we want, syslogging

AndreaLanfranchi (approved):

PR #598, opened 2022-03-02
Reduce memory reallocation in prefetch_blocks

yperbasis:
This is a small optimization to reduce memory reallocation in `prefetch_block`. 

| master@[e1dd4c37](https://github.com/torquem-ch/silkworm/commit/e1dd4c374e582693b5c99f1c03b0481363acfa41) | prefetch@[871e241e](https://github.com/torquem-ch/silkworm/commit/871e241e97627ba19a759691a8e8b1efb4a002f6) |
| --- | --- | 
|master_e1dd4c37|prefetch_e73be231|

linux-x86_64, gcc 11.2.0


node/silkworm/stagedsync/stage_execution.hpp
```diff
@@ -41,11 +42,14 @@ class Execution final : public IStage {
   private:
     std::unique_ptr<consensus::IEngine> consensus_engine_;
     BlockNum block_num_{0};
+    std::vector<Block> prefetched_blocks_;
```
greg7mdp, line 45: Because of the size of `Block` I wonder if it should not be `std::vector<unique_ptr<Block>>`

yperbasis: The idea is to reuse the same vector of prefetched blocks in the loop in `Execution::execute_batch` so that `db::read_block` uses an already allocated `Block` instance. I'm not sure how an extra level of memory indirection (`std::vector<unique_ptr<Block>>`) could help.

greg7mdp: I was writing this because the prefetch_blocks function is called with `from` and `to` parameters and resizes the vector. I thought that maybe the resize could become slow if the vector becomes very large and prefect_blocks is called often with a small `to` - `from`, but maybe it is not an issue.

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -108,52 +109,27 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
-        }
-        ByteView encoded_header{raw_header.data(), raw_header.length()};
-        rlp::success_or_throw(rlp::decode(encoded_header, block.header));
-
-        if (!db::read_body(*txn, block_key, /*read_senders=*/true, block)) {
-            throw std::runtime_error("Unable to load block body for block " + std::to_string(from));
-        }
-        ret.push(block);
+    const size_t n{std::min(static_cast<size_t>(to - from + 1), max_blocks)};
+    prefetched_blocks_.resize(n);
```
AndreaLanfranchi, line 121: My two cents. Avoiding blocks reallocation is good but it also means more heap memory required which transforms into less available memory for OS to cache MDBX data pages. In my original implementation I've chosen (totally arbitrarily) 1024 blocks which gets progressively deleted as soon as they're consumed hence leaving more memory for state lookups (if any).  Choosing too small pre-fetch segments, imho, partially voids the purpose of prefetching.
Besides the heavy part of reallocations occurs in Block's ommers and transactions retrieval (Block header is relatively small if compared to the other two).

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -108,52 +109,27 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
 
-        if (from == to || ret.size() >= max_blocks) {
-            break;
+    for (size_t i{0}; i < n; ++i) {
+        if (!db::read_block(*txn, from + i, /*read_senders=*/true, prefetched_blocks_[i])) {
+            throw std::runtime_error("Bad canonical header sequence: missing block " + std::to_string(from + i));
```
AndreaLanfranchi, line 125: This imho worsen things. Original implementation uses one cursor to read canonical hashes sequentially (only one lookup on begin of loop and then next key, next key ...) and invokes `db::read_body` by key of block_num + hash which eventually opens 1 cursor and does 1 lookup on : BlockBodies, BlockTransactions
Here instead we force 1 lookup for each block on CanonicalHeaders too

node/silkworm/stagedsync/stage_execution.hpp
AndreaLanfranchi: > I was writing this because the prefetch_blocks function is called with `from` and `to` parameters and resizes the vector. I thought that maybe the resize could become slow if the vector becomes very large and prefect_blocks is called often with a small `to` - `from`, but maybe it is not an issue.

In reality the resizing happens only 2 times (when large segments of blocks gets executed) or even 1 time (likely when at the tip of the chain) cause `prefetch` fills a batch of blocks up to segment width or up to max_blocks (whichever the smaller)

greg7mdp: > In reality the resizing happens only 2 times

Oh great then my comment is irrelevant!

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -108,52 +109,27 @@ StageResult Execution::forward(db::RWTxn& txn) {
     return is_stopping() ? StageResult::kAborted : StageResult::kSuccess;
 }
 
-std::queue<Block> Execution::prefetch_blocks(db::RWTxn& txn, BlockNum from, BlockNum to, size_t max_blocks) {
+void Execution::prefetch_blocks(db::RWTxn& txn, const BlockNum from, const BlockNum to, const size_t max_blocks) {
+    SILKWORM_ASSERT(from <= to);
```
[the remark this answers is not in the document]
yperbasis: OK, I've removed it.

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: Fair enough. I've re-introduced the batch CanonicalHeaders logic in `read_blocks` in `access_layer`.

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: 1. I restored the prefetch size to be 10240, the same as in the master.
2. `Execution::prefetch_blocks` now calls `prefetched_blocks_.clear()` so that all memory occupied by ommers and transactions is cleared.

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -109,52 +109,26 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
-        if (!db::read_body(*txn, block_key, /*read_senders=*/true, block)) {
-            throw std::runtime_error("Unable to load block body for block " + std::to_string(from));
-        }
-        ret.push(block);
-
-        if (from == to || ret.size() >= max_blocks) {
-            break;
-        }
+    const size_t n{std::min(static_cast<size_t>(to - from + 1), max_blocks)};
+    prefetched_blocks_.resize(n);
```
AndreaLanfranchi, line 121: Here we may allocate more blocks than what possibly returned by read_blocks (in case of error). In understand such error would be unrecoverable nevertheless I find this conceptually arguable.

node/silkworm/db/access_layer.cpp
```diff
@@ -190,29 +190,42 @@ void read_transactions(mdbx::cursor& txn_table, uint64_t base_id, uint64_t count
...
+        const BlockNum block_number{from + i};
+        Block& block{out[i]};
 
-    // Read header
-    key = block_key(block_number, bh.hash.bytes);
-    auto raw_header{read_header_raw(txn, key)};
-    if (raw_header.empty()) {
-        return false;
+        key = block_key(block_number);
+        auto data{canonical_hashes_cursor.find(to_slice(key), false)};
```
AndreaLanfranchi, line 206: Why build the key (which implies Bytes allocation and endianess swap) for each block number and force a find (which is slower than movenext) ?
kCanonicalHashes is already sorted by block number ... use a loop with a single find of initial key and then move next, move next till EOF or when count of retrieved elements reaches the limit.
Eventually endian load data key from cursor to check blocks are in order

AndreaLanfranchi (changes requested):

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -175,71 +149,66 @@ StageResult Execution::execute_batch(db::RWTxn& txn, BlockNum max_block_num, Bas
...
+                ++processed_blocks_;
+                processed_transactions_ += block.transactions.size();
+                processed_gas_ += block.header.gas_used;
+                gas_batch_size += block.header.gas_used;
+                gas_history_size += block.header.gas_used;
+                progress_lock.unlock();
+
+                // Flush whole buffer if time to
+                if (gas_batch_size >= gas_max_batch_size || block_num_ >= max_block_num) {
+                    prefetched_blocks_.clear();  // free the memory held by transactions, etc
```
AndreaLanfranchi, line 198: I don't see any usefulness in making `prefetched_blocks_` a private member variable if we end up clearing it on batch commit: eventually it's life cycle is within execute_batch.
A possible optimization is imho this : instead of clearing its contents on batch commit save a pointer to last processed block and, on re-entry of `execute_batch` process residual blocks instead of refetching those from db.
Example : we loaded 10240 blocks but after first 100 the buffer overflow gets triggered. We destroy 10140 loaded blocks and on re-entry we reload those again.

AndreaLanfranchi (changes requested):

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: Fair enough. I've reworked my change and now use a circular buffer with popping out blocks immediately after being processed.

node/silkworm/db/access_layer.cpp
yperbasis: Fair enough, I've switched to `db::cursor_for_count`

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: I've implemented the suggested optimization.

node/silkworm/stagedsync/stage_execution.cpp
```diff
@@ -109,52 +112,47 @@ StageResult Execution::forward(db::RWTxn& txn) {
...
+            ++from;
+            return true;
+        }};
+        num_read = db::cursor_for_count(hashes_table, walk_function, count);
+    }
 
-        ++from;
-        data = hashes_table.to_next(false);
+    if (num_read != count) {
+        throw std::runtime_error("Missing block " + std::to_string(from + num_read));
```
AndreaLanfranchi, line 149: This shows incorrect info: `from` is altered within walk function so the printed value is not coherent.

AndreaLanfranchi (changes requested):

node/silkworm/stagedsync/stage_execution.cpp
yperbasis: Indeed. Fixed in https://github.com/torquem-ch/silkworm/commit/18d9e63ce2200e43143ef959c0223e3b33fa2df9

PR #601, opened 2022-03-04
Complete structure for BackEnd RPC services

canepat:
Add server-streaming RPC support
Complete test client for for BackEnd RPCs

### DIFF
--- a/node/silkworm/rpc/backend_server.cpp
+++ b/node/silkworm/rpc/backend_server.cpp
@@ -16,8 +16,6 @@
 
 #include "backend_server.hpp"
 
-#include <sstream>
-
 #include <silkworm/common/endian.hpp>
 #include <silkworm/common/log.hpp>
 #include <types/types.pb.h>
@@ -102,17 +100,11 @@ void ProtocolVersionService::process_rpc(ProtocolVersionRpc& rpc, const remote::
     SILK_TRACE << "ProtocolVersionService::process_rpc rsp: " << &response_ << " sent: " << sent;
 }
 
-ClientVersionService::ClientVersionService()
+ClientVersionService::ClientVersionService(const ServerConfig& srv_config)
     : ClientVersionRpcService(
         [&](auto& rpc, const auto* request) { process_rpc(rpc, request); },
         &remote::ETHBACKEND::AsyncService::RequestClientVersion) {
-    /*const auto build_info{backend_kv_server_get_buildinfo()};
-    std::stringstream str_stream;
-    str_stream << "silkworm/" << build_info->project_version << "/"
-        << build_info->system_name << "-" << build_info->system_processor << " " << build_info->build_type << "/"
-        << build_info->compiler_id << "" << build_info->compiler_version;*/
-    std::stringstream str_stream{"silkworm/0.0.0"};
-    response_.set_nodename(str_stream.str());
+    response_.set_nodename(srv_config.node_name());
 }
 
 void ClientVersionService::process_rpc(ClientVersionRpc& rpc, const remote::ClientVersionRequest* request) {
@@ -152,7 +144,7 @@ void NodeInfoService::process_rpc(NodeInfoRpc& rpc, const remote::NodesInfoReque
 }
 
 BackEndServer::BackEndServer(const ServerConfig& srv_config, const ChainConfig& chain_config)
-: Server(srv_config), etherbase_service_{chain_config}, net_version_service_{chain_config}, net_peer_count_service_{} {
+: Server(srv_config), etherbase_service_{chain_config}, net_version_service_{chain_config}, client_version_service_{srv_config} {
     SILK_INFO << "BackEndServer created listening on: " << srv_config.address_uri();
 }
 
--- a/node/silkworm/rpc/backend_server.cpp
+++ b/node/silkworm/rpc/backend_server.cpp
@@ -16,6 +16,8 @@
 
 #include "backend_server.hpp"
 
+#include <sstream>
+
 #include <silkworm/common/endian.hpp>
 #include <silkworm/common/log.hpp>
 #include <types/types.pb.h>
@@ -37,7 +39,7 @@ inline static types::H160* new_H160_address(const evmc::address& address) {
     return h160;
 }
 
-void EtherbaseService::process_rpc(EtherbaseUnaryRpc& rpc, const remote::EtherbaseRequest* request) {
+void EtherbaseService::process_rpc(EtherbaseRpc& rpc, const remote::EtherbaseRequest* request) {
     SILK_TRACE << "EtherbaseService::process_rpc START rpc: " << &rpc << " request: " << request;
 
     remote::EtherbaseReply response;
@@ -48,7 +50,7 @@ void EtherbaseService::process_rpc(EtherbaseUnaryRpc& rpc, const remote::Etherba
     SILK_TRACE << "EtherbaseService::process_rpc END rsp: " << &response << " etherbase: " << to_hex(etherbase_) << " sent: " << sent;
 }
 
-void NetVersionService::process_rpc(NetVersionUnaryRpc& rpc, const remote::NetVersionRequest* request) {
+void NetVersionService::process_rpc(NetVersionRpc& rpc, const remote::NetVersionRequest* request) {
     SILK_TRACE << "NetVersionService::process_rpc rpc: " << &rpc << " request: " << request;
 
     remote::NetVersionReply response;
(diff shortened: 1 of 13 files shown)


node/silkworm/rpc/backend_server.cpp
```diff
@@ -58,8 +58,93 @@ void NetVersionService::process_rpc(NetVersionUnaryRpc& rpc, const remote::NetVe
     SILK_TRACE << "NetVersionService::process_rpc rsp: " << &response << " chain_id: " << chain_id_ << " sent: " << sent;
 }
 
+void NetPeerCountService::process_rpc(NetPeerCountRpc& rpc, const remote::NetPeerCountRequest* request) {
+    SILK_TRACE << "NetPeerCountService::process_rpc rpc: " << &rpc << " request: " << request;
```
AndreaLanfranchi, line 62: Why print reference to rpc ? I mean `<< &rpc` ?

AndreaLanfranchi: Also ... do `rpc` and `rsp` stand for `request` and `response` ? 
And again ... out of curiosity why don't write 
```
SILK_TRACE << __FUNCTION__ << " rpc: " << &rpc << " request: " << request;
```
At least, should you change function naming, don't have to rewrite in the log lines

canepat: > Why print reference to rpc ? I mean `<< &rpc` ?

I'm printing the memory address of `rpc` for tracing purposes, it is useful to debug RPC lifetime issues (live RPC objects are allocated and maintained on the heap, then deallocated on cleanup)

mriccobene:
IMO the readability of the class RpcService can be improved:

```
template <
    typename AsyncService,
    typename Request,
    typename Reply,
    template<typename, typename, typename> typename GenericRpc> // <= note the new name here
class RpcService {
    using Rpc = GenericRpc<AsyncService, Request, Reply>;  // <= added
    using RpcServiceHandlers = typename Rpc::Handlers;
    

[...]

    auto add(Rpc* rpc)    // <= note the method new name here
    auto remove(Rpc* rpc)  // <= note the method new name here
```

Also I note you are following the C++ Core Guidelines, where they say [a raw pointer (T*) is non-owning](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r3-a-raw-pointer-a-t-is-non-owning) . I don't like this point in the guidelines very much because it is a convention that not everyone knows and not everyone respects, so the reader of the signature, like me, do not know; moreover the compiler cannot reinforce it. I like where guidelines suggest the use of `owner<T*>` which can be enforced by static analysis tools and make the intent clear, but unfortunately the guidelines don't propose a `non_owner<T*>`. Someone tried to do it in the committee but the proposal seems to me to have been abandoned. I would put at least one custom attribute like `[[non_owning]] ` but in this project they are not allowed. So just keep using it.

node/silkworm/rpc/backend_server.cpp
canepat: > Also ... do `rpc` and `rsp` stand for `request` and `response` ?

`rpc` stands for `remote procedure call`, but you have a point here: naming is too much close between `rpc` and `rsp` in the trace log, I will amend it.

> And again ... out of curiosity why don't write
> 
> ```
> SILK_TRACE << __FUNCTION__ << " rpc: " << &rpc << " request: " << request;
> ```
> 
> At least, should you change function naming, don't have to rewrite in the log lines

Yes, you're right, definitely better. Maybe we can go one step further and include `__FUNCTION__` into a couple of extended `SILK_TRACE` logging macros to avoid the repetition altogether:

```
#define SILK_TRACE_ENTER SILK_TRACE << __FUNCTION__ << " ENTER"
#define SILK_TRACE_EXIT  SILK_TRACE << __FUNCTION__ << " EXIT"
```
or

```
#define SILK_TRACE_START SILK_TRACE << __FUNCTION__ << " START"
#define SILK_TRACE_END   SILK_TRACE << __FUNCTION__ << " END"
```

AndreaLanfranchi: Please don't ... 🙏 
Don't add another macro 😄

AndreaLanfranchi: > `rpc` stands for `remote procedure call`, but you have a point here: naming is too much close between `rpc` and `rsp` in the trace log, I will amend it.

Yes I knew the acronym for rpc ... seemed to me you were using it as request response placeholders

canepat: TBH I'm trying the above suggestion and hit a wall on the class name: I'm realizing that including the class name is really useful to distinguish between `process_rpc` calls in different types, yet AFAICT there is no standard way to obtain demangled class names.

AndreaLanfranchi: Can you build this sample and report the outputs ?

```
$ cat test.cpp 
#include <iostream>

int main(int argc, char **argv)
{
    std::cout << __func__ << std::endl
              << __FUNCTION__ << std::endl
              << __PRETTY_FUNCTION__ << std::endl;
}
$ g++ test.cpp 
$ ./a.out 
```

canepat: > 

I've just tried a slight variation including some classes to reproduce the class name issue:

https://godbolt.org/z/bh9MPrbeM

None of them seems to work well

canepat: > Hmmm ... I don't know on linux but at least on windows `__FUNCTION__` returns a FQDN of the function like in this example
> 
> ```
>  ERROR [03-05|19:15:51.470 UTC] HashState                     function=silkworm::stagedsync::HashState::hash_from_account_changeset exception=kAborted
> ```

The problem here is probably that `__FUNCTION__` is not standard

AndreaLanfranchi: MSVC 1 - GCC 0

greg7mdp: the whole header, or how to get the current function name in any compiler.

```
#ifndef BOOST_CURRENT_FUNCTION_HPP_INCLUDED
#define BOOST_CURRENT_FUNCTION_HPP_INCLUDED

// MS compatible compilers support #pragma once

#if defined(_MSC_VER) && (_MSC_VER >= 1020)
# pragma once
#endif

//
//  boost/current_function.hpp - BOOST_CURRENT_FUNCTION
//
//  Copyright (c) 2002 Peter Dimov and Multi Media Ltd.
//
//  Distributed under the Boost Software License, Version 1.0.
//  See accompanying file LICENSE_1_0.txt or copy at
//  http://www.boost.org/LICENSE_1_0.txt
//
//  [http://www.boost.org/libs/assert/current_function.html](https://www.boost.org/doc/libs/1_63_0/libs/assert/current_function.html)
//

namespace boost
{

namespace detail
{

inline void current_function_helper()
{

#if defined( BOOST_DISABLE_CURRENT_FUNCTION )

# define BOOST_CURRENT_FUNCTION "(unknown)"

#elif defined(__GNUC__) || (defined(__MWERKS__) && (__MWERKS__ >= 0x3000)) || (defined(__ICC) && (__ICC >= 600)) || defined(__ghs__)

# define BOOST_CURRENT_FUNCTION __PRETTY_FUNCTION__

#elif defined(__DMC__) && (__DMC__ >= 0x810)

# define BOOST_CURRENT_FUNCTION __PRETTY_FUNCTION__

#elif defined(__FUNCSIG__)

# define BOOST_CURRENT_FUNCTION __FUNCSIG__

#elif (defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 600)) || (defined(__IBMCPP__) && (__IBMCPP__ >= 500))

# define BOOST_CURRENT_FUNCTION __FUNCTION__

#elif defined(__BORLANDC__) && (__BORLANDC__ >= 0x550)

# define BOOST_CURRENT_FUNCTION __FUNC__

#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)

# define BOOST_CURRENT_FUNCTION __func__

#elif defined(__cplusplus) && (__cplusplus >= 201103)

# define BOOST_CURRENT_FUNCTION __func__

#else

# define BOOST_CURRENT_FUNCTION "(unknown)"

#endif

}

} // namespace detail

} // namespace boost

```
#endif // #ifndef BOOST_CURRENT_FUNCTION_HPP_INCLUDED

canepat: @greg7mdp thanks for your suggestion, but for this use case (including class::method in a log trace) `BOOST_CURRENT_FUNCTION` has the same drawback as `__PRETTY_FUNCTION__` (obviously, because it is the same at least for GCC): too much verbose.

I think that sticking to the hard-coded class::method names in few log traces is not so bad, anyway.

mriccobene (approved):
LGTM (see my comments in the previous post).

canepat:
> IMO the readability of the class RpcService can be improved:
> 
> ```
> template <
>     typename AsyncService,
>     typename Request,
>     typename Reply,
>     template<typename, typename, typename> typename GenericRpc> // <= note the new name here
> class RpcService {
>     using Rpc = GenericRpc<AsyncService, Request, Reply>;  // <= added
>     using RpcServiceHandlers = typename Rpc::Handlers;
>     
> 
> [...]
> 
>     auto add(Rpc* rpc)    // <= note the method new name here
>     auto remove(Rpc* rpc)  // <= note the method new name here
> ```
> 

Sure. Adding a `using` directive for `Rpc` is a good idea.

> Also I note you are following the C++ Core Guidelines, where they say [a raw pointer (T*) is non-owning](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r3-a-raw-pointer-a-t-is-non-owning) . I don't like this point in the guidelines very much because it is a convention that not everyone knows and not everyone respects, so the reader of the signature, like me, do not know; moreover the compiler cannot reinforce it. I would put at least one custom attribute like `[[non_owning]] ` but in this project they are not allowed. But since the methods "add" and "remove" copy the rpc, is it possible to pass it for copy and then move it inside the container?
> 
> Edit: I didn't realize that "add" takes ownership of rpc so I suggest you use unique_ptr<Rpc*> or gsl::owner<Rpc*> in the signature.

TBH I don't know if `std::unique_ptr` fits well here because it could be used only for `add_request` and would require a double move to put the instance into the `requests_` container. On the other hand, I didn't know about `gsl::owner` and its suggested usages: I think that also for `gsl::owner` your reasoning about the need for a convention to be known and respected applies, nevertheless I am happy to use it in this case if it helps understanding the code.

PR #602, opened 2022-03-04
eliminate unique_ptr and cursor indirection

greg7mdp:
@AndreaLanfranchi , this is what I was suggesting yesterday. It avoids the cursor indirection, and we don't need a separate ObjectPool.


AndreaLanfranchi:
Well thank you @greg7mdp ... seems brilliant. Let me review locally.

node/silkworm/db/mdbx.cpp
```diff
@@ -162,36 +162,48 @@ ::mdbx::cursor_managed open_cursor(::mdbx::txn& tx, const MapConfig& config) {
...
+        mdbx::error::throw_exception(MDBX_EINVAL);
+    ::mdbx_cursor_close(handle_);
+    handle_ = nullptr;
+}
+
+void PooledCursor::clear() { 
+    for (auto cursor : cursors_)
+        cursor.close(); 
+    cursors_.clear();
+}
```
AndreaLanfranchi, line 206: I'd not implement this as it makes, semantically, some confusion. Clearing on a cursor means deleting underlying table data. Besides, imho, does not make sense to clear the cache as it's specifically meant to last as long as possible within the thread.

greg7mdp:
@AndreaLanfranchi I just ran the node_test and I see a crash. How can I run access_layer test in Visual Studio.

AndreaLanfranchi:
> @AndreaLanfranchi I just ran the node_test and I see a crash. How can I run access_layer test in Visual Studio.

Is a bit tricky on VS ... any chance you get a copy of Clion ?

Besides I think there's a problem with this implementation :

`static inline thread_local std::vector<PooledCursor> cursors_{};`

when the thread scope terminates the vector will invoke the dtor for every owned object **but** 

```
PooledCursor::~PooledCursor() {
    cursors_.push_back(*this);
}
```
the dtor tries to push back itself in the cache creating a circular dependency

AndreaLanfranchi (changes requested):

node/silkworm/db/mdbx.cpp
AndreaLanfranchi: Is not necessary then. When db closes all opened transactions are aborted, each transaction invalidates its bound cursor(s) - if any - and eventually the thread exits causing the cache to destroy all owned objects (which are no longer bound to any active transaction)

greg7mdp: Even better, removing the clear() API then - it was not even called.

AndreaLanfranchi: As said the ObjectPool was thought with abstraction in mind. Could be used for other purposes too so there it make sense

greg7mdp:
> the dtor tries to push back itself in the cache creating a circular dependency

Thanks!! implementing a fix for this!

greg7mdp:
> the dtor tries to push back itself in the cache creating a circular dependency

Hi @AndreaLanfranchi thanks a lot for finding the issue, I just pushed a fix.

node/silkworm/db/mdbx.cpp
```diff
@@ -162,36 +162,43 @@ ::mdbx::cursor_managed open_cursor(::mdbx::txn& tx, const MapConfig& config) {
...
-    cursor_ = cursors_pool_.acquire();
-    if (!cursor_) {
-        cursor_ = std::make_unique<::mdbx::cursor_managed>();
-    }
+    if (cursors_.empty()) {
+        handle_ = ::mdbx_cursor_create(nullptr);  // create new cursor
+    } else {
+        handle_ = cursors_.back().handle_;  // steal the handle of previously used cursor
+        cursors_.pop_back();
+    }
```
AndreaLanfranchi, line 172: This has a problem:
1) On very first creation of a PooledCursor we create a new one.
2) When that pooled cursor gets destructed its handle is returned to the cache

Now picture this: a new function requires two PooledCursors

1) First one (A) is created by copying handle from the cache
2) pop_back() invokes the dtor of the cached object **but** the object dtor repushes itself in the cache
3) At this stage we have one PooledCursor plus itself in the cache
4) A second PooledCursor (B) is created ... this gets also copied from the cache and eventually bound to a new map
5) Now PooledCursor A is also bound to the same map of PooledCursor B (the share the same ptr to handle) and its position invalidated but in higher code we don't know it.

AndreaLanfranchi (changes requested):

node/silkworm/db/mdbx.cpp
greg7mdp: Thank you thank you! I was having a hard time finding the issue. This should be easily fixed.

greg7mdp: Thanks again @AndreaLanfranchi , I just checked in the fix and I believe the code is safe now. The node_test pass. Basically we want to make sure we always move (not copy) the PooledCursor, so the ownership of the internal pointer remains unique.

AndreaLanfranchi: At family dinner now ... mind if I review tomorrow ?

greg7mdp: No problem of course, enjoy your family dinner. Italian food is so delicious!

AndreaLanfranchi:
@greg7mdp I took the liberty to cherry pick your work and integrate it in branch `cursors-pool-2` (Hope you don't mind just wanted to have some fun).
Rationale of my mods:
- we've established an external object pool is de-facto inevitable to avoid renter the cache (which is actually what `CursorPool` is)
- I refactored `ObjectPool` to keep abstraction : using unique_ptr with custom deleter. By consequence deleted `CursorPool`
- Realized we don't actually need a cache of `mdbx::cursor` rather a cache of `handle_` (aka `MDBX_cursor`)

Really eager to hear your thoughts about it.

greg7mdp:
Hi @AndreaLanfranchi , I don't mind at all you integrating my changes into your branch.

First, I see an problem where we allocate a cursor with `::mdbx_cursor_create(nullptr)` (memory allocation) every time PooledCursor is created. even if we end up using a cursor from the pool.

Also we still create unneeded unique_ptr, which point to another pointer (handle_). 

And finally, while I am all in favor of abstracting common patterns, in this case the end result seems more cryptic to me, and requires careful reading of both PooledCursor and ObjectPool to understand the code. I think abstractions are good when they are simple and easily understandable. In this case the ObjectPool functionality just simplifies the following 6 lines into one:

```
    if (cursors_.empty()) {
        handle_ = ::mdbx_cursor_create(nullptr);  // create new cursor
    } else {
        std::swap(handle_, cursors_.back().handle_);  // steal the handle of previously used cursor
        cursors_.pop_back();
    } 
```

Which doesn't seem worth it to me, but it is a matter of preference. Also, instead of creating a new ObjectPool abstraction, why not reuse lru_cache?

In any case thanks for considering my change.

AndreaLanfranchi:
@greg7mdp as a matter of fact I agree aiming to abstraction inevitably complicate reading.
Besides I still have segfaults on thread termination: I reiterate ,,, had to research for some hours in the weekend instead of focusing on same things of all-week schedule.

> Also we still create unneeded unique_ptr, which point to another pointer (handle_).

My point is that unique_ptr can carry a custom deleter which (arguably) simplifies the destruction of the pool by RAII without the need to manually close each cached object before the cache itself gets destroyed.

I will play around a little bit more (is Saturday) and willing to hear your comments.
This PR stays in place and would like an opinion also from @yperbasis and @canepat before merging.

Thank you for your passionate support and kind understanding.

AndreaLanfranchi:
>  Also, instead of creating a new ObjectPool abstraction, why not reuse lru_cache?

LRU cache implements a logic which is actually not needed here (a map of iterators towards a list being spliced for every addition). We really don't care to change the order of cached items when the explicit purpose is a FIFO container.

greg7mdp:
> > Also, instead of creating a new ObjectPool abstraction, why not reuse lru_cache?
> 
> LRU cache implements a logic which is actually not needed here (a map of iterators towards a list being spliced for every addition). We really don't care to change the order of cached items when the explicit purpose is a FIFO container.

Hum, I'm not sure. Your ObjectCache has a max_size and stops inserting when the size is reached. I was thinking it might be better to insert the new ojject (which has a higher chance of being re-used) and expell the lru one.
Maybe it is not necessary for cursors, but then it means that ObjectPool is dedicated to cursors and not generic.

greg7mdp:
> I will play around a little bit more (is Saturday) and willing to hear your comments.

Well, you guys can decide. What I like about my version is that it is the minimal and simplest code to support this specific use case. But your version would be fine too. No matter what it was a fun experiment!

AndreaLanfranchi:
> Your ObjectCache has a max_size and stops inserting when the size is reached. I was thinking it might be better to insert the new ojject (which has a higher chance of being re-used) and expell the lru one.

Yes but is optional. Constructor with no limit is also possible. But rethininking actually setting a limit makes no sense in general : is duty of higher code we don't fill the cache with a humongous number of items.

greg7mdp:
> I agree aiming to abstraction inevitably complicate reading.

Yes, that's why abstractions are great when they are used a lot. For example everybody knows what `std::vector` does, so there is no cost to using it, it just makes the code simpler.

canepat:
@greg7mdp @AndreaLanfranchi I really appreciate the language expertise and the fruitful collaboration needed to design such an ingenious solution. TBH I need to play with the resulting abstractions a bit to give some more structured feedback, if any: I guess I will have some chance soon in KV server (it needs to expose MDBX cursors remotely).

After a quick look, I am wondering if the need for subclassing `mdbx::cursor` in order to access and use the MDBX C API (namely `mdbx::cursor::handle_`) could indicate that a better place for this PR is within MDBX C++ API itself. I think this could be the case for a some reasons:
- using the MDBX C API directly creates a dependency which could be hidden if these classes were into MDBX C++ API
- the MDBX maintainer could be interested to integrate into MDBX C++ API a client-driven use case as a pool of cursors
- the MDBX maintainer could give you his point of view in terms of both design and implementation details

greg7mdp:
@canepat thank you for looking into this PR, and for the thoughtful comments. Moving this code into mdbx.h++ would not be a bad idea (maybe rename the class pooled_cursor), as it is a generic optimization. I'm not as convinced that silkworm would significantly gain in lack of dependency though :-)

PR #603, opened 2022-03-06
Fix bad access to ok in CompletionRunner::run

yperbasis:
Not sure why the sanitizers in CI didn't catch it, but it was causing bad access when running `node_test` on my MacBook.


AndreaLanfranchi (approved):

canepat:
> Not sure why the sanitizers in CI didn't catch it, but it was causing bad access when running `node_test` on my MacBook.

Oh, thanks, my bad.

I would like our sanitizers in CI to catch this, but we currently don't have UBSAN enabled. So I've just tried to enabled it locally using GCC, but it seems not to report this at least on my setup (Ubuntu 20.04 gcc 11.1.0). Moreover, using Clang UBSAN doesn't work for us due to [int128 multiply issue](https://github.com/llvm/llvm-project/issues/16778), so we're a bit out of luck here.

PR #605, opened 2022-03-07
Cursors pool 2

AndreaLanfranchi:
This is an alternative proposal to #602 
It aims at the very same goal (remove cursor indirection) but adds a layer of abstraction by `ObjectPool`
Actually #602 looks simpler but nevertheless its `vector<::mdbx::cursor>` is de-facto a a vector of unique_ptrs to `handle_`


greg7mdp:
> Actually #602 looks simpler

Actually #602 **is** simpler 😺 .

AndreaLanfranchi:
> > Actually #602 looks simpler
> 
> Actually #602 **is** simpler 😺 .

Definitely can't disagree

yperbasis (approved):
Choosing between this PR and PR #602, I'd prefer this one because then we'll be able to use the ObjectPool abstraction for ExecutionStatePool.

canepat:
I know it is _just naming_, but why `Cursor` instead of `PooledCursor`? I prefer the latter because it reminds me what's going on under the hood.

AndreaLanfranchi:
Actually is not simply `Cursor` rather `db::Cursor` which differs from `mdbx::Cursor`
I decided to rename due to :
- the name is shorter
- the prefix `Pooled` implies there's another `not-pooled` cursor elsewhere which is not (if we take aside `mdbx::cursor` itself).
- the `db::Cursor` is a wrapper around `mdbx::cursor` exactly as `mdbx::cursor_managed` with the only difference that `handle_` is not destroyed on wrapper destruction rather is returned to a pool of reusable handles;
- imho it should become ubiquitous as honestly I see no point in conitnue using `mdbx::cursor_managed` due to its continuous reallocations. Hence `db::Cursor` should be our standard.

canepat:
That's sounds fair enough :+1:

PR #606, opened 2022-03-08
New implementation for pre-verified hashes

mriccobene:
This PR changes the implementation of pre-verified hashes to allow fast compilation.


node/silkworm/downloader/internals/preverified_hashes.hpp
```diff
@@ -41,25 +41,23 @@ namespace silkworm {
...
     static PreverifiedHashes none;  // A void set of hashes that can be used to turn-off pre-verified hashes usage and
                                     // that is useful for  (the default construction of) classes that use this
                                     // functionality optionally
 
-    static PreverifiedHashes mainnet;  // The mainnet set of pre-verified hashes
+    static const PreverifiedHashes& load(uint64_t chain_id); // Load a set of pre-verified hashes from low level impl
 
-    static std::map<uint64_t, const PreverifiedHashes&> per_chain;  // chain-id based access to the instances
+  private:
+    static std::map<uint64_t, PreverifiedHashes> per_chain;  // chain-id based access to the instances
```
AndreaLanfranchi, line 58: I don't get the point of this. Why a map is needed ?
Afaik there is no way we can use preverified hashes from two (or more) different chains

node/silkworm/downloader/internals/preverified_hashes.hpp
```diff
@@ -41,25 +41,23 @@ namespace silkworm {
...
  */
+
 struct PreverifiedHashes {
     std::set<evmc::bytes32> hashes;  // Set of hashes of headers that are known to belong to canonical chain
-    uint64_t height{0};              // Block height corresponding to the highest preverified header
+    uint64_t height{0};              // Block height corresponding to the highest pre-verified header
 
     [[nodiscard]] bool contains(const evmc::bytes32& hash) const { return hashes.find(hash) != hashes.end(); }
 
     static PreverifiedHashes none;  // A void set of hashes that can be used to turn-off pre-verified hashes usage and
```
AndreaLanfranchi, line 51: Why define a static empty set of itself ?

node/silkworm/downloader/internals/preverified_hashes.hpp
AndreaLanfranchi: On a wider perspective I don't get the point of the whole PreverifiedHashes struct.
Why not simply a pair<BlockNum,std::set<evmc::bytes32>> as const member of WorkingChain which is the only object using it ? Basically preverified hashes should be treated pretty much as ConsensusEngine: once is loaded it stays the same for all the duration of Silkworm instance.
Unless there are specific reasons I'd get rid of this struct

node/silkworm/downloader/internals/preverified_hashes.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    if (chain_id == 1 && result.height == 0) {
+        load_preverified_hashes(result,
+                                preverified_hashes_mainnet_data,
+                                sizeof_preverified_hashes_mainnet_data,
+                                preverified_hashes_mainnet_height);
+    }
+
+    return result;
+}
```
AndreaLanfranchi, line 57: This does not make much sense to me:
- `result.height == 0` seem to me a guard for chain_id presence in `per_chain` map **and** a `not-loaded` flag
- Hashes should be loaded at struct construction and never reloaded again imho

node/silkworm/downloader/internals/preverified_hashes.hpp
mriccobene: The PreverifiedHashes struct has 2 fields as std::pair<>, yes, but the first makes the code clearer:
- `preverified_hashes_->height` is more clear than `preverified_hashes_.first`
- `preverified_hashes_->contains(...)` is more clear than `preverified_hashes_.second.find(...) ...`

node/silkworm/downloader/internals/preverified_hashes.cpp
mriccobene: The actual code is flexible enough to let us to decide in the future to load or not pre-verified hashes using a command line switch. Also the code make tests simple. 

The code in the function is very simple IMO anyway I propose you to focus on the toolbox to find a way to insert the chain_id in the generated files so that here the code becomes generic respect to the chain.

node/silkworm/downloader/internals/preverified_hashes.hpp
mriccobene: The WorkingChain must be able to run without pre-verified hashes. So by default it uses PreverifiedHashes::none.

node/silkworm/downloader/internals/preverified_hashes.hpp
mriccobene: Regarding the map, I managed to remove it maintaining the possibility to use or not pre-verified hashes (e.g. with a command line switch).

node/silkworm/downloader/internals/preverified_hashes.cpp
mriccobene: This was also simplified in the last  commit.

canepat (approved):

PR #610, opened 2022-03-09
Add KV RPC services

canepat:
Add bidirectional-streaming RPC support
Refactor RPC Server to handle many async services
Add structure for KV RPC services
Add KV RPC empty implementations
Refactor ServerContextPool to separate start from join


mriccobene (commented):
You should consider renaming silkworm::rpc::RpcService: the current naming causes confusion with gRPC Service, which is a well-known concept in the gRPC model, reflected also in generated code, i.e. Service and AsyncService generated classes.

canepat:
Yes, you're right. Considering that the Service classes act mainly as factories of RPC calls, I'm going to rename it `silkworm::rpc::Factory` and all the hierarchy as well.

mriccobene (approved):
LGTM

PR #614, opened 2022-03-15
Add mergeForkBlock

enriavil1:
`mergeForkBlock`  corresponds to `FORK_NEXT_VALUE` of [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675#network).


yperbasis:
Silkworm uses uint64 for block numbers. Why do you want to change it to uint256?

AndreaLanfranchi:
> > Silkworm uses uint64 for block numbers. Why do you want to change it to uint256?
> 
> To match Erigon using big int in terminalBlockNumber

And why Erigon uses such a huge var type ?
Isn't uint64 enough ?

yperbasis (approved):

PR #615, opened 2022-03-16
Switch to intx 0.8.0 endian facilities

yperbasis:
Switch to the new endian facilities introduced in release [0.8.0](https://github.com/chfast/intx/releases/tag/v0.8.0) of intx.

### DIFF
--- a/core/silkworm/common/endian.hpp
+++ b/core/silkworm/common/endian.hpp
@@ -59,107 +59,38 @@ and to SILKWORM_LITTLE_ENDIAN for little-endian ones (most current architectures
 
 namespace silkworm::endian {
 
-#if SILKWORM_BYTE_ORDER == SILKWORM_LITTLE_ENDIAN
-namespace le {
-    template <class UnsignedInteger>
-    static inline UnsignedInteger load(UnsignedInteger value) noexcept {
-        return value;
-    }
-}  // namespace le
-namespace be {
-    static inline uint16_t load(uint16_t value) noexcept { return intx::bswap(value); }
-    static inline uint32_t load(uint32_t value) noexcept { return intx::bswap(value); }
-    static inline uint64_t load(uint64_t value) noexcept { return intx::bswap(value); }
-
-    template <unsigned N>
-    static inline intx::uint<N> load(const intx::uint<N>& value) noexcept {
-        return intx::bswap(value);
-    }
-}  // namespace be
-
-#elif SILKWORM_BYTE_ORDER == SILKWORM_BIG_ENDIAN
-namespace le {
-    static inline uint16_t load(uint16_t value) noexcept { return intx::bswap(value); }
-    static inline uint32_t load(uint32_t value) noexcept { return intx::bswap(value); }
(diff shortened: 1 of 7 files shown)


core/silkworm/common/endian.hpp
```diff
@@ -59,107 +32,38 @@ and to SILKWORM_LITTLE_ENDIAN for little-endian ones (most current architectures
...
-#error "byte order not supported"
-#endif
-
 // Similar to boost::endian::load_big_u16
-inline uint16_t load_big_u16(const uint8_t* bytes) noexcept {
-    uint16_t x;
-    std::memcpy(&x, bytes, sizeof(x));
-    return be::load(x);
-}
+inline uint16_t load_big_u16(const uint8_t* bytes) noexcept { return intx::be::unsafe::load<uint16_t>(bytes); }
```
AndreaLanfranchi, line 36: Just wondering if it's worth to _simply_ make aliases of intx's functions (which is apparently what we're doing here).
Less boilerplate code and more clear intention imho

Something like :
```
const auto load_big_u16 = intx::be::unsafe::load<uint16_t>;
const auto load_big_u32 = intx::be::unsafe::load<uint32_t>;
const auto load_big_u64 = intx::be::unsafe::load<uint64_t>;
const auto load_little_u16 = intx::le::unsafe::load<uint16_t>;
const auto load_little_u32 = intx::le::unsafe::load<uint32_t>;
const auto load_little_u64 = intx::le::unsafe::load<uint64_t>;
const auto store_big_u16 = intx::be::unsafe::store<uint16_t>;
const auto store_big_u32 = intx::be::unsafe::store<uint32_t>;
const auto store_big_u64 = intx::be::unsafe::store<uint64_t>;
```

yperbasis: Nice! Done in https://github.com/torquem-ch/silkworm/commit/fbf499efc77c2f6ff054a4198132b7e1d50128da.

AndreaLanfranchi: Or even
```
template <typename T, typename = std::enable_if_t<std::is_integral_v<T> && std::is_unsigned_v<T>>>
const auto load_big = intx::be::unsafe::load<T>;
```
and then use it as 
```
const auto state_mask{endian::load_big<uint16_t>(v.data()) };
```

I don't think `load_big_u16` vs `load_big<uint16_t>` makes a big difference in typed keys.

yperbasis: I think then it makes sense to get rid of the aliases completely and only use the intx functions directly. I can do that.

AndreaLanfranchi: > I think then it makes sense to get rid of the aliases completely and only use the intx functions directly. I can do that.

I personally wouldn't do that. as in namespace `endian` we already have `to/from_big_compact` probably keeping aliases makes sense so we don't have to change namespaces. But is a matter of taste

yperbasis: OK, then to my mind the current version with the aliases (same as the boost names, e.g. `load_big_u32`) is fine.

AndreaLanfranchi (approved):

PR #626, opened 2022-04-01
Add initial KV Tx RPC implementation

canepat:
This PR introduces the initial implementation of KV Tx bidirectional RPC. Key points:
- every new Tx incoming RPC creates a new MDBX read-only transaction
- one unsolicited message containing the txn ID is notified to the client
- cursor open/close and tracking by pairing `db::Cursor` with table name to support reopening
- one max txn TTL timer protects against long-lived txn by aborting/reopening the txn on behalf of the client, opened cursors are saved/restored transparently
- another idle timer protects against clients which don't send any (more) requests and eventually terminates the RPC

TODOLIST:
- check errors when limits are reached (max number of txns, max number of cursors per txn), if applicable

### DIFF
--- a/node/silkworm/rpc/call.hpp
+++ b/node/silkworm/rpc/call.hpp
@@ -642,6 +642,13 @@ class BidirectionalStreamingRpc : public BaseRpc {
             } else {
                 // TODO(canepat) test this path
                 SILK_ERROR << "BidirectionalStreamingRpc::process_write peer " << peer() << " ok: false [" << this << "]";
+                client_streaming_done_ = true;
+                end();
+
+                // Stop the idle guard timer.
+                idle_timer_.cancel();
+
+                close();
             }
         }
         SILK_TRACE << "BidirectionalStreamingRpc::process_write END [" << this << "]";
--- a/node/silkworm/rpc/call.hpp
+++ b/node/silkworm/rpc/call.hpp
@@ -125,6 +125,7 @@ class BaseRpc {
     /// Returns if one in-progress write operation exists or not.
     bool write_in_progress() const { return write_in_progress_; }
 
+    //! The single-threaded scheduler used to process this RPC asynchronously.
     boost::asio::io_context& scheduler_;
 
     //! Used to access the options and current status of the RPC.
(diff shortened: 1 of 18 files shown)


node/silkworm/rpc/kv_calls_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    SECTION("patch not relevant") {
+        KvVersion lhs1{2, 5, 0};
+        KvVersion rhs1{2, 5, 0};
+        CHECK(higher_version(lhs1, rhs1) == lhs1);
+        KvVersion lhs2{2, 5, 1};
+        KvVersion rhs2{2, 5, 0};
+        CHECK(higher_version(lhs2, rhs2) == lhs2);
+        KvVersion lhs3{2, 5, 0};
+        KvVersion rhs3{2, 5, 1};
+        CHECK(higher_version(lhs3, rhs3) == lhs3);
```
mriccobene, line 57: According to the section title ("patch not relevant") here you are returning the first argument, ok.
So a more convoluted name for higher_version() would be left_biased_higher_version_ignoring_patch(). But I am not proposing this one :-)

node/silkworm/rpc/call.hpp
```diff
@@ -479,12 +489,16 @@ class BidirectionalStreamingRpc : public BaseRpc {
         SILK_TRACE << "BidirectionalStreamingRpc::BidirectionalStreamingRpc END new request issued [" << this << "]";
     }
 
+    virtual void start() = 0;
```
mriccobene, line 492: Is this a method pertaining only to BidirectionalStreamingRpc and to its descendants or to all Rpc? The same for end() and process().

node/silkworm/rpc/call.hpp
```diff
@@ -479,12 +489,16 @@ class BidirectionalStreamingRpc : public BaseRpc {
...
     //! Hook called when a new incoming request from some client has come in for this RPC.
     /// For client-streaming and bidirectional streaming RPCs, a request from client can come in multiple times so
     /// \ref process() may be called repeatedly.
     /// @param request the incoming request or nullptr to indicate end-of-stream for client stream
     virtual void process(const Request* request) = 0;
 
+    virtual void end() = 0;
+
     bool send_response(const Response& response) {
         response_queue_.push_back(std::move(response));
```
mriccobene, line 503: move() here is an oversight or you need to pass Response by copy.

mriccobene (approved):

node/silkworm/rpc/kv_calls_test.cpp
canepat: OK, you have a point here :smile:

Maybe `higher_version_ignoring_patch` is not so convoluted but better than the current name.

node/silkworm/rpc/call.hpp
canepat: The `start/end` methods are meaningful only for bidirectional-streaming calls (and probably client-streaming ones, but we don't need to support them ATM) because they can have more than one incoming request from the same client, hence it makes sense to have hooks for call start, for each request processing and for call end.

The `process` method is indeed present in all RPC types we support (unary, server-streaming, bidirectional-streaming) but the call sites for such an hook is different in each RPC type, i.e. there is no place in code where you have a pointer or reference to `BaseRpc` and you need to upcall `process`. This is the reason it is not defined at the base class level.

node/silkworm/rpc/call.hpp
canepat: Oops, you are right, thanks for reporting.

The concrete RPC classes call the `send_response` method typically in two different use cases:
- either passing a reference to a state variable having the same lifetime as the RPC instance
- or passing a stack-allocated `Response` object after it has been filled with data

So a wise solution here could be introduce overloading and:
- getting rid of the `std::move` call in this version taking `const Response&`
- adding another version taking `Response&&` as formal parameter

PR #628, opened 2022-04-03
[WIP] Extract precompiles into standalone library

yperbasis:


canepat (approved):

chfast:
Hi, I actually started playing with the C API today. I should be able to share some results maybe tomorrow. WIP: https://github.com/torquem-ch/silkpre/compare/capi.

I try to separate core functionality from gas cost so this can be used without heap allocated buffer. The nice C++ API could be build on top of it with gas cost so this becomes hourglass pattern.

Don't you have issue with diamond dependency of secp256k1?

yperbasis:
> Don't you have issue with diamond dependency of secp256k1?

Silkworm depends on silkpre, which in turns depends on secp256k1, so no diamond dependency. I've removed secp256k1 submodule from Silkworm.

PR #632, opened 2022-04-13
Remove unnecessary early exit in EVM::call

yperbasis:
Because after [EIP-161](https://eips.ethereum.org/EIPS/eip-161) it's impossible within EVM to differentiate between empty and non-existent accounts, the early exit in `EVM::call` is unnecessary. Its purpose was to cater for this rule of EIP-161: "No account may change state from non-existent to existent-but-_empty_. If an operation would do this, the account SHALL instead remain non-existent."

Without the early exit some empty accounts might be created, but then they will be deleted at transaction's end as empty touched accounts.

### DIFF
--- a/core/silkworm/execution/evm.cpp
+++ b/core/silkworm/execution/evm.cpp
@@ -196,14 +196,6 @@ evmc::result EVM::call(const evmc_message& message) noexcept {
         return res;
     }
 
-    const bool precompiled{is_precompiled(message.code_address)};
-    const evmc_revision rev{revision()};
-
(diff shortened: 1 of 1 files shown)


chfast:
> Without the early exit some empty accounts might be created, but then they will be deleted at transaction's end as empty touched accounts.

I understand this depends on newly created accounts being marked as "touched".

yperbasis:
> I understand this depends on newly created accounts being marked as "touched".

Yes, that's right.

AndreaLanfranchi (approved):

PR #642, opened 2022-04-20
Skip test causing gRPC data race in sanitizer build

canepat:
This PR skips one specific test in TSAN build because such test triggers some data race within the gRPC library in a very specific scenario.

The "BackEndKvServer E2E: trigger server-side write error" test is purposely written to cause write-errors on Tx RPC server-side implementation by using the bidirectional-streaming API in an unorthodox way, i.e. not reading responses after sending requests and making the Tx call end abruptly on client-side by `grpc::ClientContext` destruction.

Repeating this pattern many times triggers write-errors on Tx RPC server-side implementation and allows to check that server-side lifecycle management for Tx calls is OK but seems to result also in a frequent data race detection during sanitizer builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ jobs:
       CLANG_COVERAGE: OFF
       SANITIZE: OFF
     machine:
-      image: ubuntu-2004:202201-02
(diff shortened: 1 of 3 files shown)


mriccobene (approved):
ok

PR #644, opened 2022-04-21
Clear from field from given transaction when reading them from the DB

peroket:


node/silkworm/db/access_layer.cpp
```diff
@@ -185,7 +185,10 @@ void read_transactions(mdbx::cursor& txn_table, uint64_t base_id, uint64_t count
     for (auto data{txn_table.find(to_slice(key), false)}; data.done && i < count;
          data = txn_table.to_next(/*throw_notfound = */ false), ++i) {
         ByteView data_view{from_slice(data.value)};
-        rlp::success_or_throw(rlp::decode(data_view, v.at(i)));
+
+        Transaction& transaction = v.at(i);
+        transaction.from.reset();  // rlp decode is not touching this field, so make sure it is empty
+        rlp::success_or_throw(rlp::decode(data_view, transaction));
```
AndreaLanfranchi, line 191: I find a more natural place to reset the from field in rlp::decode.
That way we can be sure data initialization is complete and atomic instead of having to remember we may/may not need to reset.

AndreaLanfranchi (changes requested):

node/silkworm/db/access_layer.cpp
peroket: You are right, changed it

AndreaLanfranchi (approved):

PR #645, opened 2022-04-22
Add max cursors per txn check in KV Tx RPC

canepat:
This PR introduces a sanity check on max number of opened cursors per txn in KV Tx RPC.

### DIFF
--- a/node/silkworm/rpc/kv_calls.hpp
+++ b/node/silkworm/rpc/kv_calls.hpp
@@ -52,6 +52,9 @@ constexpr auto kKvApiVersion = KvVersion{5, 1, 0};
 //! The max life duration for MDBX transactions (long-lived transactions are discouraged).
 constexpr boost::posix_time::milliseconds kMaxTxDuration{60'000};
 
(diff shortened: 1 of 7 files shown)


node/silkworm/rpc/kv_calls.hpp
```diff
@@ -147,6 +151,7 @@ class TxCall : public BidirectionalStreamingRpc<remote::KV::AsyncService, remote
     static mdbx::env* chaindata_env_;
     static boost::posix_time::milliseconds max_ttl_duration_;
     static uint32_t next_cursor_id_;
+    static uint32_t max_cursor_id_;
```
mriccobene, line 154: Does the cursor-id must be unique per connection or across all connections?

canepat: Good point. I've checked this on Erigon again: cursor ID is unique and restarts from 0 (zero) for each new Tx call, so the check on ID wrapping is not needed because there is the max cursors per txn check in place.

I'm going to amend the PR. Thanks for spotting this.

mriccobene (approved):

PR #650, opened 2022-04-28
Update chain config for mainnet

mriccobene:
The downloader requests (via set_status rpc) the sentry to send a Status message to peers (see [here](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#status-0x00)) This message carries the "Fork identifier for chain compatibility checks" as stated here https://eips.ethereum.org/EIPS/eip-2124. 

This PR updates the ChainConfig for Mainnet adding Arrow Glacier block number.

### DIFF
--- a/core/silkworm/chain/config.hpp
+++ b/core/silkworm/chain/config.hpp
@@ -136,6 +136,7 @@ inline constexpr ChainConfig kMainnetConfig{
         9'069'000,   // Istanbul
         12'244'000,  // Berlin
         12'965'000,  // London
+        13'773'000,  // Arrow Glacier
     },
 
     1'920'000,   // dao_block


yperbasis:
`ChainConfig` in Silkworm is a bit different from Erigon. `fork_blocks` only comprise blocks for which an `evmc_revision` exists. There is no `evmc_revision` for Arrow Glacier, thus there's a separate field in `ChainConfig` called `arrow_glacier_block`. `kMainnetConfig` has already it set correctly.

mriccobene:
> `ChainConfig` in Silkworm is a bit different from Erigon. `fork_blocks` only comprise blocks for which an `evmc_revision` exists. There is no `evmc_revision` for Arrow Glacier, thus there's a separate field in `ChainConfig` called `arrow_glacier_block`. `kMainnetConfig` has already it set correctly.

Thanks! I had not seen it and I had interpreted the sentry's difficulty in finding peers as due to that lack

PR #660, opened 2022-05-09
Allow to save transaction senders to disk

peroket:


peroket:
Co-authored with @pat-comity https://github.com/pat-comity

node/silkworm/db/access_layer.cpp
```diff
@@ -297,6 +307,15 @@ void parse_senders(mdbx::txn& txn, const Bytes& key, std::vector<Transaction>& o
...
 }
 
+void write_senders(mdbx::txn& txn, const Bytes& key, const std::vector<Transaction>& transactions) {
+    std::vector<evmc::address> buffer;
+    for (const auto& transaction : transactions) {
+        SILKWORM_ASSERT(transaction.from);
+        buffer.push_back(*transaction.from);
+    }
+    write_senders_raw(txn, key, buffer);
+}
```
AndreaLanfranchi, line 317: Don't use a vector here. Use `Bytes` which is more meaningful for downstream call `write_senders_raw`. Vector might be reallocated during loop.
Also use an early exit check.

```suggestion
void write_senders(mdbx::txn& txn, const Bytes& key, const std::vector<Transaction>& transactions) {
    if(transactions.empty()) {
        return;
    }
    Bytes data(transactions.size() * kAddressLength, '\0');
    size_t data_offset{0};
    for(const auto& transaction:transactions) {
        SILKWORM_ASSERT(transaction.from.has_value());
        std::memcpy(&data[data_offset], transaction.from.value().bytes, kAddressLength);
        data_offset += kAddressLength;
    }
    write_senders_raw(txn, key, data);
}```

File: node/silkworm/db/access_layer_test.cpp [hunk 2]
```diff
@@ -477,6 +477,23 @@ namespace db {
...
             CHECK(block.transactions[1].from == 0x941591b6ca8e8dd05c69efdec02b77c72dac1496_address);
         }
+
+        SECTION("read_block_body_and_recover_senders_from_db") {
+            Block block;
+            block.transactions.resize(2);
+
+            // Verify that we can write and recover senders from the db cache
+            evmc::address one = 0x00000000dc17e0aadc383d2db43b0a0d3e029c4c_address;
+            evmc::address two = 0x00000000ca8e8dd05c69efdec02b77c72dac1496_address;
```
AndreaLanfranchi, line 487: This is wrong and should not be doable (thank you for bringing to my attention).
You cannot set an arbitrary sender to a transaction **because we need to ensure the sender address properly matches the one recovered from the signature of the transaction**
Due to this we must change `from` field to a getter of a private member
Eventually the `write_senders` could become 
```
void write_senders(mdbx::txn& txn, const Bytes& key, std::vector<Transaction>& transactions) {
    if(transactions.empty()) {
        return;
    }
    Bytes data(transactions.size() * kAddressLength, '\0');
    size_t data_offset{0};
    for(auto& transaction:transactions) {
        transaction.recover_sender();
        std::memcpy(&data[data_offset], transaction.from.value().bytes, kAddressLength);
        data_offset += kAddressLength;
    }
    write_senders_raw(txn, key, data);
}
```

AndreaLanfranchi (changes requested):

node/silkworm/db/access_layer_test.cpp
AndreaLanfranchi: Another consideration ... 
Actually this breaks our stage loop functionality.
When we do stage `Senders` (i.e. we recover senders from a batch of blocks) we do expect we process blocks in ascending order and that NONE of the processing blocks has senders already persisted to db as [we use MDBX_APPEND](https://github.com/torquem-ch/silkworm/blob/master/node/silkworm/stagedsync/stage_senders/recovery_farm.cpp#L150) and so I am quite not convinced this should be pushed as is unless :

1. The implementation also takes into account to save a block with senders only if it is the next in sequence from a previous block with senders
2. It takes the duty to update the stage Senders progress incrementing by one.

peroket: This function is not used in the senders stage, so I am not sure it should do extra work to ensure a usage it is not planned for. It makes sense to me the caller should do that work if they wish to.
On top of that it sounds difficult to ensure from that function and/or a lot of work that would make it perform less.

As of this moment I see it used in silkworm only to write the genesis block. With this change it can support a bit more uses.
I'm happy to change the implementation as you wish, but the extra check would make it less useful to me as well. Can we agree that this is not meant for the senders stage and let it be used as an extra tool?

AndreaLanfranchi: A little clarification here: I don't mean this change is used in the Sender's stage. I mean it **affects** Senders execution.
Picture this :
- We're syncing from genesis to block 10
- For some reason before Sender's stage triggers, some portions of code have saved Blocks 1 5 and 7 with senders whilest 2 3 4 6 8 9 10 are not
- Sender's stage now cannot proceed as designed as it should detect three blocks already have senders (so we don't need to recover them) and in addition we cannot save senders in append mode (as keys for blocks 2-6 are lower than key 7 (which already have senders).

peroket: This is to support my use case, and i guess the use case that silkworm can be used as a library as well.

I thought that adding that piece of functionality could be useful to others as well.

If it doesn't make sense to you I'm not going to push it further.

AndreaLanfranchi: I understand that and really willing to help.
My concerns are :
1) We should not allow to populate sender's addresses which do not come from a sender's address recovery on behalf of transaction signature.
2) Should a transaction be assigned the wrong address then all cascading data would be invalidated and might produce martian results (wrong states, wrong tries)

I reiterate : don't want to push back use cases but we must ensure the data into db is consistent and overall performance of code is not impaired.

peroket: I'm happy to do 1. but I can't guarantee the one above, that someone would purposfully fill senders partially. I cannot afford to check that everything is filled everytime I call the `write_body` function.
So doing the implementation as you suggest, but not adding any extra check in `write_body`, is that something that would be ok?

node/silkworm/db/access_layer.cpp
peroket: Done

node/silkworm/db/access_layer_test.cpp
peroket: I applied the changes from your first comment, not sure if I should apply the changes here in this PR or if it would be better in a dedicated one?

canepat:
The proposed changes seem not to fit well into the Silkworm codebase for the reasons already outlined in previous comments by @AndreaLanfranchi.

Generally speaking, usage of Silkworm as a library even without relying on the staged sync loop is absolutely possible and we're keen to support it, but in this specific case the `write_senders` function seems better suited for inclusion in the application code rather than in Silkworm itself.

PR #667, opened 2022-06-01
Extend evm tracer

canepat:
This PR improves the `EvmTracer` hooks needed by Silkrpc for `trace_*` API implementation. In particular:
- add new `EvmTracer::on_precompiled_run` and `EvmTracer::on_reward_granted` methods used in Silkworm `EVM::call` and in Silkrpc respectively
- use `CallResult` as the `result` parameter type in `EvmTracer` hooks

The proposed changes maintain and exposes the registered tracers also in `EVM` because:
- the new hooks (`on_precompiled_run`/`on_reward_granted`) are dispatched within Silkworm/Silkrpc, hence accessing the list of registered tracers is necessary;
- obviously `evmone` already maintains the tracers, but they are accessible only inside `evmone` itself. `evmone::VM` has `get_tracer` method exposing the first tracer, but `evmone::Tracer` does not support `next_tracer` method or similar.

### DIFF
--- a/core/silkworm/execution/evm.cpp
+++ b/core/silkworm/execution/evm.cpp
@@ -50,8 +50,7 @@ class DelegatingTracer : public evmone::Tracer {
         tracer_.on_instruction_start(pc, stack_top, stack_height, state, intra_block_state_);
     }
 
-    void on_execution_end(const evmc_result& res) noexcept override {
-        CallResult result{res.status_code, static_cast<uint64_t>(res.gas_left), {res.output_data, res.output_size}};
+    void on_execution_end(const evmc_result& result) noexcept override {
         tracer_.on_execution_end(result, intra_block_state_);
     }
 
@@ -227,11 +226,8 @@ evmc::result EVM::call(const evmc_message& message) noexcept {
         }
         // Explicitly notify registered tracers (if any)
         if (!tracers_.empty()) {
-            uint64_t gas_left{static_cast<uint64_t>(res.gas_left)};
-            Bytes data{res.output_data, res.output_size};
-            CallResult result{res.status_code, gas_left, data};
             for (auto tracer : tracers_) {
-                tracer.get().on_precompiled_run(result, static_cast<uint64_t>(message.gas), state_);
+                tracer.get().on_precompiled_run(res, message.gas, state_);
             }
         }
     } else {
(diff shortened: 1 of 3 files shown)


yperbasis (approved):

core/silkworm/execution/evm.cpp
```diff
@@ -224,9 +225,18 @@ evmc::result EVM::call(const evmc_message& message) noexcept {
                 res.status_code = EVMC_PRECOMPILE_FAILURE;
             }
         }
+        // Explicitly notify registered tracers (if any)
+        if (tracers_.size() > 0) {
```
AndreaLanfranchi, line 229:
```suggestion
        if (!tracers_.empty()) {
```

core/silkworm/execution/evm.cpp
```diff
@@ -50,7 +50,8 @@ class DelegatingTracer : public evmone::Tracer {
         tracer_.on_instruction_start(pc, stack_top, stack_height, state, intra_block_state_);
     }
 
-    void on_execution_end(const evmc_result& result) noexcept override {
+    void on_execution_end(const evmc_result& res) noexcept override {
+        CallResult result{res.status_code, static_cast<uint64_t>(res.gas_left), {res.output_data, res.output_size}};
```
AndreaLanfranchi, line 54: Question ... why cast `res.gas_left` from int64_t to uint64_t ? I believe execution can return negative gas left which later raises an error of insufficient gas and casting might lead to martian value. Am I missing something ?

canepat: Good catch. I saw the same conversion applied at the end of `EVM::execute` method, but I missed that both `EVM::create` and `EVM::call` take care to set the gas left to zero in case of any error except `EVMC_REVERT`.

So now the choice is between:
- applying the same reset of gas left to zero in case of any error except `EVMC_REVERT` also in `DelegateTracer::on_execution_end` before calling the tracer. This would allow to maintain the same `CallResult` parameter type for all `EvmTracer` methods carrying an execution result and also would keep `EvmTracer` consistent with the return type of `EVM::execute`
- reverting to use the previous result type, i.e. `evmc_result` in `EvmTracer` in order to propagate to the tracer implementations exactly what the `evmone` reported after execution

I slightly prefer the former in terms of interface coherence, but the latter seems more correct to allow any tracer implementation to log also negative gas left after execution, if needed.

What do you think?

AndreaLanfranchi: Sorry for not replying yesterday.
I agree data correctness is better than code coherence.

canepat: OK done in last commit

PR #669, opened 2022-06-03
C++20 as default toolchain

canepat:
This PR makes C++20 the default toolchain for Silkworm


AndreaLanfranchi (approved):
Ok for me ... will lately add a note on minimum MSVC version for Windows users

yperbasis (approved):

PR #670, opened 2022-06-05
Small C++20 modernizations

yperbasis:
- [std::span](https://en.cppreference.com/w/cpp/container/span) instead of `gsl::span`
- [starts_with](https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with) instead of `has_prefix`


AndreaLanfranchi (approved):

mriccobene (approved):
LGTM

PR #677, opened 2022-06-09
Extract transaction_root & ommers_hash algorithms

mriccobene:
The body downloader needs to compute the transaction root and the hash of ommers.
This PR extracts the relevant code from pre_validate_block() and puts it in common.

### DIFF
--- a/core/silkworm/consensus/base/engine.cpp
+++ b/core/silkworm/consensus/base/engine.cpp
@@ -15,6 +15,7 @@
 */
 
 #include "engine.hpp"
+#include "silkworm/common/cast.hpp"
 
 #include <silkworm/chain/protocol_param.hpp>
 #include <silkworm/common/as_range.hpp>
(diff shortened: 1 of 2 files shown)


core/silkworm/consensus/base/engine.cpp
```diff
@@ -235,4 +230,26 @@ std::optional<intx::uint256> EngineBase::expected_base_fee_per_gas(const BlockHe
     }
 }
 
+evmc::bytes32 EngineBase::compute_transaction_root(const BlockBody& body) {
+    if (body.transactions.empty())
```
AndreaLanfranchi, line 234: Would be nice to keep curly braces for `if` blocks

AndreaLanfranchi (changes requested):
To keep hierarchy consistent I'd suggest to make both `compute_transaction_root` and `compute_ommers_hash` as virtual methods for the `IEngine` interface. I understand ethash engine inherits from `EngineBase` (and eventually `NoProof` engine inherits from ethash) but that is not the case for `MergeEngine`

core/silkworm/consensus/base/engine.cpp
mriccobene: done

AndreaLanfranchi (approved):

PR #680, opened 2022-06-13
Wait strategies for RPC server

canepat:
This PR plugs different wait strategies to handle idle cycles in the RPC server execution loop.

Also changes the BackEnd & KV standalone server to avoid logs when printing usage.


mriccobene (approved):
LGTM

PR #682, opened 2022-06-14
Default comparison operators

yperbasis:
Use C++20 [default comparisons](https://en.cppreference.com/w/cpp/language/default_comparisons).


AndreaLanfranchi (approved):
I like three way comp operator

PR #684, opened 2022-06-16
Remove left_pad & right_pad

yperbasis:


AndreaLanfranchi (commented):
Though I agree they might not be used at the moment generally speaking padding is often useful in logging ... maybe a change to support the "filling" char would be nice.
However if we want to keep code base as small as possible ok for me to remove.

canepat (approved):

PR #685, opened 2022-06-16
Reduce body downloader validation expense

mriccobene:
Body downloader only need to validate ommers. 
This PR 
- refactors _consensus::EngineBase_ `pre_validate_block()` introducing a `validate_ommers()` method
- in the body downloader replaces the call to `pre_validate_block()` with with `validate_ommers()`

Note that:
- ommers hash is computed elsewere
- for extensibility downloader calls `consensus::engine_factory()` to get a suitable consensus engine
- `consensus::engine_factory() `returns an object derived from _IEngine_ interface

so we had to add `validate_ommers() `to the _IEngine_ interface.

Reviewers knowing the details of the different consensus engine are asked to give a guidance here: is `validate_ommers()` meaningful in all consensus engine? do we need a more branched hierarchy to better state the differences between different consensus engines?

Also consider the position of static methods `compute_transaction_root()` and `compute_ommers_hash()` in _EngineBase_ and get guidance to move them (if needed) to the upper _IEngine_ class as noted [here](https://github.com/torquem-ch/silkworm/pull/677#pullrequestreview-1001476937)

### DIFF
--- a/core/silkworm/consensus/base/engine.cpp
+++ b/core/silkworm/consensus/base/engine.cpp
@@ -55,6 +55,12 @@ ValidationResult EngineBase::pre_validate_block(const Block& block, const BlockS
         return ValidationResult::kWrongOmmersHash;
     }
 
+    return validate_ommers(block, state);
+}
+
+ValidationResult EngineBase::validate_ommers(const Block& block, const BlockState& state) {
+    const BlockHeader& header{block.header};
+
     if (block.ommers.size() > 2) {
         return ValidationResult::kTooManyOmmers;
     }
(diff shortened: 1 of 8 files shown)


yperbasis:
In my opinion it's better no to try to extract `validate_ommers`, but to call `pre_validate_block` instead. `pre_validate_block` should be fast and performance impact of the extra checks besides ommer validation is negligible.

yperbasis:
Actually, sorry, I think there's a slow bit in `pre_validate_block`, which is `validate_seal`. One possibility is to refactor `pre_validate_block` so that it doesn't call `validate_block_header`.

yperbasis (approved):

PR #687, opened 2022-06-20
Refactor commands CMakeLists

battlmonstr:
* sort commands alphabetically
* codify a convention that each cmd has a corresponding main cpp file with the same name
* CLI11 and silkworm_node libraries (COMMON_LIBS) are required for each cmd
* adding custom libs or sources is still possible with LIBS/SRC

Note: silkworm_libs variable is introduced as a replacement for `$<$<BOOL:${MSVC}>`, which didn't expand when passed as a parameter.

Note: a similar goal refactoring was done in [erigon's Makefile](https://github.com/ledgerwatch/erigon/blob/devel/Makefile#L68) for look-alike commands to avoid code duplication and discrepancies.

### DIFF
--- a/cmd/CMakeLists.txt
+++ b/cmd/CMakeLists.txt
@@ -14,6 +14,20 @@
    limitations under the License.
 ]]
 
+function(cmd name)
+  cmake_parse_arguments(PARSE_ARGV 1 ARG "NO_NODE_LIB" "" "SRC;LIBS")
+
+  list(APPEND ARG_SRC ${name}.cpp)
+  add_executable(${name} ${ARG_SRC})
+
+  list(PREPEND ARG_LIBS CLI11::CLI11)
+  if(NOT ARG_NO_NODE_LIB)
+    list(APPEND ARG_LIBS silkworm_node)
+  endif()
+
+  target_link_libraries(${name} PRIVATE ${ARG_LIBS})
+endfunction()
+
 if(MSVC)
   add_link_options(/STACK:0x1000000)
 elseif (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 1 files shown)


canepat:
I appreciate the intent of this PR, nevertheless I agree with @AndreaLanfranchi about losing WYSIWYG. In particular, IMHO the `NO_NODE_LIB` argument of the `cmd` function is a bit obscure at first glance.

I haven't a strong opinion about this PR. I'm not against it, but I would require to remove the `NO_NODE_LIB` argument and explicitly list all the libraries where `cmd` is called, so that both sources and libraries are visible at the call site.

battlmonstr:
@canepat Removed NO_NODE_LIB. The silkworm_node library is mandatory for all the commands now. It doesn't make much sense to special case genesistool as it obscures it for the rest.
I have also extracted the COMMON_LIBS variable so that it is explicit now. Since CLI11 and silkworm_node are common, does it make sense to list them for each command?

@AndreaLanfranchi yes, functions do hide real code. In terms of readability `cmd(bla)` is as explicit as it can be. I have added some docs for you ❤️ so that it is not as scary/unfamiliar.

AndreaLanfranchi:
> @canepat Removed NO_NODE_LIB. The silkworm_node library is mandatory for all the commands now. It doesn't make much sense to special case genesistool as it obscures it for the rest. I have also extracted the COMMON_LIBS variable so that it is explicit now. Since CLI11 and silkworm_node are common, does it make sense to list them for each command?
> 
> @AndreaLanfranchi yes, functions do hide real code. In terms of readability `cmd(bla)` is as explicit as it can be. I have added some docs for you ❤️ so that it is not as scary/unfamiliar.

Did I make something to you ? I really wish not to be trated as an idiot ok ?
To be precise this is not a function rather a MACRO which rewrites the cmake build files.
For the very same reasons we do NOT want to use MACROs in C++ code as well.

battlmonstr:
@AndreaLanfranchi just to be clear, I was sincere, no sarcasm. If I offended you, I apologize.

mriccobene:
I would prefer the current version of cmd/CMakeLists.txt to be unmodified as it is easier to read and modify. Also I am concerned that Clion may not automatically recognize targets.

PR #689, opened 2022-06-21
Add PROJECT_VERSION

yperbasis:
IMHO the good old [semantic versioning](https://semver.org/) is better than calendar-based versioning for our projects. For example, with the calendar versions in Erigon there's a discrepancy between the "Erigon 2" moniker and the actual versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,12 @@ HunterGate(
 ]]
 
 project(silkworm)
+set(PROJECT_VERSION 0.1.0-dev)
+
+string(REGEX MATCH "([0-9]+)\\.([0-9]+)\\.([0-9]+)" _ ${PROJECT_VERSION})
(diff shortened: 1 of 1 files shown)


yperbasis:
> Wouldn't it be better to use tags in github repo ? So we don't need to change the code every time.

We can do potentially do something clever when we actually release a tag on github. Or follow geth and simply make the necessary change manually. But until we do make a release "0.1.0-dev" should be fine I think.

yperbasis:
So basically "0.1.0-dev" means that we haven't released version 0.1.0 yet, but we're working towards it :)

canepat:
> > Wouldn't it be better to use tags in github repo ? So we don't need to change the code every time.
> 
> We can potentially do something clever when we actually release a tag on github. Or follow geth and simply make the necessary change manually. But until we do make a release "0.1.0-dev" should be fine I think.

It's perfectly fine to use "0.1.0-dev" until we start making releases, then I'd switch to git tags.

PR #693, opened 2022-06-23
A body downloader that penalizes peers on malformed messages

mriccobene:
On some test runs the downloader received many malformed messages. 
Tests seems to show that in such circumstances the downloader has a significative performance penalty.
The body downloader should penalize the corresponding peer so that Sentry will remove the buggy peer.

### DIFF
--- a/node/silkworm/downloader/block_exchange.cpp
+++ b/node/silkworm/downloader/block_exchange.cpp
@@ -23,6 +23,7 @@ limitations under the License.
 #include <silkworm/downloader/internals/preverified_hashes.hpp>
 #include <silkworm/downloader/messages/inbound_message.hpp>
 #include <silkworm/downloader/messages/outbound_get_block_bodies.hpp>
+#include <silkworm/downloader/rpc/penalize_peer.hpp>
 
 namespace silkworm {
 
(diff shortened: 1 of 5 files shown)


node/silkworm/downloader/block_exchange.cpp
```diff
@@ -121,4 +125,12 @@ void BlockExchange::log_status() {
                  << "; stats: " << body_sequence_.statistics();
 }
 
+void BlockExchange::send_penalization(PeerId id, Penalty p) noexcept {
+    using namespace std::chrono_literals;
+    rpc::PenalizePeer penalize_peer(id, p);
+    penalize_peer.do_not_throw_on_failure();
+    penalize_peer.timeout(1s);
```
canepat, line 132: A constant in `block_exchange.hpp` instead of a magic number in code would be preferable here.

canepat (changes requested):

node/silkworm/downloader/block_exchange.cpp
mriccobene: done

canepat (approved):

PR #696, opened 2022-06-28
Sentry: GRPC calls stubs

battlmonstr:
Similar to rpc/server/BackEndService for now.
Looking forward to convert this to `awaitable`-s at some point.
Note: p_impl is used to avoid duplicating call and call factory types boilerplate in .h


canepat:
> Looking forward to convert this to `awaitable`-s at some point.

Yes, looking forward too. I'm currently using asio-grpc library in a Silkrpc unpublished branch to simplify the usage of gRPC with coroutines on client side (compilation still work in progress, then need to check performances). Once finished, we can surely try to use it server-side as well.

> Note: p_impl is used to avoid duplicating call and call factory types boilerplate in .h

Nice idea!

canepat (approved):
LGTM

PR #697, opened 2022-07-05
sentry: --nat and --port options

battlmonstr:
Note: I've tried to use [this example](https://github.com/CLIUtils/CLI11/blob/main/examples/custom_parse.cpp) to parse my custom type. It says:

> the lexical cast operator should be in the same namespace as the type

but this didn't work (it tried using a builtin CLI::detail::lexical_cast instead, and didn't find mine).

So instead I had to manually call this function as [described here](https://github.com/CLIUtils/CLI11/blob/main/book/chapters/advanced-topics.md#custom-option-callbacks).

I wonder if this is because we are not using the latest CLI11 version? (How to find out which version is used?)

### DIFF
--- a/cmd/sentry.cpp
+++ b/cmd/sentry.cpp
@@ -38,7 +38,10 @@ int sentry_main(int argc, char* argv[]) {
     cli.add_option("--sentry.api.addr", options.api_address, "GRPC API endpoint")
         ->capture_default_str()
         ->check(IPEndPointValidator(/*allow_empty=*/true));
-    cli.add_option("--port", options.port, "Network listening port")
+
+    cli.add_option("--port", options.port)
+        ->description("Network listening port for incoming peers TCP connections and discovery UDP requests")
+        ->check(CLI::Range(1024, 65535))
         ->capture_default_str();
 
     auto nat_option = cli.add_option("--nat", [&options](CLI::results_t results) {
--- a/cmd/sentry.cpp
+++ b/cmd/sentry.cpp
@@ -38,6 +38,17 @@ int sentry_main(int argc, char* argv[]) {
     cli.add_option("--sentry.api.addr", options.api_address, "GRPC API endpoint")
         ->capture_default_str()
         ->check(IPEndPointValidator(/*allow_empty=*/true));
+    cli.add_option("--port", options.port, "Network listening port")
+        ->capture_default_str();
+
+    auto nat_option = cli.add_option("--nat", [&options](CLI::results_t results) {
+       return lexical_cast(results[0], options.nat);
+    });
+    nat_option->description("NAT port mapping mechanism (none|extip:<IP>)\n"
+            "- none              no NAT, use a local IP as public\n"
+            "- extip:1.2.3.4     use the given public IP");
+    nat_option->default_str("none");
+
     add_option_num_contexts(cli, options.num_contexts);
     add_option_wait_mode(cli, options.wait_mode);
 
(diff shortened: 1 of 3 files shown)


cmd/sentry.cpp
```diff
@@ -38,6 +38,17 @@ int sentry_main(int argc, char* argv[]) {
     cli.add_option("--sentry.api.addr", options.api_address, "GRPC API endpoint")
         ->capture_default_str()
         ->check(IPEndPointValidator(/*allow_empty=*/true));
+    cli.add_option("--port", options.port, "Network listening port")
```
AndreaLanfranchi, line 41: Not sure what this `--port` is for but need to be enforced a check that is a valid number in [1-65535]

AndreaLanfranchi (dismissed):

sentry/sentry/nat_option.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+bool lexical_cast(const std::string& input, NatOption& value) {
+    if (input == "none") {
+        value = {};
+        return true;
+    }
+    if (boost::algorithm::istarts_with(input, "extip:")) {
+        std::string ip_str = input.substr(6);
+        boost::system::error_code err;
+        auto ip = boost::asio::ip::address::from_string(ip_str, err);
```
AndreaLanfranchi, line 30: Why copy allocate a new string instead of consuming current one ?

```suggestion
        input.erase(0,6);
        boost::system::error_code err;
        auto ip = boost::asio::ip::address::from_string(input, err);
```
This obviously implies removing `const` from input.

Also I believe this
`auto ip = boost::asio::ip::address::from_string(ip_str, err);`
is meant to fail if input is `xxx.xxx.xxx.xxx:yyyyy` where y are placeholders for port number

battlmonstr: I got it fixed in a different way, because such parameter modification would be an unexpected side effect from the caller's point of view.
I'd rather use something like a string_view here, but this boost API can't deal with it.

As for the port, yes, it does fail on such an input as expected. The option only accepts the IP, no port is allowed.

sentry/sentry/nat_option.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace silkworm::sentry {
+
+bool lexical_cast(const std::string& input, NatOption& value) {
+    if (input == "none") {
+        value = {};
+        return true;
+    }
+    if (boost::algorithm::istarts_with(input, "extip:")) {
+        auto ip_str = input.c_str() + 6;
```
AndreaLanfranchi, line 28: I'd rather 

```suggestion
        const auto ip_str{&input[6]};
```

canepat (approved):

sentry/sentry/nat_option.cpp
battlmonstr: I did it this way, because `data()` is not guaranteed to be null-terminated. Isn't `&input[6]` is a pointer inside data? `operator[]` docs says nothing about null-termination.

I've noticed that on macOS `data()` and `c_str()` are practically the same, but I was not sure about the other platforms, so I decided to stay safe.

AndreaLanfranchi: It is the same.
This is implementation of `c_str()` which simply returns the base pointer to data
```
return _Mypair._Myval2._Myptr();
```
This is the implementation of `data()`
```
return _Mypair._Myval2._Myptr();
```
As you see are identical and carry the same assertions.

Random access operator [] is implemented like this
```
return _Mypair._Myval2._Myptr()[_Off];
```

Hence what proposed is more concise and more C++ style.

AndreaLanfranchi: Besides we have the style consraint to brace initialize vars
so 
`auto ip_str = input.c_str() + 6;`
has to be corrected with
`auto ip_str {input.c_str() + 6};`

AndreaLanfranchi: Since C++11 data is **always** guaranteed to be null terminated.

PR #705, opened 2022-07-08
UnsignedIntegral concept

yperbasis:
Introduce `UnsignedIntegral` concept equal to [std::unsigned_integral](https://en.cppreference.com/w/cpp/concepts/unsigned_integral) + [intx::uint](https://github.com/chfast/intx) types.


mriccobene (approved):
I like it

PR #709, opened 2022-07-12
Fix include guards convention

battlmonstr:


yperbasis:
It'd also be nice to add this deviation from Google's Style Guide into our README.

Issue #674, opened 2022-06-07
Build warning: absl is using a private weak std::noshowbase


battlmonstr:
Building on macOS using:
```
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug -DGMP_INCLUDE_DIR=... -DGMP_LIBRARY=...
ninja -j4
```

Produces a warning:
```
ld: warning: direct access in function 'absl::lts_20211102::(anonymous namespace)::Uint128ToFormattedString(absl::lts_20211102::uint128, unsigned int)' from file '/Users/daniel/.hunter/_Base/4942227/e433264/ca8c68e/Install/lib/libabsl_int128d.a(int128.cc.o)' to global weak symbol 'std::__1::noshowbase(std::__1::ios_base&)' from file '/Users/daniel/.hunter/_Base/4942227/e433264/ca8c68e/Install/lib/libprotobufd.a(int128.cc.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

It looks like some absl code (int128.cc) wants to access std::noshowbase, but instead of using it directly, it finds it as a "private" symbol in libprotobuf.
This feels unexpected, and is likely some kind of misconfiguration of the build.

cmake version 3.23.2
build tools from Xcode 13.3

See also https://github.com/torquem-ch/silkworm/issues/673

Issue #688, opened 2022-06-20
link warning: libmdbxd.a(lck-windows.c.o) has no symbols


battlmonstr:
Building on macOS shows this warning during linking:
```
[6/8] Linking CXX static library third_party/libmdbx/libmdbxd.a
/Applications/Xcode133.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: third_party/libmdbx/libmdbxd.a(lck-windows.c.o) has no symbols
```

battlmonstr:
Fix: https://github.com/torquem-ch/libmdbx/pull/2

Issue #673, opened 2022-06-07
Build warning: using a private weak EpsCopyOutputStream::Encode64


battlmonstr:
Building on macOS using:
```
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug -DGMP_INCLUDE_DIR=... -DGMP_LIBRARY=...
ninja -j4
```

Produces warnings:
```
ld: warning: direct access in function 'sentry::Forks::_InternalSerialize(unsigned char*, google::protobuf::io::EpsCopyOutputStream*) const' from file 'node/libsilkworm_noded.a(sentry.pb.cc.o)' to global weak symbol 'google::protobuf::io::EpsCopyOutputStream::Encode64(unsigned long long)' from file '/Users/daniel/.hunter/_Base/4942227/e433264/ca8c68e/Install/lib/libprotobufd.a(descriptor.pb.cc.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in function 'sentry::MessagesRequest::_InternalSerialize(unsigned char*, google::protobuf::io::EpsCopyOutputStream*) const' from file 'node/libsilkworm_noded.a(sentry.pb.cc.o)' to global weak symbol 'google::protobuf::io::EpsCopyOutputStream::Encode64(unsigned long long)' from file '/Users/daniel/.hunter/_Base/4942227/e433264/ca8c68e/Install/lib/libprotobufd.a(descriptor.pb.cc.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

It looks like the generated sentry client code (sentry.pb.cc) wants to access "private" symbols in libprotobuf.
This feels to be expected, so this is likely some kind of misconfiguration of the build.

cmake version 3.23.2
build tools from Xcode 13.3

battlmonstr:
Fixed by https://github.com/torquem-ch/silkworm/pull/710

PR #710, opened 2022-07-12
Fix link warnings on macOS

battlmonstr:
Fixes https://github.com/torquem-ch/silkworm/issues/674 and https://github.com/torquem-ch/silkworm/issues/673 and https://github.com/torquem-ch/silkworm/issues/688

References:
https://github.com/protocolbuffers/protobuf/issues/8958
https://stackoverflow.com/questions/9894961/strange-warnings-from-the-linker-ld
https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/CppRuntimeEnv/Articles/SymbolVisibility.html#//apple_ref/doc/uid/TP40001670-98807

### DIFF
--- a/cmake/Hunter/config.cmake
+++ b/cmake/Hunter/config.cmake
@@ -15,6 +15,14 @@
 ]]
 
 hunter_config(
+  abseil
+  VERSION 20220623.0
+  URL https://github.com/abseil/abseil-cpp/archive/20220623.0.tar.gz
+  SHA1 144c2108e1532c642cdb6ca532ee26e91146cf28
+  CMAKE_ARGS ABSL_PROPAGATE_CXX_STD=ON
+)
+
+hunter_config(
   Microsoft.GSL
   VERSION 4.0.0
   URL https://github.com/microsoft/GSL/archive/v4.0.0.tar.gz
@@ -32,6 +40,11 @@ hunter_config(
 hunter_config(
   Protobuf
   VERSION 3.14.0-4a09d77-p0
+  CMAKE_ARGS
+    CMAKE_POLICY_DEFAULT_CMP0063=NEW
+    CMAKE_C_VISIBILITY_PRESET=hidden
+    CMAKE_CXX_VISIBILITY_PRESET=hidden
+    CMAKE_VISIBILITY_INLINES_HIDDEN=ON
 )
 
 hunter_config(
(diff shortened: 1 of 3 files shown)


battlmonstr:
@AndreaLanfranchi please help to check if it works on Windows

node/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ find_package(gRPC CONFIG REQUIRED)
 get_filename_component(SILKWORM_MAIN_DIR ../ ABSOLUTE)
 
 file(GLOB_RECURSE SILKWORM_NODE_SRC CONFIGURE_DEPENDS "*.cpp" "*.hpp" "*.c" "*.h" "*.cc")
-list(FILTER SILKWORM_NODE_SRC EXCLUDE REGEX "_test\.cpp$")
+list(FILTER SILKWORM_NODE_SRC EXCLUDE REGEX "_test\\.cpp$")
```
yperbasis, line 25: Let's make the same change in `core/CMakeLists.txt`

cmake/Hunter/config.cmake
```diff
@@ -32,6 +40,11 @@ hunter_config(
 hunter_config(
   Protobuf
   VERSION 3.14.0-4a09d77-p0
+  CMAKE_ARGS
```
yperbasis, line 43: But aren't these cmake policies passed from our toolchain `cmake/toolchain/base.cmake` ?

battlmonstr: Oh, I didn't know we already set them globally.

I see these variable values from `base.cmake` in config.cmake if I print them with `message()`,
but I'm not sure that they are propagated to the Protobuf library.

Inside
~/.hunter/_Base/Download/Hunter/0.24.3/10738b5/Unpacked/cmake/projects/Protobuf/hunter.cmake
they call `hunter_cmake_args()` ([declared here](https://github.com/ruslo/hunter/blob/master/cmake/modules/hunter_get_cmake_args.cmake)), but it is not clear to me whether it merges with the parent scope or overrides it.

I don't know how to verify it other than see if you get these link warnings before and after this patch.
Do you have them on master and on this branch?

yperbasis: Hmm, I do get linker warning w/o these explicit `CMAKE_ARGS` for protobuf. Strange, how come Hunter doesn't propagate cmake settings from the toolchain 🤷‍♂️

PR #708, opened 2022-07-11
sentry: TCP server

battlmonstr:
For Sentry::stop() I have tried to use [cancellation_signal](https://www.boost.org/doc/libs/1_79_0/doc/html/boost_asio/reference/cancellation_signal.html) directly, which would be ideal, but failed to make it work with either use_future or use_awaitable.

The idea of CancellationSignal implementation is taken from https://stackoverflow.com/questions/71310261/asio-c-coroutine-cancellation . 

see also:
* https://www.boost.org/doc/libs/1_79_0/doc/html/boost_asio/overview/core/cancellation.html
* https://cppalliance.org/richard/2021/05/30/RichardsMayUpdate.html
* https://www.youtube.com/watch?v=hHk5OXlKVFg

PR #713, opened 2022-07-20
sentry: ECIES

battlmonstr:
References:
https://github.com/ethereum/devp2p/blob/master/rlpx.md#ecies-encryption
https://github.com/ledgerwatch/erigon/tree/devel/crypto/ecies   
https://github.com/stealthyinc/ECIES-Encryption
https://github.com/Safeheron/crypto-ecies-cpp

Relies on this PR merge first:
https://github.com/torquem-ch/silkpre/pull/3
https://github.com/torquem-ch/silkworm/pull/708

Note: OpenSSL is already a dependency of gRPC, so it doesn't add anything extra.


node/silkworm/common/secp256k1_context.hpp
```diff
@@ -36,10 +37,32 @@ class SecP256K1Context final {
...
+    bool verify_private_key_data(const ByteView& data) const {
         return secp256k1_ec_seckey_verify(context_, data.data());
     }
 
+    bool create_public_key(secp256k1_pubkey* public_key, const ByteView& private_key) const {
+        return secp256k1_ec_pubkey_create(context_, public_key, private_key.data());
+    }
+
+    Bytes serialize_public_key(const secp256k1_pubkey* public_key, bool is_compressed) const {
+        size_t data_size = is_compressed ? 33 : 65;
```
canepat, line 50: Please use two constants here with [this format](https://google.github.io/styleguide/cppguide.html#Constant_Names) instead of magic numbers to improve the code readability.

sentry/silkworm/sentry/common/ecc_key_pair.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <string>
+#include <silkworm/common/base.hpp>
+
+namespace silkworm::sentry::common {
+
+class EccKeyPair {
+  public:
+    EccKeyPair();
+    explicit EccKeyPair(Bytes data);
+    explicit EccKeyPair(const ByteView& data);
```
canepat, line 28: I suggest using `ByteView` instead of `const ByteView&` here for consistency at least within this code change (methods in `ecies_cipher` use pass-by-value).

This is just a hint, currently we have both pass-by-value and pass-by-ref for `ByteView` in different places within the codebase. However, IMO passing it by value is more idiomatic (see also https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/)

canepat (changes requested):

canepat (approved):

PR #719, opened 2022-07-31
Fix Xcode 14 build

yperbasis:
Prior to this change `cmake -G Xcode ..` was failing with
```
CMake Error in interfaces/CMakeLists.txt:
  The custom command generating

    /Users/andrew/silkworm/interfaces/types/types.pb.cc

  is attached to multiple targets:

    generate_sentry_grpc
    generate_kv_grpc
    generate_ethbackend_grpc

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

When that was fixed, Xcode 14 complained about some conversions.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,7 +1,9 @@
 # This is the official list of Silkworm authors for copyright purposes.
 
 Andrey Ashikhmin
+Tullio Canepa
 Greg Colvin
 Andrea Lanfranchi
+Daniel Lazarenko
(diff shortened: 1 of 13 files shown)


battlmonstr (approved):
👍 
What do you use to reformat the code?

yperbasis:
> 👍 What do you use to reformat the code?

My VS Code calls clang-format on file save.

PR #720, opened 2022-08-02
clang-format integration

battlmonstr:
Format the code with:

	cmake -P cmake/format.cmake

I've tried to reduce the churn to the minimum so I've added `ColumnLimit: 0`
It is going to respect our formatting decisions in some places rather than joining lines if they fit in 120.

I've added `AllowShortEnumsOnASingleLine: false`,
because otherwise a few enums got their members shifted, which looked worse than having them not in line.

Our deviations from the new format are mostly that the new style:
1. always adds 2 spaces before traling commments (due to `SpacesBeforeTrailingComments: 2`)
1. always aligns trailing comments (due to `AlignTrailingComments: true`)
1. aligns `<<` after a line break (don't know how to adjust this, `AlignOperands: DontAlign` is not working for this case)
1. puts `std::endl` on a new line (haven't found how to control this)

Let's have this config as a common ground.
I suggest to not start bikeshedding discussions, but leave them for subsequent PRs.
We can always improve upon this.

Tip: If you want to format a piece of code in a particular way there's an escape hatch -
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code


battlmonstr:
Feedback:
* check if it works on M1
* is there a way to check 120 line length?

.circleci/config.yml
```diff
@@ -24,6 +24,23 @@ commands:
...
+  format:
+    steps:
+      - run:
+          name: "Format"
+          working_directory: ~/project
+          command: |
+            cmake -P cmake/format.cmake
+            if ! git diff --exit-code
+            then
+              echo "The code has a different formatting style. Try to run: cmake -P cmake/format.cmake"
```
AndreaLanfranchi, line 36:
```suggestion
              echo "Code formatting style is not compliant. Try to run: cmake -P cmake/format.cmake on your local dev environment and push changes"
```

AndreaLanfranchi (changes requested):

.circleci/config.yml
AndreaLanfranchi: As a side note I'd point out that quick changes to source files using github's in place editing feature are likely to fail (github does not format)

AndreaLanfranchi:
> * is there a way to check 120 line length?

Given that a provided fixed column width forces formatter to "stretch" long lines **up to 120** chars and given this mostly happens in large function declarations or calls maybe is worth try to play with `BinPackArguments` and `BinPackParameters` setting them to `true`

BinPackArguments
```
true:
void f() {
  f(aaaaaaaaaaaaaaaaaaaa, aaaaaaaaaaaaaaaaaaaa,
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa);
}

false:
void f() {
  f(aaaaaaaaaaaaaaaaaaaa,
    aaaaaaaaaaaaaaaaaaaa,
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa);
}
```

BinPackParameters
```
true:
void f(int aaaaaaaaaaaaaaaaaaaa, int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) {}

false:
void f(int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) {}
```

This, though a bit drastical would allow to keep the enforcement of max 120 columns (which is IMHO is reasonable) without stretching long function calls and definitions.

It does not solve the problem though of long `if` conditions like 
```
        if (previous_progress && previous_progress != 5 && previous_progress != 6 && previous_progress != 7 &&
            previous_progress != 8) {
                [...]
        }
```

AndreaLanfranchi:
Nevertheless for me is ok

canepat (approved):

PR #723, opened 2022-08-06
Intermediate Hashes Stage with multiple fixes and improvements

AndreaLanfranchi:
The long awaited IH implementation.

1. Inherited logic from Erigon in collection of state changes (incremental) to account storage tries to clean-up from `TrieStorage` bucket **and** detection of newly created accounts. Full regeneration stays roughly the same. As a result 
2. Complete rewrite of `TrieCursor` to traverse the trie in search **only** of (re)usable node hashes. Totally skip nodes which are not relevant.
3. Unlike Erigon unified the collection of state changes in a single routine : less redundant code
4. Implemented in `toolbox` some trie tools: reset (to force a full regen on next cycle), integrity (as in Erigon check there are no orphans or parents with tree_state and no children - eventually also checks state_mask point to proper hashed_states)
5. TrieCursor has now allocations reduced to bare minimum and is reusable (on every to_prefix() it's state is reset)
6. HasBuilder is now reusable (reset() method) to not loose allocated capacity of vectors

Performance : on par with Erigon for full regen; 25% faster than Erigon for incremental promotion, 6 times faster than our previous impl.


node/silkworm/stagedsync/stage_interhashes/trie_cursor.hpp
```diff
@@ -0,0 +1,137 @@
...
+        [[nodiscard]] Bytes full_key() const noexcept;  // Returns full key to child node (i.e. key + child_id)
+        [[nodiscard]] Bytes hash() const;               // Returns hash of child node (i.e. key + child_id)
+    };
+
+    uint32_t level_{0};                      // Depth level in sub_nodes_
+    bool end_of_tree_{false};                // Protects from to_next() beyond end of tree
+    Bytes curr_key_{};                       // Latest key returned on a valid hash
+    Bytes prev_key_{};                       // Same as curr_key_ but for previous cycle
+    bool skip_state_{true};                  // Whether account(s) state scan can be skipped
+    std::array<SubNode, 64> sub_nodes_{{}};  // Collection of sub-nodes being unrolled
```
yperbasis, line 100: Why 64?

node/silkworm/stagedsync/stage_interhashes/trie_cursor.cpp
```diff
@@ -0,0 +1,336 @@
...
+    state_mask = 0;
+    tree_mask = 0;
+    hash_mask = 0;
+    child_id = -1;
+    max_child_id = 0x10;  // Traverse all node for ephemeral ones.
+    hash_id = -1;
+    deleted = false;
+}
+
+void TrieCursor::SubNode::parse(ByteView k, ByteView v) {
```
yperbasis, line 270: Please reduce code duplication between `TrieCursor::SubNode::parse` and `Node::from_encoded_storage`. Perhaps `SubNode` should be a subclass of `Node`.

yperbasis (changes requested):

node/silkworm/stagedsync/stage_interhashes/trie_cursor.hpp
AndreaLanfranchi: Oh ... thank you for spotting.
It's a legacy of Erigon's reasoning where "level_" was set on behalf of nibbled key len.
Actually 32 is the theoretical max value is 32 (1 for root + 31 of max length of common prefix) even if must say is over abundant ; an analysis of nibbled key lengths in trie show this values

```
 Size    Count
 ---- --------
    0        1
    1       16
    2      256
    3     4096
    4    65536
    5  1048576
    6 15197337
    7  1563721
    8     8337
    9       38
   10        1
```

node/silkworm/stagedsync/stage_interhashes/trie_cursor.cpp
AndreaLanfranchi: Actually SubNode of TrieCursor and trie::Node are totally different beasts.
The first has no allocations and all data is held in ByteViews.
The second copies hashes in a vector every-time

AndreaLanfranchi: Besides trie::Node (as being also part of WASM build) does not implement exceptions so it makes hard to understand why a  parse from encoded storage has returned std::nullopt.

A solution would be to return a parsing code (pretty much like we do with rlp::decode) for parsing trie::Node
(though it does not remove the copying of hashes in a vector).

yperbasis: Hmm. IMHO it'd still be great to remove the code duplication unless that complicates things even more. What if we 1) return error codes in `Node::from_encoded_storage` and 2) change it to accept a `Node&` as an input/output parameter instead of returning a new instance (similar to what we do in RLP decoding)? With 2) we can probably still use `Node` as a superclass of `SubNode` because we can pre-allocate instances and re-use them in decoding w/o suffering too much memory reallocations.

yperbasis: I think it'd be very nice, but it's not critical. Let's try to do it in a follow-up PR.

core/silkworm/trie/prefix_set.cpp
```diff
@@ -17,47 +17,73 @@
...
-        if (index_ == keys_.size() - 1) {
-            return false;
+    }
+}
+
+std::pair<bool, ByteView> PrefixSet::contains_and_next_marked(ByteView prefix) {
+    bool is_contained{contains(prefix)};
+    ByteView next_created{};
+
+    bool is_storage{prefix.length() >= 40};  // This is a prefix_set for storage changes
```
yperbasis, line 62: `PrefixSet` is described as generic, but here the implementation is contaminated with specifics of a particular data set. If this optimization is material, let's expose 40 as a parameter of  `contains_and_next_marked` so that its user is warned about this peculiarity.

yperbasis (changes requested):

yperbasis:
Also, why the new files `interfaces/remote/ethbackend_mock.grpc.pb.h` and `interfaces/remote/kv_mock.grpc.pb.h`?

core/silkworm/trie/prefix_set.cpp
AndreaLanfranchi: My two cents ... PrefixSet is NOT a collection of prefixes rather a collection of nibbled hashed changed accounts or storage locations. So already the naming of the class is  imho misleading. What it provides, instead, is checking if the given nibbled prefix is present in any of the full (nibbled) keys owned.

Yes the change is material : as in PrefixSet populated with changes from HashedStorage all keys are prefixed with hashed contract plus incarnation (40 bytes) it allows the retrieval of a "created" item **only** within the boundaries of that specific contract. (to be noted that this also amends an [error present in erigon where the "nextMarkedItem" is searched regardless the prefixed trie owning it](https://github.com/ledgerwatch/erigon/blob/devel/turbo/trie/retain_list.go#L184-L191)).

IMHO I'd rather change the name of the class from `PrefixSet` to `HashedChangesSet` to be more consistent with its content and usage and, considering is used only (afaics) for IH i'd omit to add the parameter you requested.

yperbasis: I respectfully disagree. `PrefixSet` does represent logically a set with the following property: If _x_ ∈ _S_ and _x_ starts with _y_, then _y_ ∈ _S_. But it doesn't have to store all the set elements to function correctly. And it doesn't matter whether the characters are nibbles (<16) or characters in the entire range [0; 256). So `PrefixSet` is a useful generic abstraction, which can be reasoned about independently from its use in Intermediate Hashes.

yperbasis: I suggest to do the following: create a new class `HashedChangesSet` and move `contains_and_next_marked` there.  (`HashedChangesSet` can be a subclass of `PrefixSet` or has it as a private member.) Thus `PrefixSet` will stay generic and it'll be clearer than `contains_and_next_marked` is not generic.

AndreaLanfranchi: I respectfully disgree with you respectful disagreement.

> If x ∈ S and x starts with y, then y ∈ S

that is correct **BUT** is `y` being the prefix looked for.
So, imho, PrefixSet is not a "set of prefixes" rather a collection of discrete "full" items which might or might not start with given prefix. 

> But it doesn't have to store all the set elements to function correctly.

Honestly don't understand this.

yperbasis: For example, let's consider the following mathematical set: {"", "a", "ab", "b", "bb", "bd"}, whose elements are all strings. It has the property that for every string in the set (e.g. "ab") all the possible prefixes (i.e. "" and "a") also belong to the set. `PrefixSet` represents such sets. With our implementation it can only store the longest possible strings (i.e. "ab", "bb", "bd"), omitting their shorter prefixes. But `contains()` still returns true for "", "a", "ab", "b", "bb", "bd" and nothing else.

As a counter-example,  set {"", "a", "ab", "c", "bb", "bd"} doesn't have the prefix property ("b" is a prefix of an element in the set, but doesn't belong to the set).

node/silkworm/trie/intermediate_hashes_test.cpp
```diff
@@ -1,780 +0,0 @@
...
-    const std::map<Bytes, Node> fused_nodes{read_all_nodes(account_trie)};
-
-    // ------------------------------------------------------------------------------
-    // A and B should yield the same result
-    // ------------------------------------------------------------------------------
-    CHECK(fused_root == incremental_root);
-    CHECK(fused_nodes == incremental_nodes);
-}
-
-TEST_CASE("Incremental vs regeneration for storage") {
```
yperbasis, line 579: There were valuable tests in intermediate_hashes_test.cpp. Is it possible to restore them?

core/silkworm/trie/prefix_set.cpp
AndreaLanfranchi: In the set {"a", "ab", "b", "bb", "bd"} (note I've omitted empty string) `contains()` returns true for "", "a", "ab", "b", "bb", "bd" as every string `starts_with` an empty string (starts_with does a memcmp on zero bytes and is guaranteed to return 0)

In the set {"a", "ab", "bb", "bd", "c"} `contains()` still returns true for "", "a", "ab", "b", "bb", "bd" **and** also for "c"

So if we look at the assertion

> If x ∈ S and x starts with y, then y ∈ S

you mean the second set should enforce a return of `contains("b") == false` ? The `and` is not satisfied i.e. "b" !∈ S

yperbasis: The second set {"", "a", "ab", "c", "bb", "bd"} was just an example of a set that doesn't have the prefix property (If _x_ ∈ _S_ and _x_ starts with _y_, then _y_ ∈ _S_) and thus not representable by `PrefixSet`.

node/silkworm/trie/intermediate_hashes_test.cpp
AndreaLanfranchi: They need to be mostly re-written cause the logic differs.
For instance old implementation returned every possible node during movement
```
 cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "10");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "10B1");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "10B3");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "11");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "13");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "131");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "132");	
    cursor.next();	
    CHECK(nibbles_to_hex(*cursor.key()) == "133");
```
while actual implementation does not

Will close this PR until done

yperbasis: I'm not that worried about lower-level test cases like "Cursor traversal 1", but higher-level cases like "Incremental vs regeneration" and "Incremental vs regeneration for storage" are probably worth preserving.

AndreaLanfranchi:
Unfortunately have to keep this open for immediate visualization of codecov misses.
Sorry for shit emails

node/silkworm/trie/intermediate_hashes_test.cpp
AndreaLanfranchi: It will take long and and a snsible rewrite to be able to expose the proper methods to the tests and to unbind from node_settings, logging features, and remove check of expected root retrieved from canonical block (which tests do not save). Doing it but don't expect results quickly

yperbasis: > It will take long and and a sensible rewrite to be able to expose the proper methods to the tests and to unbind from node_settings, logging features, and remove check of expected root retrieved from canonical block (which tests do not save). Doing it but don't expect results quickly

OK, so do you want to restore "Incremental vs regeneration" and other test cases in this PR or in a follow-up one?

AndreaLanfranchi: I already started refactoring. Really wish this PR goes through without need for follow-ups

yperbasis (approved):

PR #724, opened 2022-08-09
sentry: auth handshake

battlmonstr:


node/silkworm/common/secp256k1_context.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "secp256k1_context.hpp"
+
+namespace silkworm {
+
+const size_t SecP256K1Context::kPublicKeySizeCompressed = 33;
+const size_t SecP256K1Context::kPublicKeySizeUncompressed = 65;
+
+Bytes SecP256K1Context::serialize_public_key(const secp256k1_pubkey* public_key, bool is_compressed) const {
+    size_t data_size = is_compressed ? kPublicKeySizeCompressed : kPublicKeySizeUncompressed;
```
canepat, line 25: `data_size` can be declared `const` here

sentry/silkworm/sentry/rlpx/auth/ecies_cipher.cpp
```diff
@@ -68,29 +75,40 @@ EciesCipher::Message EciesCipher::encrypt_message(ByteView plain_text, PublicKey
...
+    return size
+        + SecP256K1Context::kPublicKeySizeUncompressed
+        + AES_BLOCK_SIZE
+        + kMacSize;
+}
+
+// NIST SP 800-56 Concatenation Key Derivation Function (see section 5.8.1).
+// Since sha256 produces the right size, one iteration is enough.
+static Bytes kdf(ByteView secret) {
+    assert(secret.size() == kKeySize * 2);
```
canepat, line 105: I suggest using `SILKWORM_ASSERT` instead of `assert` for uniformity w/ the rest of the code base

canepat (approved):

node/silkworm/common/secp256k1_context.cpp
battlmonstr: secp256k1_ec_pubkey_serialize uses this value for output

sentry/silkworm/sentry/rlpx/auth/ecies_cipher.cpp
battlmonstr: Should I change everywhere in my code then?

I've used `assert()` as a sanity check while debugging in places where I think it'd be unexpected given the assumptions around the code, and thus redundant to check in Release mode. If I need to check in Release, I'd rather throw an exception than crash.

Could we make SILKWORM_ASSERT debug-only by default (and e.g. have a build flag for enabling it in Release)?

PR #725, opened 2022-08-10
CI: check copyright header format

battlmonstr:
Reformat to the common style:
* 3 spaces in each line
* 7 spaces in the URL line
 
This format was already followed in the vast majority of the source code files.

* Keep only the current year 2022 as the LICENSE prescribes
 ( https://github.com/torquem-ch/silkworm/blob/master/LICENSE#L189 )
This simplifies the header management.
The initial year can be extracted from git if needed.

Run locally with:

    cmake -P cmake/copyright.cmake


AndreaLanfranchi:
> * Keep only the current year 2022 as the LICENSE prescribes
>   ( https://github.com/torquem-ch/silkworm/blob/master/LICENSE#L189 )
>   This simplifies the header management.

I don't agree on this due to :

> In the hyphenated form, the first date and the second date serve different purposes.
The first date serves to indicate when the earliest material in the file was authored. Implicitly you're saying that any material in this file that was authored prior to that date is not part of what you are claiming copyright to. For example, if I take a file from you in 2006 and I start adding to it, my "Copyright 2006-8" indicates that anything in versions prior to 2006 wasn't covered by this notice.
The last date serves to indicate the latest copyright date for any material in the file. So if the copyright is held by a corporation, 99 years from then the corporation's claim on the content would have expired.

So I'm keen to keep the two dates where the latter is not the "current" as you said rather is the date when the file was last edited.

>   The initial year can be extracted from git if needed.

We might not be on github/gitlab forever and if someone does a hard clone of the project in another repo then all git history of changes is lost.

battlmonstr:
@AndreaLanfranchi I understand what you want. This would mean a more complicated checking script logic (extracting and regexing the first line). Do you think it is worth spending time on it?

Note that those year update everywhere commits should be added to our `.git-blame-ignore-revs`, and so you can still figure out the last code modification date using git blame despite what the header says.

When we migrate to GitLab I think we should use `git clone`, not raw cp/xcopy to retain the history. It contains valuable information for development and maintenance.

AndreaLanfranchi:
> When we migrate to GitLab I think we should use git clone, not raw cp/xcopy to retain the history. It contains valuable information for development and maintenance.

I'm not concerned by "our" migrations to whatever: I'm concerned by "others" cloning the repo and loosing the history of initial content protecion. 

Besides this when I see PRs that change hundreds of files I'm mentally keen to think this should not go in, Period. Too much effort in trying to understand if any file for any reason has been changed beyond intentions. And this would happen once a year according to your model. When we want to automate everyhting we have several problems IMHO: devs are lazy and/or do not want to comply with code styling guides and, again, we add the burden to maintain automation tools.

Note these are only comments and will let others review.

AndreaLanfranchi:
Addition ... when automation to transform functions declarations from `bool foo()` to `foo() -> bool` ? (clang-tidy suggests it)

battlmonstr:
@AndreaLanfranchi a way to review this is to:

1. Skim through the diff to make sure that the changes only affect the copyright headers. See a separate commit with those: https://github.com/torquem-ch/silkworm/pull/725/commits/0ca8814bb8ee1655f321b752931d1093fe0166e2
2. Check that the script is doing the right thing. It is also in a separate commit: https://github.com/torquem-ch/silkworm/pull/725/commits/d6190c88617b2dd95f57908fb0577009638ebeea

The passing CI check ensures that it covered all the needed files.
You can also run the script locally by: `cmake -P cmake/copyright.cmake`

Note that this is based on https://github.com/torquem-ch/silkworm/pull/720 - it should be merged first.

canepat (approved):

PR #736, opened 2022-08-23
format: don't force long lines wrapping

battlmonstr:
Each developer decides on the appropriate line length.

* add a helper Makefile
* wrap some exceeding lines


battlmonstr:
@canepat this is a follow up from the formatting PR. It is manual work that needed to be done after auto-formatting. I couldn't do it there, because I didn't want to rebase it. I don't intend to fix all the exceptions here, only the ones that are easy to address by some line/comment readjustments. The remaining exceptions are mainly related to tests (which have some long hex strings), and GRPC template inheritance.

canepat:
I do support merging this PR. I think the current post-check on line length limit has proven more painful than useful.

The setup with just `ColumnLimit` set to `0` in `clang-format` resulting from this PR is IMHO good enough: in this way the author's attempt to make the code as clear and expressive as possible is keep safe and we can be reasonably certain that none of us will commit very long lines because no one likes to scroll horizontally to read his own code (any gross error can be easily caught during review).

yperbasis (approved):

PR #737, opened 2022-08-24
Stage History Index

AndreaLanfranchi:
## DO NOT MERGE YET PLEASE

- Rewritten with code reuse
- Amended unwind : instead of traversing whole history unwind only accounts/storage affected
- Tests
- New db::cursor_for_prefix

Several improvements


node/silkworm/db/stages.hpp
```diff
@@ -45,6 +45,9 @@ inline constexpr const char* kIntermediateHashesKey{"IntermediateHashes"};
 // Apply Keccak256 to all the keys in the state
 inline constexpr const char* kHashStateKey{"HashState"};
 
+// Generating history index for accounts
+inline constexpr const char* kHistoryIndexKey{"HistoryIndex"};
```
yperbasis, line 49: Why add "HistoryIndex"? There exist "AccountHistoryIndex" & "StorageHistoryIndex" already.

AndreaLanfranchi: Cause `HistoryIndex` is referred to whole stage which, in reality, is divided in two substages i.e. AccountHistoryIndex and StorageHistoryIndex. 
`HistoryIndex` is only updated when **both** substages are completed: as a result we can interrupt the stage and **IF** AccountHistory has already been committed, it restarts from StorageHistoryIndex instead of redoing everything.
Basically is relevant for initial from-scratch cycles when the stage duration is in hours.

AndreaLanfranchi: Imho is an improvement from Erigon where instead the whole history is committed all-or-nothing (both accounts and storage)

yperbasis: Perhaps then it's better to explicitly have two stages, AccountHistoryIndex and StorageHistoryIndex?

AndreaLanfranchi: That would require to have two separate stages classes and move out common code in a third shared class which does the actual work. Imho not worth ... but let you decide

yperbasis: Let's skip it for now and revisit later.

node/silkworm/stagedsync/stage_history_index.hpp
```diff
@@ -0,0 +1,58 @@
...
+    StageResult unwind_impl(db::RWTxn& txn, BlockNum from, BlockNum to, bool storage);
+    StageResult prune_impl(db::RWTxn& txn, BlockNum threshold, BlockNum to, bool storage);
+
+    //! \brief Collects bitmaps of block numbers changes for each account within provided
+    //! changeset boundaries
+    void collect_bitmaps_from_changeset(db::RWTxn& txn, const db::MapConfig& source_config, BlockNum from, BlockNum to,
+                                        bool storage);
+
+    //! \brief Collects unique keys touched by changesets within provided boundaries
+    std::unordered_map<Bytes, bool, boost::hash<Bytes>> collect_unique_keys_from_changeset(
```
yperbasis, line 52: Why `boost::hash`?

AndreaLanfranchi: Cause std::hash has no specialization supported for `basic_string<uint8_t>`

AndreaLanfranchi: Actually rethinking ... I inherited unordered_map from previous code but is better to use a map cause eventually keys are processed in order.

AndreaLanfranchi: Removed the need for boost::hash

AndreaLanfranchi:
Closing temporarily not to generate too much noise.

PR #739, opened 2022-08-29
Downloader - Give different names to two different states

mriccobene:
Now HeaderPersistence class gives different names to two different conditions.

The state "repaired" occurs when the previous "forward" run inserted new headers in the database but was unable to update the canonical chain due to an unwind condition detection. A subsequent "unwind" method execution reverted the canonical chain to the forking point. But the inserted headers are still there so the canonical chain has to be updated. HeaderPersistence detects this condition in the following forward execution and updates the canonical chain.


canepat (approved):
LGTM

PR #741, opened 2022-08-30
sentry: RLPx handshake framing and Hello message

battlmonstr:
* framing: secrets, MAC, encrypt/decrypt frames and messages
* Hello message exchange

Refactorings:
* rename session keys to auth keys
* refactor ECIES crypto to reuse in framing
* refactor AES to support more modes for framing


canepat (approved):

AndreaLanfranchi:
@canepat 
I'd add a comment to remove stbrume's keccak function when/if @chfast implements a _rolling_ keccak into ethash library.
Having two different implementations can be confusing.

canepat:
> @canepat I'd add a comment to remove stbrume's keccak function when/if @chfast implements a _rolling_ keccak into ethash library. Having two different implementations can be confusing.

Good point, @battlmonstr can you please add such comment in next PR?

PR #743, opened 2022-09-01
History and Log Indexes + TxLookup

AndreaLanfranchi:
Rewrite and plug into syncloop of :
- History Indexes
- Log Indexes
- TxLookup
- Finish

The first two stages have a lot in common (basically an index of roaring bitmaps) hence the implementation of `IndexLoader` in `db\bitmap.hpp`. 
The size of bitmap shards to insert in tables is now computed dynamically on behalf of MDBX's page size
History index has two sub-stages : one for Account and one for Storage


AndreaLanfranchi:
If you want we can wait another day and also tx lookup stage will be plugged in

yperbasis (approved):

PR #748, opened 2022-09-01
Downloader printable peer-id

mriccobene:
Erigon Sentry now uses pubkey as peer-id but pubkey has not printable characters.
This PR use Bytes in place of string to represent peer-id and use hex encoding to print peer-id.

### DIFF
--- a/node/silkworm/downloader/internals/types.hpp
+++ b/node/silkworm/downloader/internals/types.hpp
@@ -76,7 +76,12 @@ inline std::ostream& operator<<(std::ostream& out, const evmc::bytes32& b32) {
     return out;
 }
 
-using PeerId = std::string;
+using PeerId = Bytes;
+
+// Bytes already has operator<<, so PeerId but PeerId is too long
+inline ByteView human_readable_id(const PeerId& peer_id) {
+    return {peer_id.data(), std::min<size_t>(peer_id.length(), 20)};
+}
 
 enum Penalty : int {
     NoPenalty = 0,
--- a/node/silkworm/downloader/internals/types.hpp
+++ b/node/silkworm/downloader/internals/types.hpp
@@ -61,25 +61,9 @@ using duration_t = std::chrono::system_clock::duration;
 using seconds_t = std::chrono::seconds;
 using milliseconds_t = std::chrono::milliseconds;
 
-inline std::ostream& operator<<(std::ostream& out, const silkworm::ByteView& bytes) {
-    out << silkworm::to_hex(bytes);
-    return out;
-}
-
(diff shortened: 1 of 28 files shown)


node/silkworm/downloader/internals/types.hpp
```diff
@@ -76,7 +76,12 @@ inline std::ostream& operator<<(std::ostream& out, const evmc::bytes32& b32) {
     return out;
 }
 
-using PeerId = std::string;
+using PeerId = Bytes;
+
+// Bytes already has operator<<, so PeerId but PeerId is too long
```
AndreaLanfranchi, line 81: This is a bit misleading. Actually Bytes/ByteView do not have any operator `<<` defined in std lib. They're custom implemented a few lines above.
I suggest moving those custom definitions in core\common or core\base so they're widely usable otherwise is a feature of compilation units which import downloader types only

AndreaLanfranchi: Also "human_readable" is simply a constructor of a byteview limited to a max of 20 chars.
Is not readable per se and we "hide" a constant cut

mriccobene: > I suggest moving those custom definitions in core\

It is a good observation however who uses it besides the donwloader? Sentry uses it but for it Peer is a class with a member like this:

`std::optional<common::EccPublicKey> peer_public_key_`

Whereas for the downloader it is just an opaque id.
It seems premature to me to move it to a more central place.

> Is not readable per se and we "hide" a constant cut

I want to put in one place in the code the decision about the best way to display to a user that number. Right now I chose to display the first 20 characters but maybe tomorrow we choose to do it differently, maybe following the choices that Sentry will make. 

Instead, returning a ByteView is dangerous, and I wonder if it is worth doing.

AndreaLanfranchi: > It is a good observation however who uses it besides the donwloader? Sentry uses it but for it Peer is a class with a member like this:

I was referring to the `<<` operator. We have tons of places where we hex encode a Bytes or a ByteView before printouts or append to strings.

mriccobene: Ah, sorry, I agree! thanks!

mriccobene: I reverted the `<< operator` move because gcc-10 finds an ambiguity. 
We will evaluate in future to make the overload on ByteView tighter (maybe on BufferBase?) to not break serialisation code of string_view.

canepat (approved):

PR #754, opened 2022-09-05
Downloader IStage conformance

mriccobene:
This PR adapts stage_headers and stage_bodies to IStage interface to prepare their integration in the stage-loop.


cmd/downloader.cpp
```diff
@@ -178,11 +198,25 @@ int main(int argc, char* argv[]) {
...
+        BodiesStage body_stage{shared_status, block_exchange, &node_settings};
+
+        // Trap os signals
+        SignalHandler::init([&](int) {
+            log::Info() << "Requesting threads termination\n";
+            header_stage.stop();
+            body_stage.stop();
+            block_exchange.stop();
+            sentry.stop();
+        });
```
AndreaLanfranchi, line 211: Why a custom handler for `SignalHandler::init()` ?
It's risky: if in another place we set another handler this one is forgotten.
Better to change in `Stoppable` the method `is_stopping` and check there `SignalHandler::signalled()`

node/silkworm/concurrency/signal_handler.cpp
```diff
@@ -103,11 +103,13 @@ inline constexpr int kHandleableCodes[] {
...
 std::atomic_int SignalHandler::sig_code_{0};
 std::atomic_bool SignalHandler::signalled_{false};
+std::function<void(int)> SignalHandler::custom_handler_;
 
-void SignalHandler::init() {
+void SignalHandler::init(std::function<void(int)> custom_handler) {
     for (const int sig_code : kHandleableCodes) {
         signal(sig_code, &SignalHandler::handle);
     }
+    custom_handler_ = custom_handler;
```
AndreaLanfranchi, line 112: See comment above

node/silkworm/concurrency/signal_handler.cpp
```diff
@@ -130,6 +132,9 @@ void SignalHandler::handle(int sig_code) {
         digit_with_endl[2] = '\0';
         std::fputs(digit_with_endl, stderr);
     }
+    if (custom_handler_) {
+        custom_handler_(sig_code);
+    }
```
AndreaLanfranchi, line 137: See comment above

cmd/downloader.cpp
mriccobene: I have a mixed feeling about this. I think SignalHandler::init() must be called only one time. Anyway I made a commit to follow your suggestion.

AndreaLanfranchi: Yes should be called only once at silkworm start. And this happens way before pointers to stages are available. 
Downloader binary is actually way more simple than that

canepat (approved):

PR #763, opened 2022-09-14
Simplify signature of WalkFunc

yperbasis:
Change
```
using WalkFunc = std::function<bool(::mdbx::cursor& cursor, ::mdbx::cursor::move_result& data)>
```
to
```
using WalkFuncRef = absl::FunctionRef<void(ByteView key, ByteView value)>
```

Rationale:
1. The early termination feature was unused, so no longer `bool` return.
2. The new function signature is simpler.
3. `absl::FunctionRef` is preferable to `const std::function&`: see C++26 proposal [P0792](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p0792r10.html) for motivation.

### DIFF
--- a/cmd/toolbox.cpp
+++ b/cmd/toolbox.cpp
@@ -263,7 +263,7 @@ dbTablesInfo get_tablesInfo(::mdbx::txn& txn) {
     ret.size += table->size();
     ret.tables.push_back(*table);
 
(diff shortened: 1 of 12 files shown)


AndreaLanfranchi (approved):
Good for me

PR #765, opened 2022-09-15
Merge ChainIdentity into ChainConfig

AndreaLanfranchi:
Actually `ChainIdentity` is totally apart from `ChainConfig` and moves around a copy of ChainConfig which is derived from hard-coded configs and not, like silkworm does, derived from what persisted into db.
This causes ChainIdentity would never work with custom chains.
Long story short ... there must be only one chain config referred around : the one retrieved from genesis initialization.


mriccobene (commented):
I think we should keep different names for different things.
Also it is useful knowing that a function takes as input a ChainIdentity only or a full ChainConfig.
Your suggestion is valid IMO, but I would implement it in reverse, putting ChainIdentity inside ChainConfig.

AndreaLanfranchi:
@mriccobene 
I honestly don't understand this.

> Your suggestion is valid IMO, but I would implement it in reverse, putting ChainIdentity inside ChainConfig.

What do you mean by "reverse". The Identity is derived from ChainConfig so it's already inside it.

AndreaLanfranchi:
> Also it is useful knowing that a function takes as input a ChainIdentity only or a full ChainConfig.

ChainIdentity as it was implemented is a superclass of ChainConfig holding a full copy of it. 
So what ?

canepat (approved):

PR #769, opened 2022-09-19
Fix flaky log tests

canepat:
This PR fixes a couple of failures in `node_test` when executed within CLion IDE that shuffles the execution order of tests.

The problem shows in `LogBuffer stores content for verbosity lower than or equal to default` test but it is indeed caused by other tests which change the default log verbosity _without_ restoring the default value at the end of the test.


mriccobene (approved):
LGTM

PR #771, opened 2022-09-27
Update pre-verified hashes for mainnet

canepat:
Update pre-verified hashes for main network up to block 15'537'376


mriccobene (approved):
LGTM

Issue #777, opened 2022-10-03
Worker termination may leave it in Stopping state


canepat:
A race condition between the `Worker::work` method completion (e.g. caused by an exception) and the `Worker::stop` method called by an external thread may leave the `Worker` in state `Stopping` (as opposed to the expected `Stopped` state).

AndreaLanfranchi:
Yes confirm.
I believe Worker::Stop should be amended as:

```
void Worker::stop(bool wait) {
    if (!thread_) return;
    
    State expected{State::kStarted);
    if(state_.compare_exchange_strong(expected, State::kStopping)) {
        kick();
    }

    if (wait) {
        thread_->join();
        thread_.reset();
    }
}
```

Hope it helps

PR #778, opened 2022-10-03
Fix stop sequence for Silkworm executable

canepat:
This PR fixes three conditions preventing Silkworm to execute a graceful termination:
- the sync-loop `Worker` sometimes remains stuck into the main loop because of a race condition in Worker thread termination that can leave it in the `Stopping` state instead of the expected `Stopped` state. Given that such loop is currently used only in `main`, adding an additional check temporarily solves the issue (#777 opened to track the problem)
- sync Sentry client remains stuck waiting for a connection in case the Sentry component is not reachable, thus preventing a clean shutdown
- sync Sentry client may remain stuck executing the `handshake` with the Sentry component, thus preventing a clean shutdown


mriccobene (approved):
LGTM

PR #779, opened 2022-10-04
Fix Sentry client shutdown hang in PeerCount RPC

canepat:
This PR fixes an hang in sync Sentry client during the shutdown sequence of Silkworm. The issue is reproducible with 30% frequency in the following scenario:
- start Silkworm w/ Sentry down
- start Sentry and let Silkworm connect to it
- stop Sentry
- try to stop Silkworm sending Ctrl-C signal

The problem can be tracked down to the blocking `PeerCount` call.


mriccobene (approved):
LGTM

PR #780, opened 2022-10-04
Fix race condition in Worker termination state

### LINKED ISSUE #777: Worker termination may leave it in Stopping state

canepat:
This PR hopefully fixes a race condition in `Worker` termination state occurring between:
- the worker thread itself updating the state to `Stopped` before terminating
- any external thread calling `Worker::stop` method to stop the worker updating the state to `Stopping`

Fixes #777


mriccobene (approved):
I would have removed the kStopped state

PR #781, opened 2022-10-04
Integrate BackEndKV server in Silkworm core component

canepat:
This PR integrates the BackEnd & KV gRPC server into Silkworm core component, also refactoring how some common options are handled. In particular:
- introduce some functions to add a single command-line option (e.g. `add_option_private_api_address `)
- support both tag and number values for enumerated options (e.g. `--log.verbosity 4` or `--log.verbosity info`)
- move function to build the node name in common place
- add BackEndKvServer startup to Silkworm main
- refactor command-line parsing in standalone Downloader executable

### DIFF
--- a/cmd/backend_kv_server.cpp
+++ b/cmd/backend_kv_server.cpp
@@ -14,8 +14,8 @@
    limitations under the License.
 */
 
+#include <filesystem>
 #include <string>
-#include <thread>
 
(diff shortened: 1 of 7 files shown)


mriccobene (approved):
ok but I had to add a fix to avoid an MSVC compile error related to CLI library `details::to_string()`

PR #783, opened 2022-10-05
Add support for reading TTD as JSON number (Erigon compatibility)

canepat:
Erigon 2 treats Terminal Total Difficulty (TTD) field of [chain configuration](https://github.com/torquem-ch/silkworm/blob/master/core/silkworm/chain/config.cpp) as JSON number, whilst Silkworm currently treats it as JSON string.

This PR aims at keeping read-only compatibility from Erigon database by adding support for reading TTD both as JSON string and JSON number (switching to JSON number format only for Silkworm is not ).

This feature is also needed by Silkrpc for interoperability testing during development using Erigon core component.

### DIFF
--- a/core/silkworm/common/util.hpp
+++ b/core/silkworm/common/util.hpp
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <cmath>
 #include <cstring>
 #include <optional>
 #include <string_view>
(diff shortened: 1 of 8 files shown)


mriccobene (approved):
Ok. I have only some observation. 
The mantissa of a double on c++ contains only 15 digits, so beyond those the result will be wrong but currently they are enough. 
Also you use intx to raise an exception and not a more generic method from the Int template parameter.

core/silkworm/common/util.hpp
```diff
@@ -82,4 +83,51 @@ size_t prefix_length(ByteView a, ByteView b);
 
 inline ethash::hash256 keccak256(ByteView view) { return ethash::keccak256(view.data(), view.size()); }
 
+//! \brief Create an intx::uint256 from a string supporting both fixed decimal and scientific notation
+template <typename Int>
+inline constexpr Int from_string_sci(const char* str) {
```
yperbasis, line 88: Better pass as `std::string_view` for generality

canepat: Right, done in commit https://github.com/torquem-ch/silkworm/pull/783/commits/040e804b018850386886a780545acd61522d4295

PR #784, opened 2022-10-06
Add creation completed hook to EVM tracer

canepat:
This PR introduces a new `on_creation_completed` hook in EVM tracer.

Such hook is necessary because in `EVM::create` we modify the `evmc_result` returned by `evmone::execute` and thus we need to give the registered tracers a chance to see such changes.


mriccobene (approved):
ok

PR #787, opened 2022-10-17
Add memory-mapped file implementation

canepat:
This PR adds a thin memory-mapped file implementation, based on [Aeron's one](https://github.com/real-logic/aeron/blob/master/aeron-client/src/main/cpp/util/MemoryMappedFile.h), as part of the Snapshot Sync functionality. The motivation for introducing this memory-mapped file lies on the performance numbers captured in [its benchmark file](https://github.com/torquem-ch/silkworm/blob/105735fe7512ffc5fa8ffca9eb1358e2458e54ea/node/silkworm/common/memory_mapped_file_benchmark.cpp).

Exactly for the purpose of adding performance benchmarks, this PR includes also the following changes:

- add Google Benchmark as project dependency
- add new `benchmark_test` target built collecting all the benchmark files (i.e. any source code file matching *_benchmark.cpp)

Minor customization in Hunter configuration: it is necessary not to treat warnings as errors in Google Benchmark build to avoid bug in GCC 12.1.0:

- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,7 +195,14 @@ endif()
 
 # Compiler enforcements on Silkworm
 if(NOT MSVC)
-  add_compile_options(-Werror -Wall -Wextra -pedantic)
+  if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU") AND (CMAKE_CXX_COMPILER_VERSION STREQUAL "12.1.0"))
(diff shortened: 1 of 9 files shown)


mriccobene (approved):
LGTM

PR #793, opened 2022-10-23
Refactor BackEndKV server to use asio-grpc

canepat:


node/silkworm/rpc/server/server_context_pool.hpp
```diff
@@ -49,23 +48,32 @@ class ServerContext {
...
-    std::unique_ptr<CompletionEndPoint> server_end_point_;
-    std::unique_ptr<grpc::CompletionQueue> client_queue_;
-    std::unique_ptr<CompletionEndPoint> client_end_point_;
+    //! The asio-grpc asynchronous event schedulers.
+    std::unique_ptr<agrpc::GrpcContext> server_grpc_context_;
+    std::unique_ptr<agrpc::GrpcContext> client_grpc_context_;
+
+    //! The work-tracking executors that keep the asio-grpc scheduler running.
+    boost::asio::executor_work_guard<agrpc::GrpcContext::executor_type> server_grpc_context_work_;
+    boost::asio::executor_work_guard<agrpc::GrpcContext::executor_type> client_grpc_context_work_;
```
mriccobene, line 76: This code suggests that perhaps there is another abstraction hidden in the code, an abstraction with an agrpc::GrpcContext and an executor_work_guard. Maybe a ConnectionEndPoint?
In any case we can think about it in the future, let's leave it as it is, since it works.

node/silkworm/rpc/server/backend_calls.cpp
```diff
@@ -36,28 +47,19 @@ void EtherbaseCall::fill_predefined_reply(const EthereumBackEnd& backend) {
...
 }
 
-EtherbaseCall::EtherbaseCall(boost::asio::io_context& scheduler, remote::ETHBACKEND::AsyncService* service, grpc::ServerCompletionQueue* queue, Handlers handlers)
-    : UnaryRpc<remote::ETHBACKEND::AsyncService, remote::EtherbaseRequest, remote::EtherbaseReply>(scheduler, service, queue, handlers) {
-}
-
-void EtherbaseCall::process(const remote::EtherbaseRequest* request) {
-    SILK_TRACE << "EtherbaseCall::process START request: " << request;
-
+awaitable<void> EtherbaseCall::operator()(grpc::ServerContext& /*server_context*/, remote::EtherbaseRequest& /*request*/,
```
mriccobene, line 50: There is an unfortunate clash between `grpc::ServerContext `and `silkworm::rpc::ServerContext`. 
Maybe a different name for the latter will help the reader.

mriccobene (approved):

node/silkworm/rpc/server/server_context_pool.hpp
canepat: OK, let's see how this works for now, IMHO it's not a big deal

node/silkworm/rpc/server/backend_calls.cpp
canepat: Yep, you're definitely right. To be completely honest, it's even worse:
- `grpc::ServerContext` from `gRPC` library
- `agrpc::GrpcContext` from `asio-grpc` library
- `boost::asio::io_context` from `Boost Asio` library
- `silkworm::rpc::ServerContext` from our own side

On the other hand, the name used in such libraries demonstrates somehow a well-established way to identify execution loops.

I will try to use a better name for our class in a future PR. Tentative proposal: `ExecutionUnit` or `ExecutionLoop` or just `Context` (yes, I understand this seems not to solve the problem at all)

mriccobene (approved):

PR #794, opened 2022-10-24
Fix mmap build issue on macOS

canepat:
This PR fixes a build issue about mmap flags on macOS platform


yperbasis (approved):

mriccobene (approved):
Ok, it also compiles on Windows. And the tests pass.

PR #799, opened 2022-10-30
Add max idle timeout in Tx RPC calls

canepat:
This PR restores the max idle timeout handling in Tx RPC calls.

Moreover, it also introduces some related changes:
- switch back to `boost::asio` timers in Tx RPC calls
- use `boost::asio::dispatch` instead of `boost::asio::post` in StateChanges RPC notification handler
- avoid stopping `asio-grpc` server context on shutdown to make all handlers complete (this can slow down the shutdown in case of heavy load; ideally we could parameterise `rpc::Server::shutdown` and use `GrpcContext::stop` to have immediate shutdown, but let's keep it simple for now)
- restore temporarily disabled BackEndKV unit tests
- avoid completion queue drain in test client

### DIFF
--- a/cmd/test/backend_kv_test.cpp
+++ b/cmd/test/backend_kv_test.cpp
@@ -1016,10 +1016,7 @@ int main(int argc, char* argv[]) {
                     SILK_DEBUG << "Completion thread post operation: " << processor;
                     (*processor)(ok);
                 } else {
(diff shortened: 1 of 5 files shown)


mriccobene (approved):

mriccobene (approved):
Discussed in call

PR #803, opened 2022-11-08
Downloader small optimization

mriccobene:
Here we avoid fetching parent header always even when there is no forking point.


canepat (approved):
LGTM

PR #817, opened 2022-12-19
Change --channels parameter default value

canepat:
This PR change the default value for the `channels` parameter of the BackEnd&KV test client because at least one channel is necessary.


wezrule:
Probably don't need this anymore then https://github.com/torquem-ch/silkworm/pull/815 as it forces `--channels` to be specified because it was 0 by default.

mriccobene (approved):

canepat:
> Probably don't need my PR anymore then #815 as it forces `--channels` to be specified because it was 0 by default.

It has already been merged because it is needed when you specify the value on command-line.

PR #827, opened 2023-01-04
sentry: use asio grpc

battlmonstr:


battlmonstr:
It is nice that I can pass ServiceState to calls now and avoid a global variable.
I was confused by how `request_` type is inferred now from the template argument. In theory we can pass it as an `operator()` parameter to make its type explicit like before.

canepat:
> I was confused by how `request_` type is inferred now from the template argument. In theory we can pass it as an `operator()` parameter to make its type explicit like before.

This is a good point for `UnaryCall` and `ServerStreamingCall`, I suggest to do this in a separate PR.

PR #829, opened 2023-01-09
sentry: refactor common RPC type conversions

battlmonstr:
Sentry is using parts of downloader's sentry_type_casts.hpp which implements interfaces' types.proto helpers.
Those could be useful for other RPCs too.


battlmonstr:
I have only moved parts that are used from 2 places (server and client), because I'm not sure that everything is going to be needed.

rpc/interfaces/types.hpp is meant to cover https://github.com/ledgerwatch/interfaces/blob/master/types/types.proto

In the future for commonalities that are specific to sentry, and are needed in both the sentry client (downloader) and the sentry server, the plan is to make a separate folder such as node/silkworm/sentry .

mriccobene (approved):

PR #830, opened 2023-01-10
sentry: send message

battlmonstr:


sentry/silkworm/sentry/rlpx/server.cpp
```diff
@@ -99,4 +111,24 @@ awaitable<void> Server::start(
...
+}
+
+awaitable<void> Server::enumerate_random_peers(size_t max_count, std::function<awaitable<void>(Peer&)> callback) {
+    co_await co_spawn(strand_, enumerate_random_peers_in_strand(max_count, callback), use_awaitable);
+}
+
+awaitable<void> Server::enumerate_peers_in_strand(std::function<awaitable<void>(Peer&)> callback) {
+    for (auto& peer : peers_) {
+        co_await callback(*peer.first);
+    }
```
battlmonstr, line 125: @canepat I'm not sure that the strands fully help here. It is not reentrant: when waiting on co_await, the strand is free to execute other queued work. In particular, it can disconnect a peer, modify peers_, and this ruins the loop iteration. I'm looking for a way to improve this, maybe copying the pointers or peer keys and then double-checking that it's still present in peers_ before awaiting?

Also exception handling needs to be implemented. E.g. if a peer dies, it's channel is closed, so here co_await would end with an exception and bubble up to ruin the whole MessageSender::start loop.

battlmonstr: I've refactored it to a safer approach.

canepat (approved):

PR #837, opened 2023-01-19
Improve DecodingResult error handling

yperbasis:


core/silkworm/common/decoding_result.hpp
```diff
@@ -35,7 +35,10 @@ enum class [[nodiscard]] DecodingResult{
     kUnexpectedEip2718Serialization,
     kInvalidHashesLength,  // trie::Node decoding
     kInvalidMasksSubsets,  // trie::Node decoding
-
 };
 
+[[nodiscard]] inline bool error(DecodingResult x) {
```
battlmonstr, line 40:
```suggestion
[[nodiscard]] inline bool is_error(DecodingResult x) {
```
maybe?
error() feels like a constructor to me.

You have added this generic name to the global namespace.
Does it mean you want to continue with this approach, make it polymorphic and implement is_error for different types?
IMO this file should be inside the `silkworm::rlp` namespace.

Consider also an alternative approach:
https://stackoverflow.com/a/53284026/1009546

This is more OOP-style, converting DecodingResult to a class with an embedded enum, and then implementing operator bool on that to allow for code like this:

    if (!res) {
    
Yet another option, Go-style, is to rename DecodingResult enum to DecodingErrorCode, and return `std::optional<DecodingErrorCode>`.

battlmonstr (approved):

yperbasis:
Closing in favour of PR #864.

PR #855, opened 2023-02-01
sentry: replace detached tasks with TaskGroup

battlmonstr:


sentry/silkworm/sentry/common/task_group.cpp
```diff
@@ -0,0 +1,125 @@
...
+        throw SpawnAfterCloseError();
+    }
+
+    auto task_id = ++last_task_id_;
+
+    auto [it, ok] = tasks_.emplace(
+        std::piecewise_construct,
+        std::forward_as_tuple(task_id),
+        std::forward_as_tuple());
+    assert(ok);
```
canepat, line 48: Just in case you need it, you can use `SILKWORM_ASSERT` instead of `assert` to have the condition checked also in release builds

canepat (approved):

PR #860, opened 2023-02-06
copyright year update

battlmonstr:


yperbasis:
Hmm. Should we really change copyright year in all files every year? As far as I can tell, setting copyright year to the year a file was created once and not changing ever again it is fine.

I've made a minimal change in PR #837 to accept 2023 alongside 2022. Replace
```
    if(NOT (header STREQUAL COPYRIGHT_HEADER))
```
with
```
    string(REPLACE "2022" "2023" COPYRIGHT_HEADER_23 "${COPYRIGHT_HEADER}")

    if(NOT ((header STREQUAL COPYRIGHT_HEADER) OR (header STREQUAL COPYRIGHT_HEADER_23)))
```

battlmonstr:
@yperbasis 👌 your suggestion is sensible, let's drop this.

I thought that it needs to be the current year, and needs to match with the product license, but after reading this: https://stackoverflow.com/a/20911485/1009546 I feel it makes little to no sense.

PR #885, opened 2023-02-15
node: fix downloader queues & tests

mriccobene:


battlmonstr (approved):
What's the reason for changing heap based to set based prio-q?

mriccobene:
> What's the reason for changing heap based to set based prio-q?
My heap based queue implementation was flawed for 2 reasons: 1) it accessed an internal member of the container, 2) the underling heap are not assured to be in order if you iterate over it. And I need to iterate over it to count expired timers (to compute how many requests can I make to peers)

PR #886, opened 2023-02-16
sentry: static peers discovery

battlmonstr:


battlmonstr:
I'm using ServerContextPool::next_io_context for each new peer so that their handling is distributed more evenly between contexts. Server peers and client peers are created in different io_contexts, so potentially next_io_context might be called from different threads. So I've added a bit of thread-safety there. Assuming that add_context/start/join are called from the same initial thread and we don't need synchronization of contexts_, stop() and next_io_context() are safe now to call from multiple threads.

canepat (approved):

PR #892, opened 2023-02-22
sentry: peer management GRPC API

battlmonstr:


silkworm/sentry/common/promise.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+    void set_value(T value) {
+        channel_.try_send(std::move(value));
+    }
+
+  private:
+    AtomicValue<std::optional<T>> value_{std::nullopt};
+    Channel<T> channel_;
+    std::mutex mutex_;
+};
```
canepat, line 69: I like it, maybe we can try to come up with a different name to clearly distinguish it from Asio `experimental::promise`

battlmonstr: Do you think we are going to use experimental::promise somewhere?

canepat: TBH I don't have a specific use case in mind, it's just a general consideration. We can keep `sentry::common::Promise` as it is until we need to use `experimental::promise` in one real case (if any) and re-evaluate at that time.

PR #914, opened 2023-03-16
rpcdaemon: add ots_getBlockDetails and ots_getBlockDetailsByHash

dario1995:


silkworm/silkrpc/commands/ots_api.cpp
```diff
@@ -77,4 +79,105 @@ boost::asio::awaitable<void> OtsRpcApi::handle_ots_has_code(const nlohmann::json
...
+        if (block_body_rlp.empty()) {
+            throw std::runtime_error{"empty block body RLP in read_body"};
+        }
+
+        silkworm::ByteView data_view{block_body_rlp};
+        auto stored_body{silkworm::db::detail::decode_stored_block_body(data_view)};
+
+        const BlockDetails block_details{block_hash, block_header, total_difficulty, stored_body.txn_count -2, stored_body.ommers};
+
+        const BlockDetailsResponse block_details_response{block_details};
```
canepat, line 165: Two fields in `BlockDetailsResponse` are missing here:

- `issuance`: it should filled using `ethash::compute_reward`
- `total_fees`: it should be filled using `core::get_receipts`

However, there is more work to do on these subjects:
- regarding the first item, we should more generally add `compute_reward` method to `consensus::Engine` class hierarchy and let each consensus algorithm implement its own block reward
- regarding the second one, we should complete the implementation of `core::get_receipts` for the case when receipts are not present on the data store and must be recomputed by executing the call

These evolutions are not required for this PR, you can do them on next PR.

canepat (changes requested):
It would be great to add the integration tests as well

canepat (approved):

PR #927, opened 2023-03-23
sentry: reconnecting GRPC client with retries

battlmonstr:
* add sentry_client_test cmd to test the client API
* refactor unary_rpc to return replies
* wrap unary_rpc/streaming_rpc with reconnect/retry logic

### DIFF
--- a/silkworm/node/rpc/client/call.cpp
+++ b/silkworm/node/rpc/client/call.cpp
@@ -1,0 +1,48 @@
+/*
+   Copyright 2023 The Silkworm Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#include "call.hpp"
+
+#include <silkworm/node/concurrency/async_thread.hpp>
+
+namespace silkworm::rpc::detail {
+
+bool is_disconnect_error(const GrpcStatusError& ex, grpc::Channel& channel) {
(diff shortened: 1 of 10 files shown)


cmd/CMakeLists.txt
```diff
@@ -74,6 +74,9 @@ if(NOT SILKWORM_CORE_ONLY)
   add_executable(sentry "sentry.cpp;${COMMON_SRC}")
   target_link_libraries(sentry PRIVATE silkworm_node silkworm-buildinfo CLI11::CLI11 silkworm_sentry)
 
+  add_executable(sentry_client_test "sentry_client_test.cpp")
+  target_link_libraries(sentry_client_test PRIVATE silkworm_sentry)
+
```
canepat, line 79: I suggest to move this `sentry_client_test` in the `cmd/test` folder where other test clients are already present (e.g. `backend_kv_test`)

silkworm/node/rpc/client/call.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+boost::asio::awaitable<void> reconnect_channel(grpc::Channel& channel) {
+    bool is_stopped = false;
+
+    std::function<void()> run = [&] {
+        bool is_connected = false;
+        while (!is_connected && !is_stopped && (channel.GetState(false) != GRPC_CHANNEL_SHUTDOWN)) {
+            log::Info() << "Reconnecting grpc::Channel...";
+            auto deadline = gpr_time_add(gpr_now(GPR_CLOCK_REALTIME), gpr_time_from_seconds(5, GPR_TIMESPAN));
```
canepat, line 36: I think we could make this hard-coded 5sec a retry interval parameter that can be specified at the calling site in `unary_rpc` and `streaming_rpc`. It's not necessary to do it in this PR, though.

canepat (changes requested):

cmd/CMakeLists.txt
battlmonstr: 👍

silkworm/node/rpc/client/call.cpp
battlmonstr: sure, let's do it later; we should probably have a back-off strategy

canepat (approved):

PR #941, opened 2023-03-29
build: intx & ethash as submodules rather than Hunter packages

yperbasis:
With this we don't need Hunter anymore in the Conan mode.


chfast (approved):
Looks good to me provided everything builds fine.

PR #960, opened 2023-04-05
sentry: embedded with direct client

battlmonstr:
The embedded client is setup in silkworm cmd if --sentry.remote.addr is not set.

* refactor silkworm cmd main to awaitable tasks
* move awaitable_wait_for_x to infra
* make node depend on sentry (for the future, and to verify the absence of a crash)


canepat:
Before merging I've done some work:

- a bunch of commits to fix compilation errors in Windows build (in the end it was necessary to skip `silkworm` target anyway because of `fatal error C1002: compiler is out of heap space in pass 2`)
- a couple of commits to fix a runtime error in `silkworm` executable caused by MDBX for r/w transaction threading issue

canepat (approved):

PR #963, opened 2023-04-05
rpcdaemon:  replace to_evmc_address with std::memcpy in cbor decode

lupin012:


canepat (approved):
I suggest inserting checks to protect against negative input values for `size` parameters both in `on_bytes` and `on_array`. Next PR is OK

PR #983, opened 2023-04-13
sync: use SentryClient from the sentry library

battlmonstr:
* refactor SentryClient to be based on the client from the silkworm_sentry library instead of its own
* keep SentryClient synchronous using use_future
* refactor SentryClient to only expose sync types in the public interface
* inbound message factory moved to SentryClient so that InboundMessage doesn't depend on sentry message type
* head_block_num added to set_status
* clean up GRPC usage
* conventions and clang-tidy fixes


canepat:
LGTM

@mriccobene please take a look because this changes the sync module

mriccobene:
I am looking at it but I need time: the changes are somewhat unexpected.
I had imagined one could rewrite my SentryClient on the asynchronous one so that mine had no threads and its exec_remote() would return a future to wait on.
Instead, the threads are still here, and there are other profound changes that I need to understand thoroughly in both the reasons and the implementation.

silkworm/sync/sentry_client.cpp
```diff
@@ -16,137 +16,300 @@
...
-
-    SILK_TRACE << "SentryClient, hand_shake sent";
-    sentry::HandShakeReply reply = handshake_->reply();
+void SentryClient::handshake() {
+    auto supported_protocol = sync_spawn(tasks_, io_context_, sentry_client_->service()->handshake());
 
-    sentry::Protocol supported_protocol = reply.protocol();
-    if (supported_protocol < sentry::Protocol::ETH66) {
-        log::Critical("SentryClient") << "remote sentry do not support eth/66 protocol, stopping...";
+    if (supported_protocol < 66) {
```
mriccobene, line 207: Why a nude number? I like numbers when constants do not add meaning. For example, I prefer `object.timeout(1s)` to a `object.timeout(kSomeNameTimeout)`. But here using `sentry::Protocol::ETH66` can lead our code to handle smoothly a change on the sentry::Protocol::ETH66 constant.

silkworm/sync/sentry_client.cpp
```diff
@@ -16,137 +16,300 @@
...
+        peer_ids.push_back(peer_key.serialized());
+    }
+    return peer_ids;
+}
+
+SentryClient::PeerIds SentryClient::send_message_by_id(const OutboundMessage& outbound_message, const PeerId& peer_id) {
+    auto message = sentry_message_from_outbound_message(outbound_message);
+    auto peer_public_key = sentry::common::EccPublicKey::deserialize(peer_id);
+    auto peer_keys = sync_spawn(tasks_, io_context_, sentry_client_->service()->send_message_by_id(std::move(message), std::move(peer_public_key)));
+    return peer_ids_from_peer_keys(peer_keys);
```
mriccobene, line 233: Copying peer ids and returning them is a performance penalty compared to the old implementation. Also, there can be cases where they are not needed at all.
We agreed on the call that my concern is premature optimization, so I am writing this note to record our decision.

silkworm/sync/sentry_client.cpp
```diff
@@ -16,137 +16,300 @@
...
 
-                publish(message);
-            }
+            std::function<awaitable<void>(silkworm::sentry::api::api_common::MessageFromPeer)> consumer = [this](auto message_from_peer) -> awaitable<void> {
+                if (!this->is_stopping()) {
+                    co_await this->publish(message_from_peer);
+                }
+            };
+
+            sync_spawn(tasks_, io_context_, sentry_client_->service()->messages(make_message_id_filter(), consumer));
```
mriccobene, line 300: I would like to rename the sentry client's `messages()` method to `receive_messages()` : with this modification, this code will be more readable.

silkworm/sync/sentry_client.cpp
```diff
@@ -16,137 +16,300 @@
...
+
     while (!is_stopping()) {
         try {
             connected_ = false;
-            log::Info("SentryClient", {"remote", sentry_addr_}) << " connecting ...";
+            log::Info(kLogTitle) << "connecting ...";
 
             // send current status of the chain
-            hand_shake();
+            handshake();
```
mriccobene, line 285: IMO we need to remove the re-connection logic (plus handshake & set_status) from here if, as it seems to me, other parts of the application will use the async client.

silkworm/sync/sentry_client.cpp
```diff
@@ -16,137 +16,300 @@
...
+    auto peer_keys = sync_spawn(tasks_, io_context_, sentry_client_->service()->send_message_by_min_block(std::move(message), max_peers));
+    return peer_ids_from_peer_keys(peer_keys);
+}
+
+void SentryClient::peer_min_block(const PeerId& peer_id, BlockNum /*min_block*/) {
+    auto peer_public_key = sentry::common::EccPublicKey::deserialize(peer_id);
+    sync_spawn(tasks_, io_context_, sentry_client_->service()->peer_min_block(std::move(peer_public_key)));
+}
 
+void SentryClient::execution_loop() {
```
mriccobene, line 259: I would like to get rid of this loop. SentryClient could expose a method like "receive_messages(callback)" that is now possible having an async client. This method will spawn a coroutine that when a message is received will send it to subscribers.

silkworm/sync/sentry_client.cpp
```diff
@@ -158,89 +321,83 @@ void SentryClient::stats_receiving_loop() {
...
-            log::Info("SentryClient") << active_peers_ << " active peers";
+            log::Info(kLogTitle) << count_active_peers() << " active peers";
 
             // receive stats
-            while (!is_stopping() && receive_peer_stats_->receive_one_reply()) {
-                const sentry::PeerEvent& stat = receive_peer_stats_->reply();
 
-                auto peerId = bytes_from_H512(stat.peer_id());
-                const char* event = "";
+            std::function<awaitable<void>(silkworm::sentry::api::api_common::PeerEvent)> consumer = [this, &peer_infos](auto stat) -> awaitable<void> {
```
mriccobene, line 329: As above, I would like to get rid of this loop. Is it possible?

silkworm/sync/sentry_client.cpp
```diff
@@ -16,137 +16,300 @@
...
+SentryClient::PeerIds SentryClient::send_message_by_id(const OutboundMessage& outbound_message, const PeerId& peer_id) {
+    auto message = sentry_message_from_outbound_message(outbound_message);
+    auto peer_public_key = sentry::common::EccPublicKey::deserialize(peer_id);
+    auto peer_keys = sync_spawn(tasks_, io_context_, sentry_client_->service()->send_message_by_id(std::move(message), std::move(peer_public_key)));
+    return peer_ids_from_peer_keys(peer_keys);
+}
+
+SentryClient::PeerIds SentryClient::send_message_to_random_peers(const OutboundMessage& outbound_message, size_t max_peers) {
+    auto message = sentry_message_from_outbound_message(outbound_message);
+    auto peer_keys = sync_spawn(tasks_, io_context_, sentry_client_->service()->send_message_to_random_peers(std::move(message), max_peers));
```
mriccobene, line 238: I need clarification on the class TaskGroup. What are its responsibilities? why it is used here?

mriccobene (approved):

silkworm/sync/sentry_client.cpp
battlmonstr: Good point. I will add it to my TODO.

silkworm/sync/sentry_client.cpp
battlmonstr: I understand. The idea was that the names exactly match to sentry.proto, but I see how it is confusing having this callback style of subscribing. I wish I could just return a lazy "stream" of messages, but it was tricker than I thought.

Thinking about a different name, we have another method like "node_info", but we are not calling it "receive_node_info". Maybe "wait_for_messages" or "on_messages"? I need to think a bit more before renaming this. Same for peer_events e.g. "on_peer_events" or "subscribe_to_peer_events_and_wait" 🤔

silkworm/sync/sentry_client.cpp
battlmonstr: Yes. This will be a part of https://github.com/torquem-ch/silkworm/pull/1008

silkworm/sync/sentry_client.cpp
battlmonstr: :+1: Let's try to improve this after https://github.com/torquem-ch/silkworm/pull/1008 . The signals/slots and sync methods could be removed if the calling code migrates to the async API, which I'm adding in #1008 for all required methods.

silkworm/sync/sentry_client.cpp
battlmonstr: I will add some docs to it in a follow-up PR.

PR #1006, opened 2023-04-21
sentry: update log messages

battlmonstr:
* add "sentry" category/module
* ensure prefix with class/method name
* prefix runtime_error messages with class/method name
* adjust log levels


canepat:
Requires #983

canepat (approved):

PR #1008, opened 2023-04-21
sentry: resend handshake/status after reconnecting

battlmonstr:


silkworm/infra/concurrency/awaitable_condition_variable.cpp
```diff
@@ -0,0 +1,92 @@
...
+        version_++;
+        for (auto& waiter : waiters_) {
+            waiter->notify();
+        }
+    }
+
+  private:
+    std::mutex mutex_;
+    std::set<std::unique_ptr<EventNotifier>> waiters_;
+    size_t version_{0};
```
canepat, line 74: I suggest to rename `version_` as `permit_`. Not blocking though, we can do it in another PR

silkworm/sentry/session_sentry_client.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <boost/asio/awaitable.hpp>
+
+#include <silkworm/sentry/api/api_common/sentry_client.hpp>
+#include <silkworm/sentry/eth/status_data.hpp>
+
+namespace silkworm::sentry {
+
+class SessionSentryClientImpl;
+
+class SessionSentryClient : public api::api_common::SentryClient {
```
canepat, line 33: I propose renaming this as `SentrySession`. Not blocking though, we can do it in another PR

cmd/backend_kv_server.cpp
```diff
@@ -175,10 +177,12 @@ int main(int argc, char* argv[]) {
...
         };
 
-        // remote sentry client
-        auto sentry_client = std::make_shared<silkworm::sentry::rpc::client::SentryClient>(
+        auto remote_sentry_client = std::make_shared<silkworm::sentry::rpc::client::SentryClient>(
             node_settings.external_sentry_addr,
             *context_pool.next_context().client_grpc_context());
+        auto sentry_client = std::make_shared<silkworm::sentry::SessionSentryClient>(
+            std::move(remote_sentry_client),
+            db::EthStatusDataProvider(db::ROAccess(database_env), node_settings.chain_config.value()).to_factory_function());
```
canepat, line 185: I'm not a huge fan of these names: `EthStatusDataProvider` and `to_factory_function` used in this way aren't as intuitive as I wish at a first read.

However, any alternative that comes to my mind (i.e. `EthStatusSource` and `data_factory`) is not very good as well, so we need to think it a bit more.

canepat (approved):

PR #1028, opened 2023-04-27
rpcdaemon: enable EVM analysis cache

lupin012:


silkworm/silkrpc/core/evm_executor.cpp
```diff
@@ -215,7 +215,11 @@ boost::asio::awaitable<ExecutionResult> EVMExecutor::call(
         [this, this_executor, &block, &txn, &tracers, &refund, &gas_bailout](auto&& self) {
             SILKRPC_TRACE << "EVMExecutor::call post block: " << block.header.number << " txn: " << &txn << "\n";
             boost::asio::post(workers_, [this, this_executor, &block, &txn, &tracers, &refund, &gas_bailout, self = std::move(self)]() mutable {
+                static constexpr size_t kCacheSize{5'000};
+                BaselineAnalysisCache analysis_cache{kCacheSize};
+
                 EVM evm{block, state_, config_};
+                evm.baseline_analysis_cache = &analysis_cache;
```
yperbasis, line 222: Subject to measurements, but perhaps sharing a single analysis cache between coroutines can give better performance. You'll need to make `BaselineAnalysisCache` thread-safe for that though.

Also, state pool should also probably help with the performance. And `ObjectPool` already has two flavours: thread-safe and not, depending on how you use it.

lupin012: Thanks.
I will update the Sw

silkworm/core/common/lru_cache.hpp
```diff
@@ -51,9 +52,57 @@ class lru_cache {
     typedef typename std::pair<key_t, value_t> key_value_pair_t;
     typedef typename std::list<key_value_pair_t>::iterator list_iterator_t;
 
-    explicit lru_cache(size_t max_size) : _max_size(max_size) {}
-
+    explicit lru_cache(size_t max_size, bool thread_safe = false) : _max_size(max_size), _thread_safe(thread_safe) {}
     void put(const key_t& key, const value_t& value) {
+        if (_thread_safe) {
+            const std::lock_guard<std::mutex> lock(_access);
```
yperbasis, line 58: You probably need to skip the mutex stuff when compiling for wasm, similar to how it's done in `core/common/object_pool.hpp`, because wasm doesn't support multi-threading (or perhaps it's experimental there).

PR #1032, opened 2023-04-28
rpcdaemon: use core lru_cache

lupin012:


silkworm/silkrpc/common/block_cache.hpp
```diff
@@ -19,11 +19,12 @@
 #include <cstddef>
 #include <mutex>
 
-#include <boost/compute/detail/lru_cache.hpp>
+#include <boost/optional.hpp>
```
canepat, line 22: You probably need to remove this include and add standard <optional> instead.

canepat (changes requested):

silkworm/silkrpc/common/block_cache.hpp
lupin012: ok

canepat (approved):

PR #1039, opened 2023-04-30
infra: sync_wait

mriccobene:
**Doing an asynchronous wait of a coroutine is easy**:

`auto result = co_await task();`

**Doing a synchronous wait of a coroutine is more verbose**:

`auto result = asio::co_spawn(io_context, task(), asio::use_future).get();`

also, this exposes implementation details.

**Using the simple sync_wait() function of this PR, the call becomes**:

`auto result = sync_wait(io_context, task());`

so at the call site, the intent of the code becomes clearer.

A simple getter enables us to simplify the code further when the io_context is not available in the scope and we need to get it from get_executor() method:

`auto result = sync_wait(in(this), task());`

### DIFF
--- a/silkworm/infra/concurrency/sync_wait.hpp
+++ b/silkworm/infra/concurrency/sync_wait.hpp
@@ -1,0 +1,76 @@
+/*
+   Copyright 2023 The Silkworm Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


canepat:
I like it! I would add to this also one `sync_wait` overload for `concurrency::TaskGroup` as done in:

https://github.com/torquem-ch/silkworm/blob/master/silkworm/sync/sentry_client.cpp#L156

You can do it in next PR

canepat (approved):

PR #1047, opened 2023-05-03
core: extract UnsignedTransaction

yperbasis:
Split `Transaction` into the `UnsignedTransaction` base struct and the signature.


silkworm/node/stagedsync/stages/stage_senders.cpp
```diff
@@ -480,7 +480,8 @@ Stage::Result Senders::add_to_batch(BlockNum block_num, std::vector<Transaction>
         }
 
         Bytes rlp{};
-        rlp::encode(rlp, transaction, /*for_signing=*/true, /*wrap_eip2718_into_string=*/false);
+        const UnsignedTransaction& unsigned_txn{transaction};
+        rlp::encode(rlp, unsigned_txn);  // encode for signing
```
canepat, line 484: I would consider to avoid overloading `rlp::encode` for `UnsignedTransaction` because this requires a conversion at call site, as shown here. If we rename this overload as `rlp::encode_for_signing`, IMO the code at call site becomes less error-prone (cannot call the wrong overload by missing the conversion) and more readable (no need for comment explaining this is for signing).

If we really need a `rlp::encode` overload for `UnsignedTransaction`, we can simply call `rlp::encode_for_signing` in its implementation.

canepat (changes requested):

silkworm/node/stagedsync/stages/stage_senders.cpp
yperbasis: Fair enough, I've made the change.

PR #1062, opened 2023-05-08
CI: update compilers

battlmonstr:
* update cpp-build-env to v20
* update clang 13 job to clang 16
* update gcc 11 conan job to gcc 13
* update magic_enum to v0.8.2 to compile on clang 16
* update libtorrent to RC_2_0 @ 272828e + custom fix https://github.com/arvidn/libtorrent/pull/7410 to compile on gcc 13
* fix cbor to compile on gcc 13 - https://github.com/torquem-ch/cbor-cpp/pull/2
* vendor wasmer.io install script to avoid CI error 400

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,13 +136,13 @@ jobs:
       - build
       - test
 
-  linux-gcc-13-conan:
+  linux-gcc-11-conan:
     environment:
       TOOLCHAIN: cxx20
       BUILD_TYPE: Debug
       BUILD_CMAKE_ARGS: -DCONAN_PACKAGE_MANAGER=ON -DCONAN_PROFILE=linux_gcc_11_debug
     docker:
-      - image: ethereum/cpp-build-env:20-gcc-13
+      - image: ethereum/cpp-build-env:20-gcc-11
     resource_class: xlarge
     steps:
       - run:
@@ -334,7 +334,7 @@ workflows:
   silkworm:
     jobs:
       - linux-gcc-12-release
-      - linux-gcc-13-conan
+      - linux-gcc-11-conan
       - linux-gcc-11-thread-sanitizer
       - linux-clang-13-coverage
       - linux-clang-15-address-sanitizer
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,7 @@ jobs:
       BUILD_TYPE: Debug
       BUILD_CMAKE_ARGS: -DSILKWORM_SANITIZE=thread
     docker:
-      - image: ethereum/cpp-build-env:18-gcc-11
+      - image: ethereum/cpp-build-env:20-gcc-11
     resource_class: xlarge
     steps:
       - checkout_with_submodules
@@ -140,7 +140,7 @@ jobs:
       BUILD_TYPE: Debug
       BUILD_CMAKE_ARGS: -DCONAN_PACKAGE_MANAGER=ON -DCONAN_PROFILE=linux_gcc_11_debug
     docker:
-      - image: ethereum/cpp-build-env:18-gcc-11
+      - image: ethereum/cpp-build-env:20-gcc-11
     resource_class: xlarge
     steps:
       - run:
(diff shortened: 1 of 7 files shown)


battlmonstr:
magic_enum fails to build on clang 16
abseil fails to build on gcc 13

yperbasis:
> magic_enum fails to build on clang 16 abseil fails to build on gcc 13

[Latest release](https://github.com/Neargye/magic_enum/releases/tag/v0.8.2) of magic_enum has a fix for  clang 16.

battlmonstr:
Hunter uses abseil version 20211102.0
which has a bug when compiling with GCC 13:
https://github.com/abseil/abseil-cpp/issues/1206
And the build fails:
https://app.circleci.com/pipelines/github/torquem-ch/silkworm/7079/workflows/79c77de4-5872-4782-b68f-3c2031bb7d2d/jobs/31127

This bug is fixed in the abseil version 20220623.0 used by Conan.

.circleci/config.yml
```diff
@@ -327,9 +334,9 @@ workflows:
   silkworm:
     jobs:
       - linux-gcc-12-release
-      - linux-gcc-11-conan
+      - linux-gcc-13-conan
       - linux-gcc-11-thread-sanitizer
-      - linux-clang-13-coverage
```
yperbasis, line 332: I'd rather not remove clang 13, since otherwise we might break compilation with it in the future. To my mind we can use this PR as an opportunity to improve our CI a little bit. I suggest to have jobs like clang-min and clang-latest, similar to evmone. Our README says that building silkworm requires "GCC >= 11.2.0 or Clang >= 12.0.0". We should check those versions in CI or update README if we in fact require higher versions.

battlmonstr: The idea is to shift supported compiler versions from clang 13-15 to 14-16, and gcc 11-12 to 12-13.
So in addition to supporting the new compilers (clang 16 and gcc 13) this PR drops support of old clang 13 and gcc 11.

The reason I'm keeping linux-gcc-11-thread-sanitizer for now is Hunter and its old abseil. This job can be bumped to gcc 13 when we move it to conan, and the first step in that direction is this PR: https://github.com/torquem-ch/silkworm/pull/1078

As for clang, "clang-min" becomes 14, and I agree that ideally we should have a job that builds with clang 14. The reason I didn't add it is that all our core developers are using Xcode toolchain currently based on clang 14. So the benefit of having it in CI is fairly low, while the benefit of not having an extra job seems tangible.

I understand that your idea is to have a clang 14 job instead of clang 15, because if we build fine with v14 and v16 the chance that we break on v15 is low. I'm fine with that, although it would mean that in practice we would never build on clang 15, so we don't really know if it works (until Xcode upgrades to it), and at the same time we would build on clang 14 even more than now, which feels redundant.

Thanks for reminding about the readme, I've updated it accordingly now.

battlmonstr: Also note that having a compiler version in the job name separates the build caches at the moment. We can definitely rename jobs to something like clang-min and clang-latest, but need to remember to revise the cache keys.

yperbasis: But why drop support of clang 12/13 and gcc 11 unless we want to use some C++ features not supported by the old compilers? The wider range of compilers we support, the better. For example, gcc11 is the default on the latest Ubuntu LTS.

README.md
```diff
@@ -59,7 +59,7 @@ git submodule update --init --recursive
 ## Building on Linux & macOS
 
 Building Silkworm requires:
-* C++20 compiler: [GCC](https://www.gnu.org/software/gcc/) >= 11.2.0 or [Clang](https://clang.llvm.org/) >= 12.0.0
+* C++20 compiler: [GCC](https://www.gnu.org/software/gcc/) >= 12 or [Clang](https://clang.llvm.org/) >= 14
```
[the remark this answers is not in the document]
battlmonstr: I've replied in the other comment. We have discussed it on a team meeting a while back, but I don't remember if you were present. Let's have a recap of this policy on the next meeting, and decide if we think there's still value in supporting these.

.circleci/config.yml
battlmonstr: Good point. Do we aim to support building on Ubuntu LTS with a default compiler?

battlmonstr: @yperbasis I've reverted this change until further discussion in order to unblock this PR. I'm not against your proposal, but I'd like to have everyone on the same page before proceeding.

canepat: OK we can merge it as it is now and discuss our compiler compatibility strategy separately

canepat (approved):

PR #1067, opened 2023-05-10
cmd: add state transition tool

JacekGlen:


cmd/state-transition/expected_state.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <iostream>
+
+#include <nlohmann/json.hpp>
+
+#include "silkworm/core/chain/config.hpp"
+#include "silkworm/core/common/test_util.hpp"
+
+namespace silkworm::cmd::state_transition {
+
+silkworm::ChainConfig ExpectedState::get_config() const {
```
canepat, line 28: Fully-qualified type name `silkworm::ChainConfig` can be simplified to just `ChainConfig` when we're already within `silkworm` namespace (this applies also to other types and namespaces in some other places)

cmd/state-transition/expected_state.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+class ExpectedSubState {
+  public:
+    unsigned index{};
+    evmc::bytes32 stateHash;
+    evmc::bytes32 logsHash;
+    unsigned long dataIndex{};
+    unsigned long gasIndex{};
+    unsigned long valueIndex{};
+    bool exceptionExpected;
```
canepat, line 35: Maybe you can give `exceptionExpected` a default `false` value

cmd/state-transition/runner.cpp
```diff
@@ -0,0 +1,72 @@
...
+    bool terminate_flag = false;
+    app.add_flag("-t,--terminateOnError", terminate_flag, "Terminate execution on failure");
+
+    bool diagnostics_flag = true;
+    app.add_flag("-d,--diagnostics", diagnostics_flag, "Enable extended diagnostics output");
+
+    CLI11_PARSE(app, argc, argv)
+
+    //    path = "/home/jacek/dev/silkworm/cmd/state-transition/state_transition_sample3.json";
+    //    path = "/home/jacek/dev/ethereum-tests/GeneralStateTests/stTransactionTest/ValueOverflow.json";
```
canepat, line 43: These lines seem a leftover after some tests

cmd/state-transition/state_transition.cpp
```diff
@@ -0,0 +1,318 @@
...
+    block.header.timestamp = std::stoull(get_env("currentTimestamp"), nullptr, /*base=*/16);
+    block.header.parent_hash = to_bytes32(silkworm::from_hex(get_env("previousHash")).value_or(silkworm::Bytes{}));
+
+    if (contains_env("currentRandom")) {
+        block.header.prev_randao = to_bytes32(silkworm::from_hex(get_env("currentRandom")).value_or(silkworm::Bytes{}));
+    }
+
+    const evmc_revision rev{chain_config.revision(block.header.number, block.header.timestamp)};
+
+    //    if (rev <= EVMC_LONDON)
```
canepat, line 113: It's better to remove this comment if it is a leftover or replace it with a `// TODO` if something needs to be done here

cmd/state-transition/state_transition.cpp
```diff
@@ -0,0 +1,318 @@
...
+    /// Example
+    // private key: 0x45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8
+    // public key : 043a514176466fa815ed481ffad09110a2d344f6c9b78c1d14afc351c3a51be33d8072e77939dc03ba44790779b7a1025baf3003f6732430e20cd9b76d953391b3
+    // address    : 0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B
+
+    auto private_key_bytes = from_hex(private_key).value();
+
+    silkworm::sentry::common::EccKeyPair pair = silkworm::sentry::common::EccKeyPair(private_key_bytes);
+
+    uint8_t out[20];
```
canepat, line 180: You can replace 20 with `kAddressLength` constant defined in `silkworm/core/common/base.hpp`

canepat (approved):
I've just found some minor issues you can fix in next PR

PR #1070, opened 2023-05-10
fix: avoid startup crash in backend_kv_server

### LINKED ISSUE #1054: backend_kv_server cmd startup crash
On master commit https://github.com/torquem-ch/silkworm/commit/3c30b26a87de9d4b722a3f85c5dc1ccfe68f68af

If I run:

    ./build/cmd/backend_kv_server --sentry.remote.addr=127.0.0.1:9091

It crashes immediately with EXC_BAD_ACCESS in `boost::asio::experimental::detail::channel_receive`.

The crash stack is useless, and comes from `silkworm::rpc::ServerContext::execute_loop_multi_threaded  - boost::asio::io_context::run`.

I've had the same crash in sentry cmd before I started using `TaskGroup` instead of `asio::detached` to properly cleanup awaitables on exceptions/cancellation.

I've checked that it is not related to sentry, because it still happens even if sentry client is not created:

        EthereumBackEnd backend{
            node_settings,
            &database_env,
            {}, // - a null shared_ptr for the sentry_client
        };

canepat:
Bug fixes:
- avoid multiple calls to `rpc::BackEndKvServer::build_and_start`
- refactor state changes simulator as async task

Bonus:
- use RAII for cleanup
- replace `rpc::ServerContextPool` w/ `rpc::ClientContextPool` (no more dummy `::grpc::ServerCompletionQueue`)

Fixes #1054

### DIFF
--- a/cmd/backend_kv_server.cpp
+++ b/cmd/backend_kv_server.cpp
@@ -32,8 +32,8 @@
 #include <silkworm/infra/common/log.hpp>
 #include <silkworm/infra/concurrency/awaitable_wait_for_one.hpp>
 #include <silkworm/infra/concurrency/context_pool_settings.hpp>
-#include <silkworm/infra/grpc/common/util.hpp>
 #include <silkworm/infra/grpc/client/client_context_pool.hpp>
+#include <silkworm/infra/grpc/common/util.hpp>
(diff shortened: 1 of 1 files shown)


battlmonstr (approved):
👍

PR #1112, opened 2023-05-16
sentry: refactor Timeout class to a function

battlmonstr:


silkworm/sentry/common/timeout.hpp
```diff
@@ -25,28 +25,15 @@
...
-    }
+    static boost::asio::awaitable<void> after(std::chrono::milliseconds duration);
 
     class ExpiredError : public std::runtime_error {
       public:
         ExpiredError() : std::runtime_error("Timeout has expired") {}
     };
-
-  private:
-    std::chrono::milliseconds duration_;
```
canepat, line 49: In order to simplify this even further, why not having just one plain function:
```
boost::asio::awaitable<void> timeout(std::chrono::milliseconds duration);
```
and maybe renaming `ExpiredError` as `TimeoutExpiredError`.

canepat (approved):

PR #1129, opened 2023-05-18
core: correct type of len in silkworm_rmd160

yperbasis:
The type of `len` in `rmd160_finish` is `uint32_t`, so it should be the same in `silkworm_rmd160`.

### DIFF
--- a/silkworm/core/crypto/rmd160.c
+++ b/silkworm/core/crypto/rmd160.c
@@ -368,8 +368,8 @@ void silkworm_rmd160(uint8_t out[20], const uint8_t* ptr, uint32_t len) {
     rmd160_init(buf);
 
     uint32_t current[16];
-    for (unsigned remaining = len; remaining >= 64; remaining -= 64) {
-        for (unsigned i = 0; i < 16; ++i) {
+    for (size_t remaining = len; remaining >= 64; remaining -= 64) {
+        for (size_t i = 0; i < 16; ++i) {
             current[i] = load32(ptr);
             ptr += 4;
         }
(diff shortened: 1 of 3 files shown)


silkworm/core/crypto/rmd160.c
```diff
@@ -362,13 +362,13 @@ static inline uint32_t load32(const void* src) {
...
 
-void silkworm_rmd160(uint8_t out[20], const uint8_t* ptr, size_t len) {
+void silkworm_rmd160(uint8_t out[20], const uint8_t* ptr, uint32_t len) {
     uint32_t buf[160 / 32];
 
     rmd160_init(buf);
 
     uint32_t current[16];
-    for (size_t remaining = len; remaining >= 64; remaining -= 64) {
+    for (unsigned remaining = len; remaining >= 64; remaining -= 64) {
```
chfast, line 371: This does not looks good. `size_t` is always better for loop counter and `remaining` looks unused for anything else.

chfast (commented):
I'd actually consider changing `len` to `size_t`.

yperbasis:
> I'd actually consider changing `len` to `size_t`.

The problem is that `rmd160_finish` expects `len` to be `uint32_t`. `rmd160_finish` is a low-level function written by somebody else; and I'd rather not spend time figuring out how to change `len` there to `uint64_t`, especially given that there's not real need for it.

silkworm/core/crypto/rmd160.c
yperbasis: Why is `size_t` better for loop counter?

chfast: It's natural register size. The `unsigned` in the other hand is 32-bit on 64-bit architecture and compiler has to guarantee wrapping behavior which is not always compatible with 64-bit architecture addressing modes.

In this case it does not make big difference, but I like to use it as a rule of thumb.

I cheated a bit by making `size_t len`, but the difference is mostly the ability to use `lea` instruction in this case.
https://godbolt.org/z/xM15GjPGd

yperbasis: OK, interesting. I've change it to `size_t`.

canepat (approved):

chfast (approved):

PR #1130, opened 2023-05-18
CI: enable conan by default

battlmonstr:


yperbasis:
Please merge the latest master so that PR #1082 is not shown in the diff.

.github/workflows/macOS.yml
```diff
@@ -36,8 +36,8 @@ jobs:
         submodules: recursive
         fetch-depth: "0"
 
-    - name: Install GMP & mimalloc
-      run: brew install gmp mimalloc
+    - name: Install Prerequisites
+      run: brew install gmp conan@1
```
canepat, line 40: This way we'd use latest Conan 1 version in `brew` formula, right? This is not the same as stated in README, where we suggest using

`pip3 install --user conan==1.58.0 chardet`

that is a different installation method with a more specific version.

battlmonstr: Yeah, initially I thought that it is possible to use an exact version with brew, and that it would add it to PATH, but it doesn't work. I'm going to try the pip version.

PR #1145, opened 2023-05-21
CI: migrate wasm job to conan

battlmonstr:


.circleci/config.yml
```diff
@@ -16,11 +16,21 @@ version: 2.1
...
+        default: true
     steps:
       - checkout
       - run:
           name: "Update submodules"
           command: |
+            if [[ <<parameters.ethereum_tests>> == false ]]
+            then
+              git config submodule.tests.update none
+            fi
```
yperbasis, line 32: Perhaps you can also unconditionally disable checkout of the `LegacyTests` submodule of Ethereum tests. We don't use those.

PR #1147, opened 2023-05-21
deps: use CRoaring from conan

battlmonstr:
The git checkout size is 340 Mb, in conan it is just 1.4 Mb.
This should make submodules update 10 sec faster on CI.


cmd/test/CMakeLists.txt
```diff
@@ -45,7 +46,7 @@ if(NOT SILKWORM_CORE_ONLY)
   if(SILKWORM_SANITIZE)
     target_compile_definitions(node_test PRIVATE GRPC_ASAN_SUPPRESSED GRPC_TSAN_SUPPRESSED)
   endif()
-  target_link_libraries(node_test silkworm_node Catch2::Catch2)
+  target_link_libraries(node_test silkworm_node Catch2::Catch2 roaring::roaring)
```
yperbasis, line 49: Why did you move roaring from SILKWORM_NODE_PUBLIC_LIBS to SILKWORM_NODE_PRIVATE_LIBS? It's used in a header, so it should be a PUBLIC dependency.

yperbasis (changes requested):

cmd/test/CMakeLists.txt
battlmonstr: I assumed that this is a private implementation detail, and that the header in which it is used is private to the "node" library. Do you see it being used outside of "node"? (other than in tests of node, because in the "node_test" case it needs to access private details for unit testing)

yperbasis: Well, it's not private because it's used in public functions (e.g. `flush_bitmaps_to_etl`). Maybe we could make the dependency private by using PImpl or something similar, but personally I don't see a reason to bother.

battlmonstr: 👌 moved back to be public

canepat (approved):

yperbasis (approved):

PR #1150, opened 2023-05-21
deps: use GMP from conan

battlmonstr:
Makes GCC 12 release build 2 minutes faster.


cmd/test/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@
 ]]
 
 find_package(Catch2 REQUIRED)
+find_package(roaring REQUIRED)
```
yperbasis, line 18: The previous roaring change was reworked in PR #1147 and shouldn't resurface in this PR.

yperbasis (approved):

PR #1174, opened 2023-05-25
core: prohibit leftover input after RLP decoding by default

yperbasis:
By default, require that an input to `rlp::decode` represents exactly the object being decoding and doesn't contains any extra characters.


battlmonstr:
@yperbasis what's the reasoning behind this?

yperbasis:
> @yperbasis what's the reasoning behind this?

My thinking is that by default the input to `rlp::encode` should not contain any extra trailing characters. In other words, the default assumption is that the input corresponds exactly to a valid RLP encoding. One can switch to the old behaviour with accepted trailing characters by setting `allow_leftover` to true.

mriccobene:
In this PR I like the removal of the templatized version of decode() and the more widespread use of the variadic decode().

About the `allow_leftover` I have a mixed feeling. 
I tend to prefer the earlier version where you usually don't take care of the trailing characters because you usually have nested decode calls, except where it is requested. And to do so you call an explicit method (to be added) only in the places where you need it. 
On the pros side: now each decode call remember you that you have to decide if accept or not the extra characters so this important extra behavior is explicitly stated in the function signature. In this case I would like to use global defines:

```
bool allow_leftover = true;
bool disallow_leftover = false;
```

so to call 

`decode(payload_view, to, allow_leftover)`

or 

`decode(payload_view, to, disallow_leftover)`

because using comments is ugly

`decode(payload_view, to, /*allow_leftover=*/true)`

yperbasis:
> because using comments is ugly
> 
> `decode(payload_view, to, /*allow_leftover=*/true)`

I could add a small enum so that we have
```
decode(payload_view, to, Leftover::kAllow)
```
instead. What do you think?

mriccobene (approved):

Issue #1171, opened 2023-05-24
Build fails: GMP missing for libff


chfast:
```
-- Found: /home/builder/.conan/data/protobuf/3.21.4/_/_/package/2dbf65f76c0469903ce48756c39d50cd4e721678/lib/libprotoc.a
-- Configuring done
CMake Error in third_party/libff/libff/CMakeLists.txt:
  Imported target "GMP::gmp" includes non-existent path
```

I believe libff is not using GMP from conan and when you don't have it installed in the system it will fail.

battlmonstr:
should be fixed by #1184

PR #1184, opened 2023-05-26
CI: use GMP from Conan on Windows too

battlmonstr:


yperbasis:
If GMP is available with Conan for windows, we can unset `MPIR_INSTEAD_OF_GMP` and use GMP on all platforms. Some time ago GMP was not compatible with Windows (or at least that was my impression) and that was why I added `MPIR_INSTEAD_OF_GMP` to libff. But perhaps now GMP from Conan works fine for Windows.

battlmonstr:
> If GMP is available with Conan for windows, we can unset `MPIR_INSTEAD_OF_GMP` and use GMP on all platforms. Some time ago GMP was not compatible with Windows (or at least that was my impression) and that was why I added `MPIR_INSTEAD_OF_GMP` to libff. But perhaps now GMP from Conan works fine for Windows.

@yperbasis 👌 will try

battlmonstr:
core_test build gets the following errors now:

```
error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in unit_test.obj
error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in unit_test.obj
```

in
absl_hash.lib(hash.obj)
absl_city.lib(city.obj)
absl_low_level_hash.lib(low_level_hash.obj)
absl_raw_hash_set.lib(raw_hash_set.obj)

and finally:
```
C:\build\cmd\test\Debug\core_test.exe : fatal error LNK1319: 8 mismatches detected [C:\build\cmd\test\core_test.vcxproj]
```

It looks like absl and unit_test.cpp from core_test suddenly use different version of the CRT, but why? 🤔 

It doesn't seem related to GMP though, could be that the Debug build on Windows CI is broken.

canepat:
@mriccobene can you please give this PR a try on your Windows machine? Unfortunately, our Windows CI job does not build the `silkworm` target because it still suffers an issue (the dreaded `compiler out of heap space in step 2`).

yperbasis (approved):

battlmonstr:
@canepat I'd like to merge it, because it also fixes #1171 . GMP is used in precompile.cpp from core, and building core is verified by the CI. It is still nice to test the full build on Windows, and if there are any further issues with the silkworm target, I can address it in a new PR.

PR #1192, opened 2023-05-28
rpcdaemon: add net in list of all API namespaces

lupin012:


silkworm/silkrpc/common/constants.hpp
```diff
@@ -34,10 +34,11 @@ constexpr const char* kTraceApiNamespace{"trace"};
 constexpr const char* kWeb3ApiNamespace{"web3"};
 constexpr const char* kOtterscanApiNamespace{"ots"};
 
-constexpr std::array<const char*, 9> kAllEth1Namespaces{
+constexpr std::array<const char*, 10> kAllEth1Namespaces{
```
yperbasis, line 37: You can drop `<const char*, 10>` from the declaration and rely on the automatic [deduction guide](https://en.cppreference.com/w/cpp/container/array/deduction_guides) instead.

yperbasis (approved):

PR #1193, opened 2023-05-28
build: no need for -fpic

yperbasis:
Position independent code is required for shared libraries, which we don't use. Moreover, cmake sets pic to true automatically for shared libs.


chfast (approved):
There is a bit more to this. You can also produce Position Independent Executable (PIE) although I'm not sure if CMake supports it directly. The PIE enhances address space layout randomization (ASLR).

To the point: some compilers (e.g. GCC on Debian) are configured to use PIC/PIE by default. Then you need to disable it manually (e.g. `-no-pie`, `-fno-PIC`).

But the change itself is good.

PR #1194, opened 2023-05-29
CI: install conan from setup.exe on Windows

battlmonstr:


battlmonstr:
The idea was to try installing Conan using the provided setup.exe, and compare if it is faster than the current method which sometimes takes 3-4 minutes.

I couldn't make it work. It seems that running conan-win-64.exe does nothing (crashes?). Debugging using various flags didn't help. Going further would require a Windows machine to see what's expected to get from this, and trying to understand what prevents it from happening on CI.

PR #1195, opened 2023-05-29
CI: use ninja-build

battlmonstr:


battlmonstr:
@yperbasis @canepat This doesn't seem to have any effect on the build time, unfortunately. The build step is still around 12-13 minutes. I'm not sure we need this change. What do you think?

Issue #1191, opened 2023-05-28
SHA256 is broken in GCC 13 Release builds


chfast:
I has been investigating why evmone's build of silkworm is failing ~300 blockchain tests.

- The bug is only present in GCC 13 Release build. Both Debug and GCC 12 Release builds are fine.
- The root cause is related to SHA256. In `core_test` 3 SHA256 tests are failing.
- This is unrelated to the selection of the architecture-specific implementation.
- The returned hash looks like corrupted/uninitialized memory. The hash is different every run.

PR #1196, opened 2023-05-29
core: fix uninitialized memory in SHA256

chfast:
Move `w[16]` buffer outside of the outer loop. Otherwise, `w` goes out of scope in every outer loop iteration and uninitialized memory is used. This bug has become visible by building with GCC 13 in Release.

I restored the correct location of `w[16]` as in the original implementation at
https://github.com/amosnier/sha-2/blob/5beabbcc0d872e15a16d51539922dbfabb806316/sha-256.c#L40-L53. The code was introduced already with the bug in
https://github.com/torquem-ch/silkworm/pull/113.

Fixes https://github.com/torquem-ch/silkworm/issues/1191.

### DIFF
--- a/silkworm/core/crypto/sha256.c
+++ b/silkworm/core/crypto/sha256.c
@@ -181,25 +181,23 @@ static inline ALWAYS_INLINE void sha_256_implementation(uint32_t h[8], const voi
 
         const uint8_t* p = chunk;
 
+        /*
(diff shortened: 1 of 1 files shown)


yperbasis (approved):
Thank you! I wonder why sanitize=undefined didn't catch it.

PR #1222, opened 2023-06-02
sync: refactor receive_peer_events to not rely on incremental active_peers updates

battlmonstr:
The count can go out of sync for various reasons.
Update it on each peer event instead.

Remove unused peer_infos_ map.

see: https://github.com/ledgerwatch/interfaces/issues/168


mriccobene (approved):
thanks

PR #1227, opened 2023-06-06
CI: clang-tidy

battlmonstr:


silkworm/node/huffman/decompressor_test.cpp
```diff
@@ -151,8 +152,8 @@ TEST_CASE("PatternTable::build_condensed", "[silkworm][snapshot][decompressor]")
     std::vector<Pattern> patterns2{{1, v1}, {2, v2}};
     std::map<std::string, std::span<Pattern>> test_spans{
         {"zero patterns", patterns0},
-        {"one pattern", patterns1},
-        {"two patterns", patterns2},
+        {"one pattern", std::span<Pattern>{&patterns1[0], patterns1.size()}},
+        {"two patterns", std::span<Pattern>{&patterns2[0], patterns2.size()}},
```
canepat, line 156: I suggest using `patterns.data()` instead of `&patterns[0]` for improved readability

silkworm/sync/internals/chain_elements.hpp
```diff
@@ -261,8 +261,8 @@ struct Segment
     using Slice = std::span<const HeaderList::Header_Ref>;  // a Segment slice
 
     [[nodiscard]] Slice slice(size_t start, size_t end) const {
-        return Slice(*this).subspan(start, end - start);
-    }  // with c++20 it can be implemented as: return Slice(begin() + start, begin() + end);
+        return Slice(&(*this)[start], &(*this)[end]);
```
canepat, line 264: I'd suggest using `return Slice(begin() + start, begin() + end);` again for readability

silkworm/node/huffman/decompressor_test.cpp
battlmonstr: 👌

silkworm/sync/internals/chain_elements.hpp
battlmonstr: For some reason the stdlib that the clang-tidy build uses doesn't contain any span constructors other than `slice(ptr, count)` and `slice(ptr1, ptr2)`, so passing an iterator like `begin() + start` fails compilation, but I can try using `data() + start`.

canepat (approved):

PR #1246, opened 2023-06-10
rpcdaemon: fix eth_callMany state on worker thread

lupin012:


silkworm/silkrpc/core/call_many.cpp
```diff
@@ -43,6 +46,124 @@ namespace silkworm::rpc::call {
 
 using boost::asio::awaitable;
 
+awaitable<CallManyResult> CallExecutor::executes_all_bundles(const silkworm::ChainConfig* config,
```
canepat, line 49: I suggest making `executes_all_bundles` an ordinary function (i.e. non-coroutine) with return type `CallManyResult ` and moving the `boost::asio::async_compose` + `boost::asio::post` trick into `execute`.

This way it is more straightforward which `CallExecutor` method is sync vs async.

canepat (changes requested):

silkworm/silkrpc/core/call_many.cpp
lupin012: ok

canepat (approved):

PR #1247, opened 2023-06-12
windows: raise tests file descriptors limit

battlmonstr:
backend_kv_server_test.cpp(898): FAILED:
explicitly with message:
  insufficient number of process file descriptors, increase to 1024 at least


yperbasis (changes requested):
It's better to do it generically (non-windows-specific) using `os::set_max_file_descriptors`, similar to how it done in `NodeImpl::setup_snapshots`.

PR #1251, opened 2023-06-12
node: fix ASAN + UBSAN issues in snapshot RecSplit index

canepat:
- refactor Golomb-Rice read_next to fix buffer overflow
- use byte buffer in Elias-Fano encoding as input data to fix misaligned address reference
- refactor Elias-Fano encoding to avoid duplicated code and add invariant checks
- add `print_stacktrace=1` UBSAN option in CI sanitizer job


mriccobene (approved):

yperbasis:
Should the test disabled in PR #1244 be reenabled after this fix?

PR #1263, opened 2023-06-15
node: more support for snapshots in db access layer

canepat:
node: read sequence of txs from tx snapshot
node: remove unused access layer APIs
rpcdaemon: fix read block for missing system txs

### DIFF
--- a/silkworm/node/db/access_layer.cpp
+++ b/silkworm/node/db/access_layer.cpp
@@ -22,6 +22,7 @@
 #include <silkworm/core/common/cast.hpp>
 #include <silkworm/core/common/endian.hpp>
 #include <silkworm/infra/common/decoding_exception.hpp>
+#include <silkworm/infra/common/ensure.hpp>
 #include <silkworm/node/db/bitmap.hpp>
 #include <silkworm/node/db/tables.hpp>
 
@@ -457,11 +458,6 @@ bool has_body(ROTxn& txn, BlockNum block_number, const evmc::bytes32& hash) {
     return db::has_body(txn, block_number, hash.bytes);
 }
 
-bool has_sibling(ROTxn&, BlockNum) {
-    return false;
-    // todo: implement!
-}
-
 void write_body(RWTxn& txn, const BlockBody& body, const evmc::bytes32& hash, BlockNum bn) {
     write_body(txn, body, hash.bytes, bn);
 }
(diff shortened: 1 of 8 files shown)


silkworm/node/db/access_layer.cpp
```diff
@@ -882,121 +874,261 @@ void write_last_finalized_block(RWTxn& txn, const evmc::bytes32& hash) {
...
 
-bool DataModel::read_canonical_block(BlockNum height, Block& block) {
-    return db::read_canonical_block(txn_, height, block);
+std::optional<BlockHeader> DataModel::read_canonical_header(BlockNum height) const {
+    // Canonical chain is likely on db: first lookup there
+    auto canonical_header{db::read_canonical_header(txn_, height)};
+    if (canonical_header) return canonical_header;
+
+    // Then search for it in the snapshots (if any)
+    return read_header_from_snapshot(height);
```
mriccobene, line 995: Same as above. We need to implement it with a first read_canonical_hash(), then, if present we will read from db or snapshot.

```    
std::optional<evmc::bytes32> h = read_canonical_hash(txn, b);
if (!h) return std::nullopt;  // not found
return read_header(txn, b, h->bytes);  // we should raise exception if it fails
```

silkworm/node/db/access_layer.cpp
```diff
@@ -882,121 +874,261 @@ void write_last_finalized_block(RWTxn& txn, const evmc::bytes32& hash) {
...
+    const bool found = db::read_canonical_block(txn_, height, block);
+    if (found) return found;
+
+    auto block_header{read_header_from_snapshot(height)};
+    if (!block_header) return false;
+
+    const auto hash{block_header->hash()};
+    block.header = std::move(*block_header);
+
+    return read_body(height, hash.bytes, /*read_senders=*/false, block);
```
mriccobene, line 1008: Same as above

silkworm/node/db/access_layer.cpp
```diff
@@ -882,121 +874,261 @@ void write_last_finalized_block(RWTxn& txn, const evmc::bytes32& hash) {
...
 }
 
-bool DataModel::has_sibling(BlockNum block_number) {
-    return db::has_sibling(txn_, block_number);
+bool DataModel::has_body(BlockNum height, const Hash& hash) {
+    return has_body(height, hash.bytes);
 }
 
 bool DataModel::read_block_by_number(BlockNum number, bool read_senders, Block& block) {
     return db::read_block_by_number(txn_, number, read_senders, block);
```
mriccobene, line 1023: We need to reimplement read_block_by_number here because it reads from canonical table (see comments above)

mriccobene (changes requested):

mriccobene (approved):

PR #1274, opened 2023-06-23
rpcdaemon: estimate gas oracle rework

lupin012:


silkworm/silkrpc/core/estimate_gas_oracle.hpp
```diff
@@ -40,6 +40,7 @@ namespace silkworm::rpc {
 
 const std::uint64_t kTxGas = 21'000;
 const std::uint64_t kGasCap = 25'000'000;
+const std::uint64_t pre_check_error = 1000;
```
canepat, line 43: The logic behind this `pre_ckeck_error` value and its usage goes back to the definition of `ExecutionResult`, which currently contains a `int64_t error_code` field used to carry both the `evmc_status_code` values and this special value 1000.
This is kinda weird and above all can lead to a value clash in the unlikely but possible case `evmone` will add a new value 1000 in its own `evmc_status_code`.
I suggest to change `error_code` as `std::optional<evmc_status_code>` or `std::variant< evmc_status_code, std::string>` and add a couple of methods to `ExecutionResult`:
- `bool success() const`: returns `true` if execution has been successful false if any error (pre-check or execution) has happened
- `std::string error_message() const`: returns the error message (pre-check error message or error message for `dvmc_status_code` by calling `EVMExecutor::get_error_message`)

Please do this in a separate PR.

canepat (changes requested):

canepat (approved):

PR #1341, opened 2023-07-17
rpcdaemon: Remote backend get block interface rework

lupin012:


silkworm/silkrpc/storage/remote_chain_storage.cpp
```diff
@@ -41,49 +41,34 @@ Task<BlockNum> RemoteChainStorage::highest_block_number() const {
...
-    silkworm::Block block;
-    const bool success = co_await backend_->get_block({number, hash}, /*.read_senders=*/false, block);
-    std::optional<BlockHeader> header;
-    if (success) {
-        header = std::move(block.header);
-    }
-    co_return header;
+Task<std::optional<BlockHeader>> RemoteChainStorage::read_header(BlockNum number, HashAsArray& hash) const {
+    Hash hash_type{hash};
+    co_return co_await core::rawdb::read_header(reader_, hash_type, number);
```
canepat, line 63: You need to switch back to the previous implementation to make it work also with remote snapshots

silkworm/silkrpc/storage/remote_chain_storage.cpp
```diff
@@ -41,49 +41,34 @@ Task<BlockNum> RemoteChainStorage::highest_block_number() const {
...
 }
 
 Task<std::optional<BlockHeader>> RemoteChainStorage::read_header(BlockNum number, const Hash& hash) const {
-    co_return co_await read_header(number, hash.bytes);
+    co_return co_await core::rawdb::read_header(reader_, hash, number);
```
canepat, line 67: Same thing as above

canepat (changes requested):

PR #1367, opened 2023-07-24
api: new module containing C API for Erigon

canepat:


silkworm/api/silkworm_api.h
```diff
@@ -0,0 +1,88 @@
...
+ * Not written to if no blocks were executed, otherwise *last_executed_block ≤ max_block.
+ * @param[out] mdbx_error_code If an MDBX error occurs (this function returns kSilkwormMdbxError)
+ * and mdbx_error_code isn't NULL, it's populated with the relevant MDBX error code.
+ *
+ * @return A non-zero error value on failure and kSilkwormSuccess(=0) on success.
+ * kSilkwormBlockNotFound is probably OK: it simply means that the execution reached the end of the chain
+ * (blocks up to and incl. last_executed_block were still executed).
+ */
+SILKWORM_EXPORT
+enum SilkwormStatusCode silkworm_execute_blocks(
```
yperbasis, line 79: It's probably good for start experimenting, but I suspect that the current Silkworm implementation is not exactly equivalent to the current Erigon implementation of the execution stage. Off the top of my head I can think of 2 discrepancies:
1.  Lack of support of Gnosis Chain & Polygon in Silkworm
2. No txpool notifications

canepat: Yep, thanks for the heads up. This first iteration is a feasibility assessment, so I think we can postpone both topics for further development if the solution proves to be consistent and working at least for Ethereum w/o block production

PR #1370, opened 2023-07-25
rpcdaemon: implement debug_getRaw* endpoints

JacekGlen:
This implements:
- debug_getRawBlock
- debug_getRawHeader
- debug_getRawTransaction

The two endpoints were omitted (for now):
- debug_getRawReceipts (cbor_decode function required)
- debug_getBadBlocks (data not available)

Some functions are not optimized, as they require decoding data and then encoding it back again. 
Also, snapshot access is not implemented. This will be done in the later phase.


silkworm/node/db/access_layer.cpp
```diff
@@ -1308,6 +1308,29 @@ bool DataModel::read_rlp_transactions(BlockNum height, const evmc::bytes32& hash
...
+    return read_tx_lookup_from_snapshot(tx_hash);
+}
+
+std::optional<BlockNum> DataModel::read_tx_lookup_from_db(const evmc::bytes32& tx_hash) const {
+    auto cursor = txn_.ro_cursor(table::kTxLookup);
+    auto data{cursor->find(to_slice(tx_hash.bytes), /*throw_notfound = */ false)};
+    if (!data) {
+        return std::nullopt;
+    }
+    auto block_num = endian::load_big_u64(static_cast<const unsigned char*>(data.value.data()));
```
canepat, line 1326: Here you can use our `from_slice` utility function to avoid spreading the cast all over the places like this:
```
return endian::load_big_u64(from_slice(data.value).data());
```

silkworm/silkrpc/core/blocks.cpp
```diff
@@ -78,10 +79,13 @@ boost::asio::awaitable<std::pair<uint64_t, bool>> get_block_number(const std::st
...
         is_latest_block = true;
-    } else {
-        block_number = static_cast<uint64_t>(std::stol(block_id, nullptr, 0));
+    } else if (is_valid_hex(block_id)) {
+        block_number = static_cast<uint64_t>(std::stol(block_id, nullptr, 16));
         check_if_latest = latest_required;
+    } else {
+        throw std::invalid_argument("Invalid Block Id");
     }
+
```
canepat, line 88: Is this change required specifically by some Execution API unit test? If this is not the case, I would prefer the previous conversion because by using `0` as numeric base we allow for both hex and decimal block number in input string.

This version is stricter, which can be fine or maybe better but we need to check all the API implementations using directly or indirectly the `get_block_number` function. Currently, this requires you to run the integration tests, which in turn requires an Erigon core component running on Goerli testnet.

I hope we can move this kind of non-regression validations to our unit test suite for `silkrpc` you're building, but for the time being we need to ask @lupin012 to run the integration tests on this branch.

canepat (changes requested):

canepat:
I forgot to mention that we have [this doc](https://github.com/torquem-ch/silkworm/blob/1355-debug-getraw/docs/JSON-RPC-API.md) to keep track of the overall API status, you can add these 5 new APIs with their status there.

silkworm/node/db/access_layer.cpp
JacekGlen: done

silkworm/silkrpc/storage/local_chain_storage.cpp
```diff
@@ -114,6 +116,27 @@ Task<bool> LocalChainStorage::read_rlp_transactions(BlockNum number, const evmc:
     co_return data_model_.read_rlp_transactions(number, hash, rlp_txs);
 }
 
+Task<Bytes> LocalChainStorage::read_rlp_transaction(const evmc::bytes32& txn_hash) const {
+    auto block_id = data_model_.read_tx_lookup(txn_hash);
```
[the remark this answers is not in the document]
JacekGlen: done

silkworm/silkrpc/storage/chain_storage.hpp
```diff
@@ -87,6 +87,8 @@ class ChainStorage {
     //! Read the RLP encoded block transactions at specified height
     virtual Task<bool> read_rlp_transactions(BlockNum number, const evmc::bytes32& hash, std::vector<Bytes>& rlp_txs) const = 0;
 
+    [[nodiscard]] virtual Task<Bytes> read_rlp_transaction(const evmc::bytes32& txn_hash) const = 0;
```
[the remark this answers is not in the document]
JacekGlen: done

canepat (approved):

PR #1378, opened 2023-07-26
build, core: switch to std::bit_cast. Bump min Clang to 14

yperbasis:
Also bump the minimum version of Clang to 14 and [Xcode](https://developer.apple.com/xcode/cpp/) to 14.3.


canepat (approved):

chfast (commented):
LGTM

PR #1405, opened 2023-08-03
build: use intx as conan package

pwqbot:
intx/0.10.1 has been added to conan-center: https://github.com/conan-io/conan-center-index/pull/18858, we can use conan to manage it instead of git submodule

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -33,9 +33,6 @@
 [submodule "third_party/blst"]
 	path = third_party/blst
 	url = https://github.com/supranational/blst.git
-[submodule "third_party/intx"]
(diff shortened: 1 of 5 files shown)


battlmonstr:
The conan install command fails with this on CI:

```
ERROR: There are invalid packages (packages that cannot exist for this configuration):
intx/0.10.1: Invalid ID: intx/0.10.1 requires C++20, which your compiler does not support.
```

I'm getting the same error when running this command locally:

    conan install --install-folder=build/conan --build=missing --profile=cmake/profiles/macos_arm_clang_13_debug .

Please fix and reopen the PR.

PR #1406, opened 2023-08-03
rpcdeamon: enable eth_getTransactionByHash test

JacekGlen:
- implement snapshot get block number by txn hash function
- add mandatory yParity field to the txn json output
- block_from <= block_to logic changed, this is due to kFileNameBlockScaleFactor, as in `const BlockNum block_from{scaled_block_from * kFileNameBlockScaleFactor};` in which case if the real block numbers are from 1'500'012 to 1'500'013, both would be represented as 1'500 and hence fail the check


canepat:
@lupin012 can you please take a look at changes in `silkrpc` module?

silkworm/silkrpc/json/transaction.cpp
```diff
@@ -53,6 +53,7 @@ void to_json(nlohmann::json& json, const Transaction& transaction) {
         json["chainId"] = rpc::to_quantity(*transaction.chain_id);
         json["v"] = rpc::to_quantity(uint64_t(transaction.odd_y_parity));
         json["accessList"] = transaction.access_list;  // EIP2930
+        json["yParity"] = rpc::to_quantity(transaction.odd_y_parity);
```
lupin012, line 56: I propose to make this interface change later.
This change can have impact in many integration test see tests/integration/goerly/....
We need agree rpcademon/erigon to be used that support this new field and updates all the integration test impacted.

silkworm/silkrpc/commands/eth_api.cpp
```diff
@@ -604,6 +604,8 @@ awaitable<void> EthereumRpcApi::handle_eth_get_transaction_by_hash(const nlohman
         }
     } catch (const std::invalid_argument& iv) {
         reply = make_json_content(request["id"], {});
+    } catch (const boost::system::system_error& se) {
```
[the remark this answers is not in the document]
JacekGlen: It's a silly thing really. The [test ](https://github.com/ethereum/execution-apis/blob/main/tests/eth_getTransactionByHash/get-notfound-tx.io) specifies that not found transaction hash should return null, not error. Now, the logic in our code says:
- find tx in local db
- if not found, search TransactionPool
It's the second step that causes troubles in unit testing. TransactionPool is `final` hence you can't use a dummy subclass. Also there is no easy way to mock grpc.
In effect, the least invasive way to satisfy the unit test, is to catch that exception. Handling that exception also makes sense business-wise, in some way.

silkworm/silkrpc/core/cached_chain.cpp
```diff
@@ -125,6 +125,11 @@ awaitable<BlockWithHash> read_block_by_transaction_hash(BlockCache& cache, const
 
 awaitable<BlockWithHash> read_block_by_transaction_hash(BlockCache& cache, const ChainStorage& storage, const evmc::bytes32& transaction_hash) {
     auto block_number = co_await storage.read_block_number_by_transaction_hash(transaction_hash);
+
+    if (!block_number) {
+        throw std::runtime_error{"block not found"};
+    }
```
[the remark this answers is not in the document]
JacekGlen: done

silkworm/silkrpc/json/transaction.cpp
JacekGlen: This field was introduced in EIP-1559 and is mandatory, according to the [spec](https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L177). Geth has already implemented it, but Erigon seems to be lagging behind.
So I can revert this change and disable unit tests, or we can fix integration tests. It seems easy enough, just updating json files, I can create another PR. It's your call!

canepat: I really understand the point raised by @lupin012 about interface change, especially in terms of running integration tests against Erigon RpcDaemon for comparison purposes. Nevertheless, I appreciate the great value added by being compliant with the correct behaviour according to the spec, so I suggest the following:
- merging this PR as it is
- adding a new command-line option `--erigon-compatibility` to `silkrpcdaemon` to force behaviour equal to Erigon, so we can still run the integration tests for comparison

canepat (approved):

PR #1414, opened 2023-08-04
Rpcademon fix local db no snapshot

lupin012:


lupin012:
branch of branch. create new one from master

PR #1415, opened 2023-08-04
rpcdaemon: fix local db access without snapshots

lupin012:


silkworm/silkrpc/core/cached_chain.hpp
```diff
@@ -34,7 +34,7 @@ using boost::asio::awaitable;
 awaitable<std::shared_ptr<BlockWithHash>> read_block_by_number(BlockCache& cache, const ChainStorage& storage, BlockNum block_number);
 awaitable<std::shared_ptr<BlockWithHash>> read_block_by_hash(BlockCache& cache, const ChainStorage& storage, const evmc::bytes32& block_hash);
 awaitable<std::shared_ptr<BlockWithHash>> read_block_by_number_or_hash(BlockCache& cache, const ChainStorage& storage, const rawdb::DatabaseReader& reader, const BlockNumberOrHash& bnoh);
-awaitable<BlockWithHash> read_block_by_transaction_hash(BlockCache& cache, const ChainStorage& storage, const evmc::bytes32& transaction_hash);
+awaitable<std::shared_ptr<BlockWithHash>> read_block_by_transaction_hash(BlockCache& cache, const ChainStorage& storage, const evmc::bytes32& transaction_hash);
```
JacekGlen, line 37: Why not use `std::unique_ptr` in this scenario? Is there any case when we would need shared ownership of the block object?
Not a problem, just curious.

JacekGlen (approved):

silkworm/silkrpc/core/cached_chain.hpp
canepat: > Why not use `std::unique_ptr` in this scenario? Is there any case when we would need shared ownership of the block object?

Because we need shared ownership within the LRU [block cache](https://github.com/torquem-ch/silkworm/blob/master/silkworm/silkrpc/daemon.cpp#L252), which is available as shared service in API handlers (`EthereumRpcApi`, `DebugRpcApi`...) and can be accessed by different `Context` threads

PR #1448, opened 2023-08-12
rpcdaemon: extract backend mock

lupin012:


canepat (changes requested):
Please fix the memory leak as in merge from master also in the following files:
- engine_api_test.cpp
- evm_debug_test.cpp
- evm_executor_test.cpp
- evm_trace_test.cpp

canepat (approved):

PR #1453, opened 2023-08-15
rpcdaemon: remove core::rawdb::read header, read_body, read_block, read_senders

lupin012:


canepat (changes requested):
Please always use `Task` instead of `boost::asio::awaitable` now that PR #1435 has been merged and adjust the include directives accordingly.

canepat (approved):

PR #1478, opened 2023-08-30
rpcdaemon: add checks on no block retrieved from chain

lupin012:


silkworm/silkrpc/core/account_dumper.cpp
```diff
@@ -49,6 +49,10 @@ Task<DumpAccounts> AccountDumper::dump_accounts(
     const auto chain_storage = transaction_.create_storage(tx_database, backend);
 
     const auto block_with_hash = co_await core::read_block_by_number_or_hash(cache, *chain_storage, tx_database, bnoh);
+    if (!block_with_hash) {
+        SILK_ERROR << "dump_accounts:  core::read_block_by_number_or_hash  block not found\n";
+        co_return dump_accounts;
+    }
```
canepat, line 55: I think we need an explicit error handling strategy here (i.e. exception or more specific return value) rather than just returning a valid return value when we detect an error.

Moreover, I don't think we really need to mark this as an error in our logs because it depends on external input, which can be whatever and so this is not really an internal error. I would demote this log trace to `SILK_TRACE`.

silkworm/silkrpc/core/block_reader.cpp
```diff
@@ -45,6 +45,10 @@ Task<void> BlockReader::read_balance_changes(BlockCache& cache, const BlockNumbe
     ethdb::TransactionDatabase tx_database{transaction_};
 
     const auto block_with_hash = co_await core::read_block_by_number_or_hash(cache, chain_storage_, tx_database, bnoh);
+    if (!block_with_hash) {
+        SILK_ERROR << "read_balance_changes: core::read_block_by_number_or_hash block not found";
+        co_return;
+    }
```
canepat, line 51: Same as above

silkworm/silkrpc/core/call_many.cpp
```diff
@@ -167,6 +167,10 @@ Task<CallManyResult> CallExecutor::execute(
     const auto chain_config_ptr = co_await chain_storage->read_chain_config();
 
     const auto block_with_hash = co_await rpc::core::read_block_by_number_or_hash(block_cache_, *chain_storage, tx_database, context.block_number);
+    if (!block_with_hash) {
+        result.error = "block not found";
+        co_return result;
+    }
```
canepat, line 173: I do think here the error handling strategy is indeed correct because this a special case, isn't it?

silkworm/silkrpc/core/evm_trace.cpp
```diff
@@ -1566,7 +1566,17 @@ Task<void> TraceCallExecutor::trace_filter(const TraceFilter& trace_filter, cons
...
+        const Error error{-32000, "invalid parameters: fromBlock not found"};
+        stream->write_field("error", error);
+        co_return;
+    }
     const auto to_block_with_hash = co_await core::read_block_by_number_or_hash(block_cache_, storage, database_reader_, trace_filter.to_block);
+    if (!to_block_with_hash) {
+        const Error error{-32000, "invalid parameters: fromBlock not found"};
+        stream->write_field("error", error);
+        co_return;
+    }
```
canepat, line 1579: Another special case: streaming API. It seems OK here.

silkworm/silkrpc/core/logs_walker.cpp
```diff
@@ -144,6 +144,10 @@ Task<void> LogsWalker::get_logs(std::uint64_t start, std::uint64_t end,
 
         if (!filtered_block_logs.empty()) {
             const auto block_with_hash = co_await core::read_block_by_number(block_cache_, *chain_storage, block_to_match);
+            if (!block_with_hash) {
+                SILK_ERROR << "LogsWalker::get_logs core::read_block_by_number block not found";
+                co_return;
+            }
```
canepat, line 150: We need a more explicit error handling strategy here, instead.

canepat (changes requested):

silkworm/silkrpc/core/account_dumper.cpp
lupin012: ok

silkworm/silkrpc/core/block_reader.cpp
lupin012: ok

silkworm/silkrpc/core/call_many.cpp
lupin012: ok

silkworm/silkrpc/core/evm_trace.cpp
lupin012: yes

silkworm/silkrpc/core/logs_walker.cpp
lupin012: ok

canepat (approved):

PR #1480, opened 2023-08-31
rpcdaemon: add "don't care" expected response in integration tests

lupin012:


tests/integration/goerly/eth_callBundle/test_8.json
```diff
@@ -9,10 +9,7 @@
         "response": {
             "id": 1,
             "jsonrpc": "2.0",
-            "error": {
-                "code": -32000,
-                "message": "fee cap less than block base fee: address 0xf8561dc64e577b0cf53dd3fc11329e80b1a8343e, gasFeeCap: 3000000014 baseFee: 5657547293"
-            }
+            "any":null
```
canepat, line 12: I would avoid to add a non-standard JSON RPC field (`any`) to handle this case. I propose instead to use a response containing neither `result` nor `error` fields as don't-care

canepat (changes requested):

tests/integration/goerly/eth_callBundle/test_8.json
lupin012: ok

canepat (approved):

PR #1482, opened 2023-08-31
rpcdaemon: implementing HTTP CORS for RPC API

dario1995:


silkworm/silkrpc/settings.hpp
```diff
@@ -34,6 +34,7 @@ struct DaemonSettings {
     std::string eth_api_spec{kDefaultEth1ApiSpec};
     std::string private_api_addr{kDefaultPrivateApiAddr};
     uint32_t num_workers{std::thread::hardware_concurrency() / 2};
+    std::vector<std::string> cors_domain;
```
canepat, line 37: Please add `#include <vector>` because it is used here

silkworm/silkrpc/http/server.hpp
```diff
@@ -72,6 +73,9 @@ class Server {
     //! The acceptor used to listen for incoming TCP connections
     boost::asio::ip::tcp::acceptor acceptor_;
 
+    //! The list of allowed origins for CORS
+    std::vector<std::string>& allowed_origins_;
```
canepat, line 77: I would keep this by value here to avoid the lifecycle dependency with the caller implied by the usage of reference. The same applies to the constructor argument, which in turn can be passed by value and then moved into the member variable.

silkworm/silkrpc/http/connection.hpp
```diff
@@ -50,6 +50,7 @@ class Connection {
     Connection(boost::asio::io_context& io_context,
                commands::RpcApi& api,
                commands::RpcApiTable& handler_table,
+               std::vector<std::string>& allowed_origins,
```
canepat, line 53: I suggest changing this to a `const` reference to clearly communicate to the reader and also enforce that `Connection` won't change the allowed origins

silkworm/silkrpc/http/request_handler.hpp
```diff
@@ -77,6 +80,8 @@ class RequestHandler {
     const commands::RpcApiTable& rpc_api_table_;
 
     const std::optional<std::string> jwt_secret_;
+
+    std::optional<std::vector<std::string>> allowed_origins_;
```
canepat, line 84: It seems not necessary to make this field optional because I don't see any difference between `absence of value` (i.e. no CORS enabled) and `empty allowed origins` (i.e. no allowed origins, which again means no CORS). So, I suggest changing this to `const std::vector<std::string>&`.

silkworm/silkrpc/commands/rpc_api_test.cpp
```diff
@@ -245,7 +245,7 @@ class RequestHandler_ForTest : public silkworm::rpc::http::RequestHandler {
                            commands::RpcApi& rpc_api,
                            const commands::RpcApiTable& rpc_api_table,
                            std::optional<std::string> jwt_secret)
-        : silkworm::rpc::http::RequestHandler(socket, rpc_api, rpc_api_table, std::move(jwt_secret)) {
+        : silkworm::rpc::http::RequestHandler(socket, rpc_api, rpc_api_table, std::nullopt, std::move(jwt_secret)) {
```
canepat, line 248: I would change this `std::nullopt` with an empty vector of allowed origins (see comments below on the same topic).

canepat (changes requested):

silkworm/silkrpc/commands/rpc_api_test.cpp
dario1995: If I use an empty vector here it means the request_handler class must get it and save it by value

canepat (approved):

PR #1485, opened 2023-09-01
infra: setup GRPC logging in log::init

battlmonstr:


cmd/api/execute.cpp
```diff
@@ -182,6 +182,7 @@ int main(int argc, char* argv[]) {
         ExecuteSettings settings;
         parse_command_line(argc, argv, app, settings);
 
+        settings.log_settings.log_grpc = false;
```
canepat, line 185: IMHO setting this `log_rpc` flag to `false` in all these places is not needed. When it's `true` (as per default value) we're just redirecting any gRPC error to Silkworm log rather than to application console, so it's suitable for any case I think.

canepat (changes requested):

cmd/api/execute.cpp
battlmonstr: 👍 updated

canepat (approved):

PR #1514, opened 2023-09-12
Implement transient storage (EIP-1153)

chfast:
This implements the EIP-1153: "Transient storage opcodes"
https://eips.ethereum.org/EIPS/eip-1153.

- The `IntraBlockState` is extended with transient storage
  map account => (map key => value).
- The state journal is able to track transient storage changes with
  new class `TransientStorageChangeDelta`.
- evmone/EVMC are upgraded.

This has been tested with EIP-4788 disabled on:
- https://github.com/ethereum/tests/tree/develop/EIPTests/BlockchainTests/bcEIP1153-transientStorage
- https://github.com/ethereum/tests/tree/develop/EIPTests/BlockchainTests/StateTests/stEIP1153-transientStorage

### DIFF
--- a/silkworm/core/state/intra_block_state.cpp
+++ b/silkworm/core/state/intra_block_state.cpp
@@ -301,6 +301,17 @@ void IntraBlockState::set_storage(const evmc::address& address, const evmc::byte
     journal_.emplace_back(new state::StorageChangeDelta{address, key, prev});
 }
 
+evmc::bytes32 IntraBlockState::get_transient_storage(const evmc::address& addr, const evmc::bytes32& key) {
+    return transient_storage_[addr][key];
+}
+
+void IntraBlockState::set_transient_storage(const evmc::address& addr, const evmc::bytes32& key, const evmc::bytes32& value) {
+    auto& v = transient_storage_[addr][key];
+    const auto prev = v;
+    v = value;
+    journal_.emplace_back(std::make_unique<state::TransientStorageChangeDelta>(addr, key, prev));
+}
+
 void IntraBlockState::write_to_db(uint64_t block_number) {
     db_.begin_block(block_number);
 
@@ -371,6 +382,8 @@ void IntraBlockState::clear_journal_and_substate() {
     // EIP-2929
     accessed_addresses_.clear();
     accessed_storage_keys_.clear();
+
+    transient_storage_.clear();
 }
 
 void IntraBlockState::add_log(const Log& log) noexcept { logs_.push_back(log); }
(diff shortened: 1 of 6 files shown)


silkworm/core/state/intra_block_state.cpp
```diff
@@ -301,6 +301,17 @@ void IntraBlockState::set_storage(const evmc::address& address, const evmc::byte
...
 
+evmc::bytes32 IntraBlockState::get_transient_storage(const evmc::address& addr, const evmc::bytes32& key) {
+    return transient_storage_[addr][key];
+}
+
+void IntraBlockState::set_transient_storage(const evmc::address& addr, const evmc::bytes32& key, const evmc::bytes32& value) {
+    auto& v = transient_storage_[addr][key];
+    const auto prev = v;
+    v = value;
+    journal_.emplace_back(std::make_unique<state::TransientStorageChangeDelta>(addr, key, prev));
```
chfast, line 312: I have not verified this. Is journal empty at the start of every transaction?

silkworm/core/state/intra_block_state.cpp
```diff
@@ -301,6 +301,17 @@ void IntraBlockState::set_storage(const evmc::address& address, const evmc::byte
     journal_.emplace_back(new state::StorageChangeDelta{address, key, prev});
 }
 
+evmc::bytes32 IntraBlockState::get_transient_storage(const evmc::address& addr, const evmc::bytes32& key) {
+    return transient_storage_[addr][key];
```
chfast, line 305: This is "minimal" implementation. But we probably don't want to create map entries on every access for null value. Not sure about address though.

silkworm/core/state/intra_block_state.hpp
```diff
@@ -112,6 +112,10 @@ class IntraBlockState {
 
     const FlatHashSet<evmc::address>& touched() const noexcept { return touched_; }
 
+    evmc::bytes32 get_transient_storage(const evmc::address& address, const evmc::bytes32& bytes32);
```
yperbasis, line 115: Better name the last argument `key` rather than `bytes32`

silkworm/core/state/intra_block_state.cpp
yperbasis: Yes, `ExecutionProcessor::execute_transaction` calls `clear_journal_and_substate`

silkworm/core/state/intra_block_state.cpp
chfast: This is not a problem?

yperbasis: It's probably fine. I'd keep it simple for now. If it turns out to be a bottleneck in performance profiling, we can optimize this later.

PR #1515, opened 2023-09-13
sentry: refactor string to Bytes conversions

battlmonstr:
Group `char*` to ByteView conversions.


yperbasis (changes requested):
2 functions in `silkworm/sentry/common/string_bytes.hpp` are duplicates of existing functions in `silkworm/core/common/cast.hpp`. Please use those and move the remaining function into `common/cast.hpp`.

yperbasis (approved):

PR #1517, opened 2023-09-13
Move destruct_suicides & destruct_touched_dead into finalize_transaction

yperbasis:


silkworm/core/state/intra_block_state.cpp
```diff
@@ -351,7 +355,11 @@ void IntraBlockState::revert_to_snapshot(const IntraBlockState::Snapshot& snapsh
     logs_.resize(snapshot.log_size_);
 }
 
-void IntraBlockState::finalize_transaction() {
+void IntraBlockState::finalize_transaction(evmc_revision rev) {
+    destruct_suicides();
+    if (rev >= EVMC_SPURIOUS_DRAGON) {
+        destruct_touched_dead();
```
chfast, line 361: Fun fact: after Byzantium where intermediate state hashes are not part of the receipts, you can do this once per block.

PR #1535, opened 2023-09-17
rpcdaemon: fix trace_filter without reward and after != 0

lupin012:


silkworm/silkrpc/core/evm_trace.cpp
```diff
@@ -1188,11 +1188,11 @@ Task<std::vector<Trace>> TraceCallExecutor::trace_block(const BlockWithHash& blo
         co_return traces;
     }
 
-    if (filter.count > 0 && filter.after == 0) {
-        const auto chain_config_ptr = co_await chain_storage_.read_chain_config();
-        const auto rule_set_factory = protocol::rule_set_factory(*chain_config_ptr);
-        const auto block_rewards = rule_set_factory->compute_reward(block_with_hash.block);
+    const auto chain_config_ptr = co_await chain_storage_.read_chain_config();
```
canepat, line 1191: You should check an invariant here (i.e. `ensure`) to avoid a crash if for some unlikely reason `chain_config_ptr` does not contain a valid chain configuration. You can do it in next PR.

We should always strive to leave the code we touch in a state slightly better than we have found it, if necessary.

canepat (approved):

silkworm/silkrpc/core/evm_trace.cpp
lupin012: ok

PR #1548, opened 2023-09-21
Huffman insert word without shared ptr

lupin012:


silkworm/node/huffman/decompressor.cpp
```diff
@@ -144,30 +144,34 @@ std::size_t PatternTable::build_condensed(std::span<Pattern> patterns, uint64_t
...
         const auto code_from = codeword->code();
         const auto code_to =
             bit_length_ != codeword->code_length() && codeword->code_length() > 0 ? code_from | 1 << bit_length_ : code_from + code_step;
         for (auto c{code_from}; c < code_to; c += code_step) {
-            auto* cw = codewords_[c].get();
+            auto cw = codewords_[c];
             if (cw == nullptr) {
                 codewords_[c] = codeword;
-                inserted = codewords_[c].get();
+                inserted = codewords_[c];
```
mriccobene, line 174: IMO the line 
`codewords_[c] = codeword;`
will work very differently: the same pointer will be inserted in different and unrelated smart_ptrs leading to a double free error.

mriccobene (changes requested):

mriccobene (approved):

PR #1554, opened 2023-09-26
cmake: refactor third_party/CMakeLists.txt

battlmonstr:


canepat (changes requested):
I can understand the logical reason why you moved the `erigon-snapshot` submodule out of the `third-party` folder, nevertheless IMHO it's better to have *all* submodules in one place so that it's a no brainer where to go for looking at or updating one submodule.
Moreover, `erigon-snapshot` is really a third-party module for this project anyway.

canepat (approved):

PR #1561, opened 2023-09-27
sentry: disc v4 ping back

battlmonstr:
Add a new node to the ping check queue and ping immediately.
This is needed for erigon's [ensureBond()](https://github.com/ledgerwatch/erigon/blob/d62ef03cad02754795199c5e00d01fb53eeecaeb/p2p/discover/v4_udp.go#L600) to work,
otherwise ENRRequest and FindNode requests from new nodes are ignored
(even if they ping us first).
See erigon's v4_udp.go handlePing "Ping back":
https://github.com/ledgerwatch/erigon/blob/d62ef03cad02754795199c5e00d01fb53eeecaeb/p2p/discover/v4_udp.go#L701

bug fix:
Sending multiple UDP messages didn't work,
because bind() fails with "address is busy".
We have to use the same port on both client and server sockets,
so instead let's reuse the socket.

### DIFF
--- a/silkworm/sentry/discovery/disc_v4/discovery.cpp
+++ b/silkworm/sentry/discovery/disc_v4/discovery.cpp
@@ -22,6 +22,7 @@
 
 #include <silkworm/infra/common/log.hpp>
 #include <silkworm/infra/concurrency/awaitable_wait_for_all.hpp>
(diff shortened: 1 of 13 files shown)


canepat (approved):
LGTM

PR #1571, opened 2023-10-04
core: start implementing BorRuleSet

yperbasis:
Continuation of PR #1552 (one of many to add support of Bor/Polygon PoS to Silkworm).
 
`RuleSetType` enum is replaced with `std::variant<EthashConfig, CliqueConfig, BorConfig>`.  `kNoProof` is merged with `kEthash` into `EthashConfig`.


silkworm/core/chain/genesis.cpp
```diff
@@ -45,19 +45,19 @@ namespace silkworm {
...
-        case kSepoliaConfig.chain_id:
+        case 11155111:
             assert(sizeof_genesis_sepolia_data() != 0);
             return {genesis_sepolia_data(), sizeof_genesis_sepolia_data()};
-        case kPolygonConfig.chain_id:
+        case 137:
             assert(sizeof_genesis_polygon_data() != 0);
             return {genesis_polygon_data(), sizeof_genesis_polygon_data()};
-        case kMumbaiConfig.chain_id:
+        case 80001:
```
canepat, line 60: Is hard-coding the various chain id here related to `constinit` usage?

silkworm/core/chain/config.hpp
```diff
@@ -119,95 +127,19 @@ struct ChainConfig {
...
-    .byzantium_block = 0,
-    .constantinople_block = 0,
-    .petersburg_block = 0,
-    .istanbul_block = 2'722'000,
-    .muir_glacier_block = 2'722'000,
-    .berlin_block = 13'996'000,
-    .london_block = 22'640'000,
-    .eip1559_fee_collector = 0x70bca57f4579f58670ab2d18ef16e02c17553c38_address,
-};
+extern const ChainConfig kMumbaiConfig;
```
canepat, line 142: Why not `constinit` for `kPolygonConfig` and `kMumbaiConfig`?

canepat (approved):

silkworm/core/chain/config.hpp
yperbasis: > Why not `constinit` for `kPolygonConfig` and `kMumbaiConfig`?

Because it's not compatible with `std::map` unfortunately. Strictly speaking, using `std::map` in static variables is in violation of the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables), but I'm thinking that it's not that big a deal.

silkworm/core/chain/genesis.cpp
yperbasis: > Is hard-coding the various chain id here related to `constinit` usage?

It's because the config constants are not `constexpr` anymore.

PR #1575, opened 2023-10-05
rpcdaemon: add fuzzer tests

JacekGlen:
This is the first PR in the series related to fuzzy tests.

This adds two executables:
- rpcdaemon_fuzzer_test
- rpcdaemon_fuzzer_diagnostics

The builds are available for CLang and sanitiser builds only.

The `rpcdaemon_fuzzer_test` is a fuzzing test for all RPC endpoints. Currently, it only uses generic fuzzing, specific corpuses to be delivered in later PRs.
The `rpcdaemon_fuzzer_diagnostics` is a tool to run an individual request in a manner similar to a fuzzy test.


silkworm/silkrpc/test/api_test_database.hpp
```diff
@@ -0,0 +1,280 @@
...
+        .path = chaindata_dir,
+        .create = true,
+        .exclusive = true,
+        .in_memory = true,
+        .shared = false};
+
+    return db::open_env(chain_conf);
+}
+
+InMemoryState populate_genesis(db::RWTxn& txn, const std::filesystem::path& tests_dir) {
```
canepat, line 68: This `populate_genesis` function looks very similar if not identical to the same function in `rpc_api_test`, I think we have room for some good refactoring here. Next PR.

canepat (approved):

PR #1584, opened 2023-10-08
core: implement BorRuleSet::get_beneficiary & add_fee_transfer_log

yperbasis:


silkworm/core/protocol/clique_rule_set.cpp
```diff
@@ -30,11 +30,11 @@ static evmc::address ecrecover(const BlockHeader& header) {
...
 
     evmc::bytes32 seal_hash = header.hash(false, true);
-    Bytes signature = header.extra_data.substr(header.extra_data.length() - kExtraSealSize, kExtraSealSize - 1);
-    bool odd_y_parity = header.extra_data[header.extra_data.length() - 1] != 0;
+    ByteView signature{&header.extra_data[header.extra_data.length() - kExtraSealSize], kExtraSealSize - 1};
+    uint8_t recovery_id{header.extra_data[header.extra_data.length() - 1]};
 
     static secp256k1_context* context{secp256k1_context_create(SILKWORM_SECP256K1_CONTEXT_FLAGS)};
-    if (!silkworm_recover_address(beneficiary.bytes, seal_hash.bytes, signature.c_str(), odd_y_parity, context)) {
+    if (!silkworm_recover_address(beneficiary.bytes, seal_hash.bytes, signature.data(), recovery_id, context)) {
```
canepat, line 37: Currently, the `silkworm_recover_address` function has `bool odd_y_parity` as fourth formal parameter. It calls the `recover` function, which in turns calls `secp256k1_ecdsa_recoverable_signature_parse_compact` passing the received `odd_y_parity` as actual parameter for `int recid`.

So, I see why you are changing the parameter type and name here. I think in another PR we should change also the formal parameter type and name in `silkworm_recover_address` trying to accommodate all needs (i.e. somewhere else it's called with `v != 27`).

canepat (changes requested):

silkworm/core/protocol/bor_rule_set.cpp
```diff
@@ -30,20 +40,95 @@ ValidationResult BorRuleSet::validate_block_header(const BlockHeader& header, co
...
+
+    if (header.extra_data.length() < kExtraVanityLength) {
         return ValidationResult::kMissingVanity;
     }
-    if (header.extra_data.length() < kExtraVanitySize + kExtraSealSize) {
+    if (header.extra_data.length() < kExtraVanityLength + kExtraSealSize) {
         return ValidationResult::kMissingSignature;
     }
+
+    const bool is_sprint_end{is_sprint_start(header.number + 1, config().sprint_size(header.number))};
```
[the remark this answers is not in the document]
yperbasis: Here the names look fine to me. The end of a sprint at block _n_ == the start of next sprint at block _n_+1. I'll add a comment.

silkworm/core/protocol/clique_rule_set.cpp
yperbasis: In this PR I've changed the parameter of `silkworm_recover_address` to `uint8_t recovery_id`. It's called with `v != 27` in the ecrec_run precompile, where `v` is checked to be either 27 or 28.

canepat: Sorry, I completely missed the change is already in this PR.

silkworm/core/protocol/bor_rule_set.cpp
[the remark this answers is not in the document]
canepat: OK, I see. Now I understand that `is_sprint_end` refers to the current sprint and the function call `is_sprint_start` with value `header.number + 1` checks "does the next sprint start at next block?".

canepat (approved):

PR #1589, opened 2023-10-13
core: revert fix for create-after-destruct

canepat:
This reverts the change from PR #1553 because it breaks the state trie root at block [116525](https://etherscan.io/block/116525)


yperbasis (approved):
I see. It's a shame it wasn't caught by any test.

Issue #1459, opened 2023-08-21
infra: Potential leak of memory in conversion.cpp


battlmonstr:
[conversion.cpp](https://github.com/torquem-ch/silkworm/blob/641509b40dc95da48e1cee2f52bef670f3dfde5d/silkworm/infra/grpc/common/conversion.cpp#L172) uses a bunch of naked `new` which is not exception-safe and results in the [clang-tidy warnings](https://output.circle-artifacts.com/output/job/71db91ac-7229-473b-811e-e1c3ab388e61/artifacts/0/tmp/report.html#) of type clang-analyzer-cplusplus.NewDelete and clang-analyzer-cplusplus.NewDeleteLeaks

Those could be rewritten with `unique_ptr` and `release()`.

vlnistor:
Hey, I'm looking for a good first issue. Can I be assigned to this

Claudio2000:
Is this still open? Also looking for a good first issue.

PR #1600, opened 2023-10-21
Replace new with unique_ptr and release #1459

claudioutt:
Replaced new with unique_ptr and release to combat potential memory leakage as Issue suggested.


battlmonstr (approved):
:+1:

PR #1604, opened 2023-10-23
rpcdaemon: Json version as stringview

lupin012:


canepat (approved):

silkworm/silkrpc/json/types.hpp
```diff
@@ -81,7 +81,7 @@ void to_json(nlohmann::json& json, const BlockHeader& ommer);
 
 namespace silkworm::rpc {
 
-#define jsonVersion "2.0"
+static constexpr std::string_view jsonVersion{"2.0"};
```
yperbasis, line 84: Better use `inline constexpr` rather than `static constexpr` for constants in headers. See https://abseil.io/tips/140

lupin012: ok, thanks I change SW

Issue #1398, opened 2023-08-02
Improve the error message: Address already in use


battlmonstr:
When starting silkworm without arguments after staring erigon without arguments on the same computer, silkworm crashes with "Unrecoverable failure: Address already in use" and no other clues. Apparently it wants to occupy some default TCP/UDP ports that are in use by the erigon process.

The error message needs to be improved to explain what port is occupied, what for is it needed, and which command line parameter must be passed to customize it.

battlmonstr:
PR 1: https://github.com/erigontech/silkworm/pull/1611
PR 2: https://github.com/erigontech/silkworm/pull/1788

battlmonstr:
One place where this happens is in the sentry GRPC servers. In this case BuildAndStart returns null here: https://github.com/erigontech/silkworm/blob/master/silkworm/infra/grpc/server/server.hpp#L84

It is possible to implement [GlobalCallbacks::AddPort](https://github.com/grpc/grpc/blob/0b1e381d56da805045ce4c89d47ca4f13549c251/include/grpcpp/server.h#L86), and call `grpc::Server::SetGlobalCallbacks(...)` to register the implementation. `BuildAndStart()` calls it [here](https://github.com/grpc/grpc/blob/0b1e381d56da805045ce4c89d47ca4f13549c251/src/cpp/server/server_builder.cc#L434) and that should trigger a [callback](https://github.com/grpc/grpc/blob/master/src/cpp/server/server_cc.cc#L1131). If `port == 0` in the `AddPort` callback, we can assume that the port is busy.

This gives information to extend the error message here: https://github.com/erigontech/silkworm/blob/master/silkworm/infra/grpc/server/server.hpp#L86

PR #1611, opened 2023-10-25
Add Err msg when Sentry RLPx port is occupied #1398

claudioutt:
Also improved occupied port 9090 error message.

### DIFF
--- a/silkworm/infra/grpc/server/server.hpp
+++ b/silkworm/infra/grpc/server/server.hpp
@@ -82,7 +82,7 @@ class Server {
         server_ = builder.BuildAndStart();
         SILK_TRACE << "Server " << this << " bound at selected port: " << selected_port;
         if (server_ == nullptr) {
-            SILK_ERROR << "Server " << this << " BuildAndStart failed [" << settings_.address_uri << "]";
+            SILK_ERROR << "Server " << this << " BuildAndStart failed. Cannot start gRPC server at [" << settings_.address_uri << "] Port is probably in use!";
             throw std::runtime_error("cannot start gRPC server at " + settings_.address_uri);
         }
 
--- a/silkworm/sentry/rlpx/server.cpp
+++ b/silkworm/sentry/rlpx/server.cpp
@@ -59,7 +59,15 @@ Task<void> Server::run(
     acceptor.set_option(detail::socket_option::boolean<SOL_SOCKET, SO_REUSEPORT>(true));
 #endif
 
-    acceptor.bind(endpoint);
+    try {
+        acceptor.bind(endpoint);
+    } catch (const boost::system::system_error& ex) {
+        if (ex.code() == boost::system::errc::address_in_use) {
+            throw std::runtime_error("Sentry RLPx server has failed to stat. Port: " + std::to_string(port_) + " is busy! Try another port with --port=MMM. Error: " + ex.what());
+        } else {
+            throw;
+        }
+    }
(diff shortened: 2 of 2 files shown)


silkworm/infra/grpc/server/server.hpp
```diff
@@ -82,7 +82,7 @@ class Server {
         server_ = builder.BuildAndStart();
         SILK_TRACE << "Server " << this << " bound at selected port: " << selected_port;
         if (server_ == nullptr) {
-            SILK_ERROR << "Server " << this << " BuildAndStart failed [" << settings_.address_uri << "]";
+            SILK_ERROR << "Server " << this << " BuildAndStart failed. Cannot start gRPC server at [" << settings_.address_uri << "] Port is probably in use!";
```
battlmonstr, line 85: is there a way to get an error from the builder?

silkworm/sentry/rlpx/server.cpp
```diff
@@ -59,7 +59,11 @@ Task<void> Server::run(
     acceptor.set_option(detail::socket_option::boolean<SOL_SOCKET, SO_REUSEPORT>(true));
 #endif
 
-    acceptor.bind(endpoint);
+    boost::system::error_code ec;
+    acceptor.bind(endpoint, ec);
+    if(ec){
+        SILK_ERROR << "Sentry RLPx server has failed to start. TCP port: "<< port_ << " is busy. Try another port with --port=MMM.";
```
battlmonstr, line 65: * we use exceptions, please catch this as `const boost::system::system_error&`
* the check should examine the ex.code() and/or message and understand that it is related to the busy port
* throw `std::runtime_error` in the end, no need to log here

battlmonstr (changes requested):

battlmonstr:
please run `make fmt` and commit for the CI to pass

silkworm/infra/grpc/server/server.hpp
claudioutt: I looked into exactly that. builder.BuildAndStart() just returns nullptr when an Error occurs.

battlmonstr: Try to implement [GlobalCallbacks::AddPort](https://github.com/grpc/grpc/blob/0b1e381d56da805045ce4c89d47ca4f13549c251/include/grpcpp/server.h#L86). Call `grpc::Server::SetGlobalCallbacks(...)` to register your implementation. `BuildAndStart()` calls it [here](https://github.com/grpc/grpc/blob/0b1e381d56da805045ce4c89d47ca4f13549c251/src/cpp/server/server_builder.cc#L434) and that should trigger a [callback](https://github.com/grpc/grpc/blob/master/src/cpp/server/server_cc.cc#L1131). If `port == 0` in the `AddPort` callback, we can assume that the port is busy.

Also please enhance the `std::runtime_error` thrown below with this extra information.

battlmonstr (changes requested):

battlmonstr:
@claudioutt do you have time to finish this in the upcoming weeks? We're considering to merge a part that is ready (rlpx), and keep the unfinished part (grpc) as a task for newcomers.

battlmonstr:
superceded by https://github.com/erigontech/silkworm/pull/1788

claudioutt:
Sorry I am currently really busy with work. Feel free to open the issue for new contributors.

PR #1618, opened 2023-10-27
core: fix ConfigMap conversion error in MSVC

canepat:
This PR fixes MSVC build broken on master after PR #1612. You can check [problem and solution in isolation](https://godbolt.org/z/oda56xWrc).


yperbasis (commented):
This looks ugly to me. I'd rather change the return type of `begin()` and `end()` to `auto` instead of `const ValueType*`

yperbasis (approved):

PR #1621, opened 2023-10-29
capi: block execution commits after state batch has been flushed

canepat:
This PR contains the following changes related to block execution API implementation:
- commit the current state when the batch size crosses over threshold
- commit the state history after each block 
- change the batch size unit from cumulated gas to bytes
- fix many batch size computations in db buffer
- avoid caching read values in db buffer
- sort storage locations to insert ordered data into db
- fix empty element in call traces
- fix unmanaged RW txn destruction

Moreover, some other things as well:
- improve logging in block execution
- add unit tests

All these changes have been made in the process of reaching full binary data compatibility between Erigon and Erigon+Silkworm, step by step, over a long sequence of comparative tests: that's why this PR required quite a long time to be finalised.

### DIFF
--- a/silkworm/capi/silkworm.cpp
+++ b/silkworm/capi/silkworm.cpp
@@ -80,10 +80,12 @@ static log::Args log_args_for_version() {
 }
 
 //! Generate log arguments for execution flush at specified block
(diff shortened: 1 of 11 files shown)


JacekGlen (approved):
Looks good! 
We probably need to check if this PR has any performance hit, but I can do it separately.

PR #1633, opened 2023-11-07
rpcdaemon: CBOR decode must be ok when encoded value is null

dario1995:


silkworm/silkrpc/core/rawdb/chain.cpp
```diff
@@ -172,14 +172,13 @@ Task<Receipts> read_raw_receipts(const DatabaseReader& reader, BlockNum block_nu
...
+        co_return receipts;
     }
-    Receipts receipts{};
     const bool decoding_ok{cbor_decode(data, receipts)};
     if (!decoding_ok) {
-        SILK_WARN << "cannot decode raw receipts in block: " << block_number;
-        co_return receipts;
+        throw std::runtime_error("cannot decode raw receipts in block: " + std::to_string(block_number));
     }
     SILK_DEBUG << "#receipts: " << receipts.size();
```
canepat, line 184: You can add here an early return when there are zero receipts to avoid the next `walk` call.

silkworm/silkrpc/core/rawdb/chain.cpp
```diff
@@ -172,14 +172,13 @@ Task<Receipts> read_raw_receipts(const DatabaseReader& reader, BlockNum block_nu
     const auto block_key = silkworm::db::block_key(block_number);
     const auto data = co_await reader.get_one(db::table::kBlockReceiptsName, block_key);
     SILK_TRACE << "read_raw_receipts data: " << silkworm::to_hex(data);
+    Receipts receipts{};
     if (data.empty()) {
-        co_return Receipts{};  // TODO(canepat): use std::null_opt with Task<std::optional<Receipts>>?
+        co_return receipts;
```
canepat, line 177: I think we need to change the return type of `read_raw_receipts` to be `Task<std::optional<Receipts>>` as per my comment and return `std::null_opt` here because this case is "there is no data for specified block in the db".

Then, we need to check the optional value at the call site in `get_receipts`.

All this thing is needed to distinguish among three cases:
- we have data in the db and it's valid
- we have no data in the db (i.e. pruning) and we may need to recompute the receipts by executing the tx
- we have an error decoding the data

silkworm/silkrpc/core/receipts.cpp
```diff
@@ -22,16 +22,7 @@
...
-    if (!cached_receipts.empty()) {
-        co_return cached_receipts;
-    }
-
-    // If not already present, retrieve receipts by executing transactions
-    // auto block = co_await core::rawdb::read_block(db_reader, hash, number);
-    // TODO(canepat): implement
-    SILK_WARN << "retrieve receipts by executing transactions NOT YET IMPLEMENTED";
-    co_return Receipts{};
+    co_return co_await core::rawdb::read_receipts(db_reader, block_with_hash);
```
canepat, line 25: I suggest to restore something similar to the previous code here with a single difference being the condition to check is on `cached_receipts` equal to `std::null_opt` or not. Then, you should re-add the warning because we do not currently have implemented transaction re-execution here, but at least empty receipts are returned with such warning when we should recompute them instead.

canepat (changes requested):

canepat (approved):

PR #1643, opened 2023-11-13
node: in-memory ETL collector

mriccobene:
This is a memory-only reduced version of ETL Collector, with a compatible interface, that I used to measure the performance gain or loss when using only the memory to perform ETL tasks.

### DIFF
--- a/silkworm/node/etl/mini_collector.hpp
+++ b/silkworm/node/etl/mini_collector.hpp
@@ -1,0 +1,154 @@
+/*
+Copyright 2023 The Silkworm Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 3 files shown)


silkworm/node/etl/mini_collector.hpp
```diff
@@ -0,0 +1,154 @@
+/*
```
canepat, line 1: I suggest using `InMemoryCollector` instead of `MiniCollector` and renaming all stuff (also header file) accordingly.

silkworm/node/etl/mini_collector.hpp
```diff
@@ -0,0 +1,154 @@
...
+    // for progress tracking only
+    void set_loading_key(ByteView key) {
+        std::unique_lock l{mutex_};
+        loading_key_ = to_hex(key, true);
+    }
+    mutable std::mutex mutex_{};  // To sync loading_key_
+    std::string loading_key_{};   // Actual load key (for log purposes)
+};
+
+}  // namespace silkworm::etl
```
canepat, line 154: Please add also a test module `in_memory_collector_test.cpp`.

canepat (changes requested):

silkworm/node/etl/mini_collector.hpp
mriccobene: done

silkworm/node/etl/mini_collector.hpp
mriccobene: I have done the renaming; I also have made the collector container a template parameter and provided two implementations, one based on map and one based on vector, with different memory footprints.

canepat (approved):

PR #1661, opened 2023-11-17
cmd: add compare subcommand to db toolbox

canepat:
cmd: fix layout for table dump in db toolbox tool
cmd: add log trace in snapshots tool

The new `compare` subcommand is currently quite basic but it has already proven useful to diagnose some database discrepancies.


mriccobene (approved):

AndreaLanfranchi:
👏

PR #1671, opened 2023-11-23
node: treat clang-tidy warnings as errors

yperbasis:


silkworm/node/stagedsync/execution_engine.cpp
```diff
@@ -64,7 +64,7 @@ void ExecutionEngine::insert_blocks(const std::vector<std::shared_ptr<Block>>& b
     }
 }
 
-bool ExecutionEngine::insert_block(const std::shared_ptr<Block> block) {
+bool ExecutionEngine::insert_block(const std::shared_ptr<Block>& block) {
```
mriccobene, line 67: I don't always agree with clang-tidy's suggestions. For example here clang-tidy suggests us to use const& but I consider passing shared_ptr by reference a wrong pattern: shared_ptr is built to be copied, otherwise we need to use it carefully, for example paying attention to not to pass the reference to shared_ptr to a closure that than is pushed to a thread.

silkworm/node/stagedsync/forks/main_chain.cpp
```diff
@@ -281,7 +281,7 @@ bool MainChain::notify_fork_choice_update(Hash head_block_hash, std::optional<Ha
 
         const auto finalized_block_number = get_block_number(*finalized_block_hash);
         last_finalized_head_.number = *finalized_block_number;
-        last_finalized_head_.hash = std::move(*finalized_block_hash);
+        last_finalized_head_.hash = *finalized_block_hash;
```
mriccobene, line 284: This is another example where I do not like the clang-tidy rules: it is true that _at the moment_ finalized_block_hash; is trivially-copyable so we do not need std::move but what happens if someone changes the type of finalized_block_hash making it non trivially-copyable? it will be very difficult to find all the places where a non-trivial copy is made and it is better to do a move. 
So I consider "performance defensive programming"  to code the above line as before:

`last_finalized_head_.hash = std::move(*finalized_block_hash);`

it is not an error, it states that we no longer need finalized_block_hash and it can be moved. It will be a performance gain in a future where finalized_block_hash is not trivially copyable.

silkworm/node/stagedsync/execution_engine.cpp
wezrule: There's a nice article here by Herb Sutter about smart pointer parameter passing https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

mriccobene: I know, thanks. Not many agree with those rules because they clash with past conventions. In any case, the point is another: is a rule encoded in clang-tidy always valid, regardless of the context, so that we can rely on clang-tidy to ensure that it is respected?

yperbasis: I think it's fine to follow Herb's advise since we don't have any legacy code in Silkworm and thus past conventions aren't relevant to us.

clang-tidy isn't perfect, but it can be tuned to be of net benefit. We can disable any problematic check altogether in `.clang-tidy` or at a specific code site with `// NOLINT`. To my mind we should [use the tools available](https://github.com/cpp-best-practices/cppbestpractices/blob/master/02-Use_the_Tools_Available.md). C++ is a complicated beast, so we should use all the help we can get to rein it in :)

silkworm/node/stagedsync/forks/main_chain.cpp
yperbasis: Oh, in that case there's another clang-tidy check that'd suggest to use an `std::move`. So the answer is more clang-tidy, not less ;)

canepat (approved):

silkworm/node/stagedsync/forks/main_chain.cpp
mriccobene: this might convince me ;-)

PR #1677, opened 2023-11-29
rpcdaemon: Add tx hash on cache

lupin012:


silkworm/core/types/transaction.cpp
```diff
@@ -17,6 +17,7 @@
 #include "transaction.hpp"
 
 #include <bit>
+#include <iostream>
```
canepat, line 20: This new include is not necessary

silkworm/core/types/transaction.hpp
```diff
@@ -87,6 +87,8 @@ struct Transaction : public UnsignedTransaction {
 
     // sender recovered from the signature
     std::optional<evmc::address> from{std::nullopt};
+    // store calculated hash
+    mutable std::optional<evmc::bytes32> cached_hash;
```
canepat, line 91: I suggest to initialise `cached_hash` with `std::nullopt` as done for `from` field. This keeps the type definition more coherent and also seems to be still preferable [at least in GCC](https://godbolt.org/z/ooMq1TY8x)

canepat (changes requested):

silkworm/core/types/transaction.hpp
lupin012: ok

canepat:
Purely in terms of the `Transaction` type definition, this change is not ideal because IMO such abstract data type now lacks a bit of information hiding:

- all fields are public, so can be modified without any chance to ensure invariants
- the `hash` method now fills and reuses a cached value

This can easily lead to invariant break, e.g. if you call the `hash` method on a `Transaction` object, then change any field and call the `hash` method again obtaining an invalid hash value.


On the other hand, such use cases are not frequent (currently we have none), so I'm going to merge this PR as it is and open an issue to track the need to refactor `Transaction` into a class with proper encapsulation.

lupin012:
> Please reset this field in `decode_transaction`
lupin : thanks
The fix is ​​a little bit more "complex" because with the addition of a private member the compiler no longer accepts brace-enclosed initialization and therefore I have to add the constructors to Transaction() and UnassignedTransaction() and I have to call them where before was used brace-enclosed initialization.

canepat (approved):

yperbasis (approved):

PR #1710, opened 2023-12-13
core: exclude tracing for create/create2 failing pre-checks

canepat:
This PR fixes yet another incompatibility between Silkworm and Erigon database content related to call traces.

Specifically, the discrepancy happens on mainnet at block [1878695](https://etherscan.io/block/1878695), where transaction [5](https://etherscan.io/tx/0x0ccccec7ba6a56abfcf0c5c82688de0ddfdeee3176316cf69ff39797b654d308) runs out of gas trying to deploy contract [0xAE050f0402c2F4C7fe0Abb0ea1F0A466061AE967](https://etherscan.io/address/0xAE050f0402c2F4C7fe0Abb0ea1F0A466061AE967): the failure is triggered within `evmone::baseline` in function [`check_requirements`](https://github.com/ethereum/evmone/blob/781875d95dd186ab1bf36712979b479f884a3748/lib/evmone/baseline.cpp#L147) for opcode `CREATE`.

In this specific case, Erigon does not include the `CREATE` sender and recipient into the call trace set for the block. On the other hand, Silkworm does after the changes introduced by #1706, namely using the `on_instruction_start` callback to trace sender and recipient of early-returning `CREATE` operations. Hence, given that we need to correlate the out-of-gas failure with the `CREATE` opcode execution, what we can do with the current tracing interface is just reproducing the same requirement failure here.

Moreover the `check_requirements` function, which detects the out-of-gas condition during pre-execution checks, is currently not exported by `evmone`, so this PR temporarily copies the code here.

@chfast for a proper solution we either need to export the `check_requirements` function from `evmone` or even better to extend the tracing interface (perhaps by adding `on_instruction_end` hook or carrying over the last executed opcode in `on_execution_end`? not sure) so that we don't need to redo the check

All the above considerations can be applied to the `CREATE2` operation as well.

### DIFF
--- a/silkworm/core/execution/call_tracer.cpp
+++ b/silkworm/core/execution/call_tracer.cpp
@@ -68,7 +68,7 @@ inline evmc_status_code check_requirements(const CostTable& cost_table, int64_t&
 
     return EVMC_SUCCESS;
 }
-}
+}  // namespace
 
 namespace silkworm {
 
--- a/silkworm/core/execution/call_tracer.cpp
+++ b/silkworm/core/execution/call_tracer.cpp
@@ -18,10 +18,57 @@
 
 #include <evmc/hex.hpp>
 #include <evmc/instructions.h>
+#include <evmone/baseline_instruction_table.hpp>
 #include <evmone/execution_state.hpp>
 #include <evmone/instructions.hpp>
 
 #include <silkworm/core/types/address.hpp>
+
+using namespace evmone;
+using namespace evmone::baseline;
+
+// The following check_requirements function has been temporarily copied from evmone because it is not exported.
+// We need to ask evmone for it to be exported or for a tracing interface extension (e.g. on_instruction_end?).
+namespace {
+template <Opcode Op>
+inline evmc_status_code check_requirements(const CostTable& cost_table, int64_t& gas_left,
+                                           const uint256* stack_top, const uint256* stack_bottom) noexcept {
+    static_assert(
+        !instr::has_const_gas_cost(Op) || instr::gas_costs[EVMC_FRONTIER][Op] != instr::undefined,
+        "undefined instructions must not be handled by check_requirements()");
+
+    auto gas_cost = instr::gas_costs[EVMC_FRONTIER][Op];  // Init assuming const cost.
+    if constexpr (!instr::has_const_gas_cost(Op)) {
+        gas_cost = cost_table[Op];  // If not, load the cost from the table.
+
+        // Negative cost marks an undefined instruction.
+        // This check must be first to produce correct error code.
+        if (INTX_UNLIKELY(gas_cost < 0))
+            return EVMC_UNDEFINED_INSTRUCTION;
+    }
+
+    // Check stack requirements first. This is order is not required,
+    // but it is nicer because complete gas check may need to inspect operands.
+    if constexpr (instr::traits[Op].stack_height_change > 0) {
+        static_assert(instr::traits[Op].stack_height_change == 1,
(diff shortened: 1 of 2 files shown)


silkworm/core/execution/call_tracer.cpp
```diff
@@ -18,11 +18,58 @@
...
 #include <silkworm/core/types/address.hpp>
 
+using namespace evmone;
+using namespace evmone::baseline;
+
+// The following check_requirements function has been temporarily copied from evmone because it is not exported.
+// We need to ask evmone for it to be exported or for a tracing interface extension (e.g. on_instruction_end?).
+namespace {
+template <Opcode Op>
+inline evmc_status_code check_requirements(const CostTable& cost_table, int64_t& gas_left,
```
JacekGlen, line 34:
```suggestion
[[deprecated("Temporary fix, await tracing interface extension")]]
inline evmc_status_code check_requirements(const CostTable& cost_table, int64_t& gas_left,
```

silkworm/core/execution/call_tracer.cpp
```diff
@@ -18,11 +18,58 @@
...
 #include <evmone/execution_state.hpp>
 #include <evmone/instructions.hpp>
 
 #include <silkworm/core/types/address.hpp>
 
+using namespace evmone;
+using namespace evmone::baseline;
+
+// The following check_requirements function has been temporarily copied from evmone because it is not exported.
+// We need to ask evmone for it to be exported or for a tracing interface extension (e.g. on_instruction_end?).
```
JacekGlen, line 31: My vote is for the tracing interface extension, seems to be more aligned with the current code

silkworm/core/execution/call_tracer.cpp
```diff
@@ -38,25 +85,33 @@ void CallTracer::on_execution_start(evmc_revision /*rev*/, const evmc_message& m
...
 
-void CallTracer::on_instruction_start(uint32_t pc, const intx::uint256* stack_top, int stack_height, int64_t /*gas*/,
+void CallTracer::on_instruction_start(uint32_t pc, const intx::uint256* stack_top, int stack_height, int64_t gas,
                                       const evmone::ExecutionState& state, const IntraBlockState& intra_block_state) noexcept {
     const auto op_code = state.original_code[pc];
     if (op_code == evmc_opcode::OP_CREATE) {
+        const auto& cost_table{get_baseline_cost_table(state.rev, state.analysis.baseline->eof_header.version)};
+        if (const auto status{check_requirements<Opcode::OP_CREATE>(cost_table, gas, stack_top, stack_top - stack_height)}; status != EVMC_SUCCESS) {
+            return;  // Early failure in pre-execution checks, do not trace anything for compatibility w/ Erigon
+        }
```
JacekGlen, line 95: Duplicated code, can we move `cost_table` and `status` to check_requirements? Then we would have something like:
```
if (!requirements_checked<Opcode::OP_CREATE>(state, gas, stack_top, stack_heightd) {
    return; // Early failure in pre-execution checks, do not trace anything for compatibility w/ Erigon
}
```

JacekGlen (approved):

silkworm/core/execution/call_tracer.cpp
canepat: I tried to include this `deprecated` attribute but then the build fails because we do not accept any deprecated code call, I forgot about it. So, given that there's a comment clarifying the presence of the `check_requirements` function here, I decided not to include the attribute in order not to make this translation unit too cluttered.

Issue #1733, opened 2024-01-03
rpcdaemon: invalid request passes validation and returns data


JacekGlen:
The following request has an invalid parameter:
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "eth_getBlockReceipts",
    "params": [
        "5x3"
    ]
}
```

but it is parsed and returns some (test) data :
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "result": [
        {
            "blockHash": "0x7a6c6683eb052de1d9debcb78905d427464566c7a305a39d10e6fd309b797845",
            "blockNumber": "0x5",
            "contractAddress": null,
            "cumulativeGasUsed": "0x538d",
            "effectiveGasPrice": "0x1ec7cd47",
            "from": "0x658bdf435d810c91414ec09147daa6db62406379",
            "gasUsed": "0x538d",
            "logs": [],
            "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
            "status": "0x0",
            "to": "0x0000000000000000000000000000000000031ec7",
            "transactionHash": "0x54b25c11650dca0253ef7b91b5415680eea8dac54b029863e12db48908ad386c",
            "transactionIndex": "0x0",
            "type": "0x2"
        }
    ]
}
```

expected result should be similar to:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "message": "Invalid params",
        "data": "hex string without 0x prefix at line 1 column 5"
    }
}
```

PR #1739, opened 2024-01-05
rpcdaemon: add request validation

JacekGlen:
related to: #1733 invalid request passes validation and returns data

This PR introduces formal validation of RPC requests. Rather than trying to validate parameters in individual handlers, this change uses the centralized endpoints specification from https://github.com/ethereum/execution-apis.

I could not find an existing library that would do it. The closest thing I found was https://github.com/cinemast/libjson-rpc-cpp, but I don't think it meets our requirements (lightweight validation, low footprint, fast). So instead I wrote our own `JsonRpcValidator`. As this is compatible with Open API JSON RPC format, we can repack and open-source it at some point.

Notes:
- The generated specification in JSON format is included in the code base and then transformed to cpp using `genesistool`. Alternatively, this could be loaded from the JSON file to lower the build footprint
- RPC methods not included in the API Specification (e.g. Erigon API, OTS API, etc) are ignored by default. We can create the spec for them if we decide to do so later
- The validator is not yet optimized. The performance improvements will come in the next PR


silkworm/rpc/http/json_rpc_validator.cpp
```diff
@@ -0,0 +1,337 @@
...
+#include <vector>
+
+#include "json_rpc_specification.hpp"
+
+namespace silkworm::rpc::http {
+
+#define REQUEST_FIELD_METHOD "method"
+#define REQUEST_FIELD_ID "id"
+#define REQUEST_FIELD_PARAMETERS "params"
+#define REQUEST_FIELD_JSONRPC "jsonrpc"
```
canepat, line 37: Please use `constexpr` or `const` variables for constants instead of `#define`, as specified in [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Constant_Names). You can do it in next PR.

silkworm/rpc/http/json_rpc_validator.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+static const std::string valid_jsonrpc_version = "2.0";
```
canepat, line 39: Please use `constexpr` or `const` variables for constants with format specified in [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Constant_Names). You can do it in next PR.

silkworm/rpc/http/json_rpc_validator.cpp
```diff
@@ -0,0 +1,337 @@
...
+}
+
+JsonRpcValidator::JsonRpcValidator(nlohmann::json& spec_) {
+    json_spec = spec_;
+    accept_unknown_methods = true;
+}
+
+JsonRpcValidator::~JsonRpcValidator() {
+    // Destructor implementation goes here
+}
```
canepat, line 53: You can use `default` destructor in class declaration. Again next PR

canepat (approved):

PR #1744, opened 2024-01-08
rpcdaemon: separate HTTP connection from JSONRPC request handler

lupin012:


silkworm/rpc/http/channel_writer.hpp
```diff
@@ -0,0 +1,56 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#pragma once
+
+namespace silkworm::rpc::http {
```
canepat, line 19: If I understand correctly the purpose of defining this new interface `ChannelWriter`, it should be placed in `silkworm::rpc` namespace (it is not meant to be HTTP-specific, in fact the opposite is true).

silkworm/rpc/http/channel_writer.hpp
```diff
@@ -0,0 +1,56 @@
...
+        unauthorized,
+        forbidden,
+        not_found,
+        internal_server_error,
+        not_implemented,
+        bad_gateway,
+        service_unavailable
+    };
+
+    struct MessageResponse {
```
canepat, line 43: I suggest renaming `MessageResponse` as `Response`. Moreover, we can evaluate if this can be moved outside of the enclosing class but let's come back to this when this refactoring is finished.

canepat (approved):

PR #1746, opened 2024-01-09
core: amend incarnation logic

canepat:


silkworm/core/state/intra_block_state.cpp
```diff
@@ -85,7 +85,9 @@ void IntraBlockState::create_contract(const evmc::address& address) noexcept {
         created.initial = prev->initial;
         if (prev->current) {
             created.current->balance = prev->current->balance;
-            prev_incarnation = prev->current->incarnation;
+            if (prev->initial) {
+                prev_incarnation = std::max(prev->current->incarnation, prev->initial->incarnation);
+            }
```
yperbasis, line 90: Shall we add 
```
} else {
    prev_incarnation = prev->current->incarnation;
}
```
?

yperbasis:
It's worth changing the logic in `Buffer::update_account` so that it updates incarnations not only if `account_deleted`, but also when `current->incarnation == 0 && initial->incarnation > 0`, similar to https://github.com/ledgerwatch/erigon/pull/8954.

yperbasis (approved):

PR #1756, opened 2024-01-13
CI: fix fuzzer build and add it to integration workflow

canepat:
PR #1744 recently broke the fuzzer build, but it became evident only during weekly `fuzzer` workflow because the fuzzer build is not executed in any other workflow.

This PR splits the `linux-clang-fuzzer` job into separate build and run steps, including the former also in the `integration` workflow and activating the latter only in the `fuzzer` workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,6 +121,28 @@ commands:
           paths:
             - ~/.conan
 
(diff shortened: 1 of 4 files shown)


JacekGlen (approved):
Good stuff

PR #1763, opened 2024-01-18
node: modules refactoring

battlmonstr:
snapshots: move snapshots modules

* rename snapshot -> snapshots, recsplit -> rec_split, support -> common
* move rec_split, huffman, bittorrent under snapshots
* make sure that namespaces match the directory structure

db: move db modules

* move cbor types to db
* move etl to db

node: move test utils

* files.hpp -> infra/test_util/temporary_file.hpp
* snapshots.hpp -> snapshots/test_util/common.hpp
* encode_varint: don't use test_util in index.cpp
* xoroshiro128pp.hpp -> snapshots/rec_split/test_util/
* mock_execution_client.hpp -> sync/test_util

node: refactor test db context

* rename test::Context to TempChainData
* move it to db/test_util
* extract NodeSettings adapter to node/test_util
* make db::PruneMode plain copyable value type eliminating unique_ptr


JacekGlen (approved):
Oh quite a change, but I think I went through all of it. I don't necessarily agree with the plural term `snapshots` but that's just my preference.

Good job!

battlmonstr:
thanks @JacekGlen !

PR #1767, opened 2024-01-20
db: move access_layer to store

battlmonstr:


battlmonstr:
decided to not go this route

PR #1770, opened 2024-01-22
infra: migrate `ensure` to the dynamic format for performance reasons

JacekGlen:
Both compilers, gcc and clang, do not efficiently optimize string arguments for inline methods. As the result, when the following function:
```c++
inline void ensure(bool condition, const std::string& message) {
    if (!condition) [[unlikely]] {
        throw std::logic_error(message);
    }
}
```

is called with the code:
```c++
ensure(envelope_result.has_value(),
           "TransactionSnapshot: cannot decode tx envelope: " + to_hex(tx_envelope) + " error: " + to_string(envelope_result));
```
then the string argument is built every time. This is unnecessary as the string argument is only required in the unlikely event of an unmet condition.

During the execution stage, those `ensure()` methods can be called many times per block. This results in wasted resources and additional time required for completion. The preliminary tests showed ~4.5% improvement in block execution performance after implementing this change. See the attached logs for comparison: 
[block_execution_silkworm.txt](https://github.com/erigontech/silkworm/files/14009949/block_execution_silkworm.txt)

### DIFF
--- a/cmd/capi/execute.cpp
+++ b/cmd/capi/execute.cpp
@@ -193,7 +193,7 @@ std::vector<SilkwormChainSnapshot> collect_all_snapshots(const SnapshotRepositor
                 transactions_snapshot_sequence.push_back(raw_transactions_snapshot);
             } break;
             default:
-                ensure(false, "unexpected snapshot type: " + std::string{magic_enum::enum_name(segment_file.type())});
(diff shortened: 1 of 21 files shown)


canepat:
Good catch!

Can you please add a couple of unit tests to `ensure_test.cpp` covering the new function overloads?

canepat (approved):

PR #1782, opened 2024-01-29
third-party: update erigon-snapshot to commit bec7b2c

canepat:
node: add support for 100k snapshot size
cmd: add option to skip beacon-chain segments in embed_toml


silkworm/node/snapshots/path.hpp
```diff
@@ -90,7 +92,7 @@ class SnapshotPath {
     }
 
     [[nodiscard]] bool seedable() const {
-        return block_to_ - block_from_ == segment_size_;
+        return std::ranges::find(kDefaultSegmentSizes, segment_size()) != kDefaultSegmentSizes.cend();
```
yperbasis, line 95: better use `std::ranges::contains`

silkworm/node/snapshots/path.hpp
```diff
@@ -32,7 +34,7 @@ namespace silkworm::snapshots {
 constexpr int kFileNameBlockScaleFactor{1'000};
 
 //! The segment size measured as number of blocks included in each segment
-constexpr uint64_t kDefaultSegmentSize{500'000};
+constexpr std::array kDefaultSegmentSizes{500'000u, 100'000u};
```
yperbasis, line 37: Namespace constants in headers should be `inline`. See https://abseil.io/tips/140

cmd/dev/embed_toml.cpp
```diff
@@ -82,11 +97,9 @@ int main(int argc, char* argv[]) {
             output << "#include <array>\n\n";
             output << "#include <silkworm/node/snapshots/entry.hpp>\n\n";
             output << "namespace silkworm::snapshots {\n\n";
-            output << "inline constexpr std::array<Entry, " << table.size() << "> k" << snapshot_name << "Snapshots{\n";
-            for (auto&& [key, value] : table) {
-                std::string key_str{key.begin(), key.end()};
-                std::string val_str{value.as_string()->get()};
-                output << "    Entry{\"" << key_str << "\"sv, \"" << val_str << "\"sv},\n";
+            output << "inline constexpr std::array<Entry, " << filtered_table.size() << "> k" << snapshot_name << "Snapshots{\n";
```
yperbasis, line 100: Can be simply `std::array` without the template arguments (they are auto-deducted).

silkworm/node/snapshots/path.hpp
canepat: It was my first thought as well 😄, but it will be available only [since C++23](https://en.cppreference.com/w/cpp/algorithm/ranges/contains)

yperbasis (approved):

PR #1799, opened 2024-02-02
rpcdaemon: add ws ping in case of inactivity

lupin012:


silkworm/rpc/ws/connection.hpp
```diff
@@ -35,6 +36,9 @@
 
 namespace silkworm::rpc::ws {
 
+//! The default ping interval
+constexpr std::chrono::milliseconds kDefaultPingInterval{60'000};
```
canepat, line 40: You can move this constant to `connection.cpp` so that you avoid recompilation of user code when it changes.

silkworm/rpc/ws/connection.hpp
```diff
@@ -61,13 +65,20 @@ class Connection : public Channel {
...
 
     boost::beast::websocket::stream<boost::beast::tcp_stream> ws_;
 
     //! The handler used to process the incoming request.
     http::RequestHandler request_handler_;
+
+    boost::asio::steady_timer retry_timer_;
+
+    //! The retry interval between successive ping attempts
+    static std::chrono::milliseconds ping_interval_;
```
canepat, line 81: At the moment this is not needed, you can remove it and just use the constant in `connection.cpp`.

canepat (changes requested):

canepat (approved):

PR #1809, opened 2024-02-07
snapshots seg: compressor

battlmonstr:
depends on https://github.com/erigontech/silkworm/pull/1827


silkworm/core/common/bit_count.hpp
yperbasis, line 1: You don't need this guy because C++20 has `<bit>`

canepat (approved):

PR #1811, opened 2024-02-07
rpcdaemon: move HTTP chunk handling in http::Connection

lupin012:


silkworm/rpc/json/stream.cpp
```diff
@@ -303,6 +307,7 @@ Task<void> Stream::do_async_write(ChunkPtr chunk) {
 Task<void> Stream::run() {
     uint32_t total_writes{0};
     std::size_t total_bytes_sent{0};
+    co_await open();
```
canepat, line 310: I suggest moving this call in `RequestHandler::handle_request` where `Channel::open_stream` was called before

canepat (changes requested):

silkworm/rpc/json/stream.cpp
lupin012: ok

canepat (approved):

PR #1812, opened 2024-02-08
rpcdaemon: expected result in JSON RPC validator

canepat:
This PR modernises the JSON RPC validator by using `tl::expected` to implement the validation result.

The code is shorter and a bit cleaner and we also gain ~ 8% performance wise.


JacekGlen (approved):
Looks good!

PR #1818, opened 2024-02-09
capi, infra: add TTY detection to customize log format

canepat:
This PR fixes the log format used in C API to reproduce the Erigon format exactly.


battlmonstr (approved):
🧙

PR #1827, opened 2024-02-13
snapshots seg compressor: pattern aggregator

battlmonstr:


battlmonstr:
merged as a part of https://github.com/erigontech/silkworm/pull/1809

PR #1839, opened 2024-02-15
ci: fix fuzzer build targets for macOS arm64

canepat:


cmd/test/CMakeLists.txt
```diff
@@ -39,10 +39,23 @@ if(NOT SILKWORM_CORE_ONLY)
...
+      set(FUZZER_TEST_COMPILE_SANITIZE address,undefined,fuzzer)
+      set(FUZZER_TEST_LINK_SANITIZE address,undefined)
+      set(FUZZER_DIAG_COMPILE_SANITIZE address,undefined)
+      set(FUZZER_DIAG_LINK_SANITIZE address,undefined)
+    else()
+      set(FUZZER_TEST_COMPILE_SANITIZE address,leak,undefined,fuzzer)
+      set(FUZZER_TEST_LINK_SANITIZE address,leak,undefined)
+      set(FUZZER_DIAG_COMPILE_SANITIZE address,leak,undefined)
+      set(FUZZER_DIAG_LINK_SANITIZE address,leak,undefined)
+    endif()
```
battlmonstr, line 53: I suggest to refactor this having a default value "address,undefined", and then appending "leak"/"fuzzer" to it as needed, e.g.:

```suggestion
      set(FUZZER_SANITIZE_CHECKS "address,undefined")

      # macOS on 64-bit ARM does not support option -fsanitize=leak
      if(NOT (CMAKE_SYSTEM_NAME MATCHES "Darwin" AND CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL arm64))
          set(FUZZER_SANITIZE_CHECKS ${FUZZER_SANITIZE_CHECKS},leak)
      endif()

      set(FUZZER_TEST_COMPILE_SANITIZE ${FUZZER_SANITIZE_CHECKS},fuzzer)
      set(FUZZER_TEST_LINK_SANITIZE ${FUZZER_SANITIZE_CHECKS})
      set(FUZZER_DIAG_COMPILE_SANITIZE ${FUZZER_SANITIZE_CHECKS})
      set(FUZZER_DIAG_LINK_SANITIZE ${FUZZER_SANITIZE_CHECKS})
```

battlmonstr (approved):

cmd/test/CMakeLists.txt
canepat: I thought about this possibility, but in the end I found the `if-else` clearer in terms of which options are used on what platform.

Using the suggested way you surely avoid code repetitions, but IMO at the cost of less readability: in order to figure out which compiler options are used on macOS for `rpcdaemon_fuzzer_test`, you need to put three pieces together (default value, `if` evaluation, suffix appended) instead of simply looking at one line.

Obviously, this is definitely not a hill... 😄

canepat: I thought about this possibility, but in the end I found the `if-else` clearer in terms of which options are used on what platform.

Using the suggested way you surely avoid code repetitions, but IMO at the cost of less readability: in order to figure out which compiler options are used on macOS for `rpcdaemon_fuzzer_test`, you need to put three pieces together (default value, `if` evaluation, suffix appended) instead of simply looking at one line.

Obviously, this is definitely not a hill... 😄

PR #1841, opened 2024-02-15
capi: some basic unit tests

canepat:
This PR introduces very basic unit tests for C API, we will hopefully expand them in the near future.

Moreover, this PR also refactors the `silkworm_capi` target to use the existing `silkworm_library` CMake facility and fixes one unit test for macOS in `infra` library.


battlmonstr (approved):
:+1:

PR #1848, opened 2024-02-19
rpcdaemon: detect nil character in JSON request

JacekGlen:
nlohmann::parse() does not handle well nil characters (0x0). As per the spec https://www.ietf.org/rfc/rfc4627.txt these characters are allowed in string values, but not anywhere else. But for nlohmann nil character always marks the end of the string. So there are situations where invalid tokens are ignored if they appear after nil character, e.g.:
```json
{"jsonrpc":"2.0","id":1,"method":"eth_feeHistory","params":["0x1A","0x2",[95,99]]}\0x0 not_valid
```

This PR updates the way Silkworm handles nil characters, and makes it the same as in Erigon/Geth


canepat:
I would add a configuration flag (e.g. `rpc.json.strict`) to enable/disable strict validation for nil character according to JSON spec, so that we can optionally avoid the performance penalty.

canepat (approved):

PR #1861, opened 2024-02-28
node: fix snapshot sync progress update with STOP_AT_BLOCK

canepat:
node: clean exit from execution pipeline with STOP_AT_BLOCK


chfast (approved):
Looks good. Fixes my problem.

Issue #1855, opened 2024-02-26
Assertion failure in `disc_v4/ping/ping_check.cpp`


chfast:
In a release build with assertions enabled of silkworm:

```
silkworm: /home/ipsilon/silkworm/silkworm/sentry/discovery/disc_v4/ping/ping_check.cpp:111: silkworm::Task<silkworm::sentry::discovery::disc_v4::ping::PingCheckResult> silkworm::sentry::discovery::disc_v4::ping::ping_check(silkworm::sentry::EccPublicKey, boost::asio::ip::udp::endpoint, silkworm::sentry::EnodeUrl, uint64_t, silkworm::sentry::discovery::disc_v4::ping::MessageSender&, boost::signals2::signal<void(silkworm::sentry::discovery::disc_v4::ping::PongMessage, silkworm::sentry::EccPublicKey)>&, std::optional<std::chrono::time_point<std::chrono::_V2::system_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > > >, size_t): Assertion `false' failed.
```

Location: https://github.com/erigontech/silkworm/blob/master/silkworm/sentry/discovery/disc_v4/ping/ping_check.cpp#L111

PR #1863, opened 2024-02-28
sentry: assertion in ping_check.cpp (#1855)

battlmonstr:
Make sure that we never insert the local node into the nodes database. Warn if it did happen by accident with a fix suggestion.

### DIFF
--- a/silkworm/sentry/discovery/disc_v4/discovery.cpp
+++ b/silkworm/sentry/discovery/disc_v4/discovery.cpp
@@ -172,8 +172,16 @@ class DiscoveryImpl : private MessageHandler {
     Task<void> ping_check(EccPublicKey node_id) {
         using namespace std::chrono_literals;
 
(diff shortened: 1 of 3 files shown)


chfast (commented):
Any option to have some unit tests for this?

battlmonstr:
@chfast I don't plan to cover this with unit tests. Besides, I don't think it would help with this particular issue. I still don't understand how this might have happened, because the check already exists in find_neighbors, so this is not a true fix, but some defensive programming tricks. It would be helpful if you run this and tell if it produces these log warnings for you. I recommend to delete or backup the `<datadir>/nodes` before trying again.

chfast (approved):
Looks ok to me, but I have not tested it. If you want I can report back the testing results.

battlmonstr:
@chfast yes, please report to the original issue if you come across it again - https://github.com/erigontech/silkworm/issues/1855

PR #1867, opened 2024-02-29
tests: a script to list dependencies of a module

battlmonstr:
Example:

```
% ./tests/deps/deps.py silkworm/node/snapshots
silkworm/node/snapshots module depends on:
 1. core/chain
 2. core/common
 3. core/types
 4. infra/common
 5. infra/concurrency
 6. infra/test_util
 7. node/db
 8. node/db/etl
 9. node/db/test_util
10. node/snapshots
11. node/snapshots/bittorrent
12. node/snapshots/config
13. node/snapshots/rec_split
14. node/snapshots/rec_split/common
15. node/snapshots/rec_split/encoding
16. node/snapshots/seg
17. node/snapshots/seg/common
18. node/snapshots/test_util
```

### DIFF
--- a/tools/deps.py
+++ b/tools/deps.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+
+import os.path
+import glob
+import argparse
+import re
+import itertools
+
+def read_lines(path, limit):
+	with open(path, 'r') as f:
+		for line in f:
+			if limit == 0:
+				break
+			limit -= 1
+			yield line
+
+def include_lines(path):
+	return [line for line in read_lines(path, 100) if line.startswith("#include <silkworm/")]
+
+def parse_include_line_path(line):
+	return re.search('<([^>]+)>', line).group(1)
+
+def include_paths(path):
+	return [parse_include_line_path(line) for line in include_lines(path)]
+
+def module_of_path(path):
+	return os.path.dirname(path)[len("silkworm/"):]
+
+def parse_args():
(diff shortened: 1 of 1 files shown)


canepat (changes requested):
I like the idea and usefulness, please move it directly to the project folder for better usability (moreover, it seems not related to tests)

chfast:
If these modules are CMake libs, CMake can generate a graph. https://cmake.org/cmake/help/latest/module/CMakeGraphVizOptions.html

battlmonstr:
@canepat maybe we have a special directory for such things? Like `tools` or `scripts`? There are some scripts in `cmake` folder which could also be moved there.

@chfast nice. I have a dream to use this tool and print out a wallpaper :) We are moving to a more modular approach, but we still have some monoliths. This script helps to understand what needs to be refactored to extract modules from monoliths. For example `node/snapshots` needs to be untied from `node/db` to become a separate library.

JacekGlen:
> > @canepat maybe we have a special directory for such things? Like `tools` or `scripts`? There are some scripts in `cmake` folder which could also be moved there.
> 
> @battlmonstr we don't but we should have a dedicated directory for such things: you can create a `scripts` directory and move there `deps.py` and other scripts.

Actually, can we make it a `tools` directory? I would also like to move `tests/docker/` there, as that is not strictly testing related.

canepat (approved):

PR #1873, opened 2024-03-01
capi: correct signal handling

JacekGlen:
Silkworm CAPI has two threads:
- main: for block execution and writing state changes 
- block provider: for pre-loading blocks into the buffer

This PR makes sure that when the termination signal (Ctrl+C) is issued then:
- all open db transactions are closed 
- threads are joined
- operations can be resumed (db integrity preserved)

Test notes: I've added some unit tests, but the true tests were conducted manually. I ran Erigon with Silkworm and then during the Execution stage pressed Ctrl+C to terminate. Then I restarted Erigon to make sure that the execution was resumed at the point of the last state write. Then I repeated the process multiple times.


JacekGlen:
This will be redone on master to avoid difficult conflicts.
Closing...

PR #1883, opened 2024-03-06
workflow rpctest: run single rpc-test with "--transport_type both" to make http and ws tests

lupin012:


canepat:
In next PR it's worth changing the `transport_type` parameter in `rpc-tests` command to become a list of tags like:
```
--transport=http
--transport=http,websockets
```
and in the future
```
--transport=http,websockets,ipc
```

PR #1885, opened 2024-03-06
db: decouple db module

battlmonstr:
* move access_layer Account test
* refactor EthStatusDataProvider, move it to sentry
* add silkworm_db library target
* cleanup silkworm_rpcdaemon cmake script, it doesn't depends on silkworm_node now 🎉 
* cleanup some cmds to not depend on silkworm_node


silkworm/node/db/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+    silkworm_infra
+    silkworm_snapshots
+)
+# cmake-format: on
+
+silkworm_library(
+  silkworm_db
+  PUBLIC ${LIBS_PUBLIC}
+  PRIVATE ${LIBS_PRIVATE}
+  EXCLUDE_REGEX "db/etl"
```
JacekGlen, line 50:
```suggestion
  EXCLUDE_REGEX "etl"
```
not sure about it, can you double check?

silkworm/node/db/head_info.cpp
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+        log::Warning("db::HeadInfo") << "canonical hash at height " << std::to_string(head_height) << " not found in db";
+        return head_info;
+    }
+
+    auto head_total_difficulty = db::read_total_difficulty(txn, head_height, *head_hash);
+    if (head_total_difficulty) {
+        head_info.total_difficulty = head_total_difficulty.value();
+    } else {
+        log::Warning("db::HeadInfo") << "total difficulty of canonical hash at height " << std::to_string(head_height) << " not found in db";
```
JacekGlen, line 45: We should log the warning only in cases when total difficulty is required

JacekGlen (approved):
Great change!
I've got some minor comments, but you can ignore if you think they are not important

canepat:
> * cleanup silkworm_rpcdaemon cmake script, it doesn't depends on silkworm_node now 🎉

👏

silkworm/node/db/CMakeLists.txt
battlmonstr: EXCLUDE_REGEX is a temporary hack to exclude sub-libraries. The plan is to detect them automatically and exclude their subfolders from the sources list. It's on my TODO list ✍️

silkworm/node/db/head_info.cpp
battlmonstr: I agree. This function should probably be refactored to not log, but return optional values in these fields. Then we can check them and log on the call site.

PR #1901, opened 2024-03-10
Ci/fix in exclude regx mgt

lupin012:


lupin012:
open branch of branch

PR #1909, opened 2024-03-14
rpcdaemon: add HTTP gzip compression (no streaming)

lupin012:


silkworm/rpc/http/connection.cpp
```diff
@@ -190,10 +192,11 @@ Task<void> Connection::handle_actual_request(const boost::beast::http::request<b
     vary_ = req[boost::beast::http::field::vary];
     origin_ = req[boost::beast::http::field::origin];
     method_ = req.method();
+    auto encoding = req[boost::beast::http::field::accept_encoding];
 
     auto rsp_content = co_await request_handler_.handle(req.body());
     if (rsp_content) {
```
canepat, line 198: The definition of local variable `encoding` should be moved inside this code branch for better readability

silkworm/rpc/http/connection.cpp
```diff
@@ -190,10 +192,11 @@ Task<void> Connection::handle_actual_request(const boost::beast::http::request<b
     vary_ = req[boost::beast::http::field::vary];
     origin_ = req[boost::beast::http::field::origin];
     method_ = req.method();
+    auto encoding = req[boost::beast::http::field::accept_encoding];
```
canepat, line 195: `encoding` should be declared `const` (next PR is OK for this and all following suggestions)

silkworm/rpc/http/connection.cpp
```diff
@@ -190,10 +192,11 @@ Task<void> Connection::handle_actual_request(const boost::beast::http::request<b
...
 
     auto rsp_content = co_await request_handler_.handle(req.body());
     if (rsp_content) {
-        co_await do_write(rsp_content->append("\n"), boost::beast::http::status::ok);
+        co_await do_write(rsp_content->append("\n"), boost::beast::http::status::ok, !encoding.empty());
```
canepat, line 199: The value of the `Accept-Encoding` field should be passed to the `do_write` method instead of a `bool`. This will allow for checking within `do_write` which kind of content compression has been requested and accept/reject it depending on the supported compression on our side

silkworm/rpc/http/connection.cpp
```diff
@@ -265,8 +268,22 @@ Task<void> Connection::do_write(const std::string& content, boost::beast::http::
...
-        res.body() = content;
+        if (compress) {
+            const std::string compression_type = "gzip";
+            res.set(boost::beast::http::field::content_encoding, compression_type);
+            std::string compressed_data;
+            try {
+                compress_data(content, compressed_data);
+            } catch (const std::exception& e) {
+                SILK_ERROR << "Connection::compress_data exception: " << e.what();
+                throw;
```
canepat, line 279: Not sure we do really want to raise an exception here, probably we could directly return an `Internal server error` code or something similar

silkworm/rpc/http/connection.hpp
```diff
@@ -114,6 +116,8 @@ class Connection : public StreamWriter {
     std::string vary_;
     std::string origin_;
     boost::beast::http::verb method_{boost::beast::http::verb::unknown};
+
+    char temp_compressed_buffer_[10 * 1024 * 1024];
```
canepat, line 120: We should use `std::string` or `std::vector<char>` here

canepat (approved):
You can address suggestions in next PR

PR #1914, opened 2024-03-17
rpcdaemon: negotiation and validation of HTTP compression scheme

lupin012:


silkworm/rpc/http/connection.cpp
```diff
@@ -310,33 +341,10 @@ Task<void> Connection::do_write(const std::string& content, boost::beast::http::
...
-            // append the block to the output string
-            compressed_data.append(temp_compressed_buffer_, strm.total_out - compressed_data.size());
-        }
-    } while (ret != Z_STREAM_END);
-
-    deflateEnd(&strm);
+    boost::iostreams::filtering_ostream out;
+    out.push(boost::iostreams::gzip_compressor());
+    out.push(boost::iostreams::back_inserter(compressed_data));
+    boost::iostreams::copy(boost::make_iterator_range(clear_data), out);
```
canepat, line 347: You can probably improve here by keeping the `boost::iostreams::filtering_ostream` instance and the compressed data buffer as state variables in `http::Connection` in order to have gzip initialisation/finalisation happen only once per connection (as opposed to doing it for every data compression). Next PR

canepat (approved):

PR #1917, opened 2024-03-18
Capi main loop rework txn

JacekGlen:
This PR splits the existing `silkworm_execute_blocks` endpoint in CAPI and creates two separate endpoints:
- `silkworm_execute_blocks_ephemeral`: takes a db txn as a parameter and uses it to flush the data, but does not commit the transaction
- `silkworm_execute_blocks_perpetual`: creates its own db txn, commits the data and leaves the database in a coherent state

The change cleans up the code and structures it in a more logical way. It also lays the ground for future improvements.

Associated PRs:
- https://github.com/erigontech/silkworm-go/pull/4


silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
+        : write_receipts_{write_receipts},
+          write_call_traces_{write_call_traces},
+          write_change_sets_{write_change_sets},
+          analysis_cache_{kAnalysisCacheSize},
+          state_pool_{},
+          progress_{.start_time = std::chrono::steady_clock::now()},
+          log_time_{progress_.start_time + 20s},
+          max_batch_size_{max_batch_size} {
+        const auto chain_info = kKnownChainConfigs.find(chain_id);
+        chain_config_ = *chain_info;
```
canepat, line 492: We're missing the check on unknown chain id here and it's an unchecked input for both our external endpoints. I suggest to change `BlockExecutor` to receive the chain configuration as `const ChainConfig* chain_config` and restore the check:
```c++
const auto chain_info = kKnownChainConfigs.find(chain_id);
if (!chain_info) {
    return SILKWORM_UNKNOWN_CHAIN_ID;
}
const ChainConfig* chain_config{*chain_info};
```
both in `silkworm_execute_blocks_ephemeral` and `silkworm_execute_blocks_perpetual`.

You could also add a new unit test covering this negative input case.

silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
+        if (chain_config_) {
+            protocol_rule_set_ = protocol::rule_set_factory(*chain_config_);
+        }
+    }
+
+    bool has_known_chain_id() const {
+        return chain_config_ != nullptr && protocol_rule_set_ != nullptr;
```
canepat, line 499: This condition can be simplified to `protocol_rule_set_ != nullptr` with the change suggested above and maybe also the name should become `has_protocol_rule_set`

silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
+
+        db::Buffer state_buffer{txn, /*prune_history_threshold=*/0};
+        BoundedBuffer<std::optional<Block>> block_buffer{kMaxBlockBufferSize};
+        BlockProvider block_provider{&block_buffer, unmanaged_env, start_block, max_block};
+        boost::strict_scoped_thread<boost::interrupt_and_join_if_joinable> block_provider_thread(block_provider);
+
+        const size_t max_batch_size{batch_size};
+        auto signal_check_time{std::chrono::steady_clock::now()};
+
+        std::optional<Block> block;
```
canepat, line 693: You can move line 693 just before the `while` block at line 701 just for better readability.

silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
-                    log::Info("[4/12 Execution] Commit state+history",  // NOLINT(*-unused-raii)
-                              log_args_for_exec_commit(sw.since_start(elapsed), db_path));
+        if (!block_executor.has_known_chain_id()) {
+            return SILKWORM_UNKNOWN_CHAIN_ID;
+        }
+
+        while (block_number <= max_block) {
+            while (state_buffer.current_batch_state_size() < max_batch_size && block_number <= max_block) {
+                block_buffer.pop_back(&block);
+                if (!block || !block.has_value()) {
```
canepat, line 704: It seems a redundant double-check testing exactly the same condition (we already had it but it's redundant)

silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
-        // Always save the Execution stage progess when last state batch is flushed
-        db::stages::write_stage_progress(*txn, db::stages::kExecutionKey, max_block);
-        // Commit only in case of internally managed transaction
-        if (!use_external_txn) {
             StopWatch sw{/*auto_start=*/true};
-            txn->commit_and_stop();
+            log::Info{"[4/12 Execution] Flushing state",  // NOLINT(*-unused-raii)
+                      log_args_for_exec_flush(state_buffer, max_batch_size, block->header.number)};
+            state_buffer.write_state_to_db();
+            // Always save the Execution stage progess when state batch is flushed
```
canepat, line 728: small typo: `progess` -> `progress`

silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
+
+                ++block_number;
+                prefetched_blocks.pop_front();
             }
 
+            auto last_block_number = block_number - 1;
+            log::Info{"[4/12 Execution] Flushing state",  // NOLINT(*-unused-raii)
+                      log_args_for_exec_flush(state_buffer, max_batch_size, last_block_number)};
+            state_buffer.write_state_to_db();
+            // Always save the Execution stage progess when state batch is flushed
```
canepat, line 639: small typo: `progess` -> `progress`

silkworm/capi/silkworm.cpp
```diff
@@ -473,171 +477,267 @@ class BlockProvider {
...
         }
 
-        std::optional<Block> block;
-        db::DataModel da_layer{*txn};
-        Block b;
-        for (BlockNum block_number{start_block}; block_number <= max_block; ++block_number) {
-            if (use_external_txn) {
-                if (const bool ok{da_layer.read_block(block_number, /*read_senders=*/true, b)}; ok) {
-                    block = std::move(b);
+        // for (BlockNum block_number{start_block}; block_number <= max_block; ++block_number) {
```
canepat, line 605: You can remove this commented line

canepat (changes requested):

silkworm/capi/silkworm.cpp
JacekGlen: The check already exists in line 601:
```c++
        if (!block_executor.has_known_chain_id()) {
            return SILKWORM_UNKNOWN_CHAIN_ID;
        }
```

I've added unit tests to cover that scenario

silkworm/capi/silkworm.cpp
JacekGlen: no longer needed, removed

canepat (approved):

PR #1921, opened 2024-03-19
tests: pause/resume Erigon, save test results

mriccobene:


.github/workflows/rpc-performance-tests.yml
```diff
@@ -40,19 +41,24 @@ jobs:
...
+
       - name: Run Silkworm RpcDaemon
         working-directory: ${{runner.workspace}}/silkworm/build/cmd
         run: |
           ./rpcdaemon --datadir $ERIGON_DATA_DIR --api admin,debug,eth,parity,erigon,trace,web3,txpool,ots,net --log.verbosity 1 --erigon_compatibility --jwt ./jwt.hex --skip_protocol_check &
           RPC_DAEMON_PID=$!
           echo "RPC_DAEMON_PID=$RPC_DAEMON_PID" >> $GITHUB_ENV
 
-      - name: Run RPC Performances Tests
+      - name: Run RPC Integration Tests
```
lupin012, line 55: why integration not performance

.github/workflows/rpc-performance-tests.yml
```diff
@@ -40,19 +41,24 @@ jobs:
...
+      - name: Pause the Erigon instance dedicated to db maintenance
+        run: |
+          python3 $ERIGON_QA_PATH/test_system/db-producer/pause_production.py || true
+
       - name: Run Silkworm RpcDaemon
         working-directory: ${{runner.workspace}}/silkworm/build/cmd
         run: |
           ./rpcdaemon --datadir $ERIGON_DATA_DIR --api admin,debug,eth,parity,erigon,trace,web3,txpool,ots,net --log.verbosity 1 --erigon_compatibility --jwt ./jwt.hex --skip_protocol_check &
           RPC_DAEMON_PID=$!
           echo "RPC_DAEMON_PID=$RPC_DAEMON_PID" >> $GITHUB_ENV
```
lupin012, line 54: Here is it possible run also rpcdaemon (erigon) ?
What's the corresponding command "${{runner.workspace}}/silkworm/build/cmd"  for rpcdaemon ?
Obviosly silkrpc/rpcdaemon must be works on different port.

canepat: It will be possible soon but currently it is not because standalone Erigon RpcDaemon is not built yet during the provisioning stage of the self-hosted runner machine. @mriccobene will let us know when this is available, then we can add `ERIGON_DIR: /opt/erigon` definition into the environment and the activation command is:
```
- name: Run Erigon RpcDaemon
        working-directory: $ERIGON_DIR/build/bin
        run: |
          ./rpcdaemon --datadir $ERIGON_DATA_DIR --api admin,debug,eth,parity,erigon,trace,web3,txpool,ots,net --jwt ./jwt.hex &
          ERIGON_RPC_DAEMON_PID=$!
          echo "ERIGON_RPC_DAEMON_PID=$ERIGON_RPC_DAEMON_PID" >> $GITHUB_ENV
```
probably better in a separate workflow step. We will also need to have the two daemons run onto separate ports as you pointed out and finally stop Erigon RpcDaemon as well when done.

canepat (approved):

PR #1929, opened 2024-03-21
ci: macOS job - use Xcode 15

battlmonstr:


canepat (changes requested):
It's OK for me to bump the Xcode version in CI job, but I would pair this change with a proper fix for the build regression introduced by #1926 surrounding `add_link_options` with a check on compiler version:

```cmake
  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 15)
    # https://stackoverflow.com/questions/77164140/
    add_link_options(-Wl,-no_warn_duplicate_libraries)
  endif()
```

battlmonstr:
@canepat 👌 implemented here: https://github.com/erigontech/silkworm/pull/1928

> It's OK for me to bump the Xcode version in CI job, but I would pair this change with a proper fix for the build regression introduced by #1926 surrounding `add_link_options` with a check on compiler version:
> 
> ```cmake
>   if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 15)
>     # https://stackoverflow.com/questions/77164140/
>     add_link_options(-Wl,-no_warn_duplicate_libraries)
>   endif()
> ```

canepat (approved):

PR #1931, opened 2024-03-22
rpc, sync: support Engine API V3

canepat:
This PR introduces in RPC Daemon and Sync components the necessary changes to support [Engine API V3](https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#engine-api----cancun) as per Cancun hard-fork.

Moreover, a couple of small changes are also included:
- PoS sync keeps track of the number of received payload validation requests and early returns `SYNCING` as `payloadStatus` if any chain validation is already in progress
- some hard-coded error codes in RPC request handler have been replaced with `rpc::ErrorCode` constants

Depends on #1924


silkworm/rpc/json/execution_payload.cpp
```diff
@@ -66,14 +66,23 @@ void from_json(const nlohmann::json& json, ExecutionPayload& execution_payload)
...
         }
     }
-    // Optionally parse withdrawals
+    // Optionally parse V2 fields
     std::optional<std::vector<Withdrawal>> withdrawals;
     if (json.contains("withdrawals")) {
         withdrawals = json.at("withdrawals").get<std::vector<Withdrawal>>();
     }
 
+    // Optional parse V3 fields
```
JacekGlen, line 75: Need to add tests for V3 fields in `execution_payload_test.cpp`

silkworm/rpc/json/execution_payload.cpp
```diff
@@ -88,7 +97,20 @@ void from_json(const nlohmann::json& json, ExecutionPayload& execution_payload)
...
-        .withdrawals = withdrawals};
+        .withdrawals = std::move(withdrawals),
+        .blob_gas_used = blob_gas_used,
+        .excess_blob_gas = excess_blob_gas,
+    };
+
+    // Set the ExecutionPayload version (default is V1)
+    SILKWORM_ASSERT(execution_payload.version == ExecutionPayload::V1);
+    if (execution_payload.withdrawals) {
+        if (execution_payload.blob_gas_used && execution_payload.excess_blob_gas) {
```
JacekGlen, line 108: Both fields are optional so probably the condition is more like
```suggestion
        if (execution_payload.blob_gas_used || execution_payload.excess_blob_gas) {
```

silkworm/sync/sync_pos.hpp
```diff
@@ -43,16 +43,20 @@ class PoSSync : public ChainSync {
...
     Task<rpc::ExecutionPayloadBodies> get_payload_bodies_by_range(BlockNum start, uint64_t count, std::chrono::milliseconds timeout);
 
   private:
     static std::shared_ptr<Block> make_execution_block(const rpc::ExecutionPayload& payload);
+    static tl::expected<void, std::string> validate_blob_hashes(const Block& block,
+                                                                const std::optional<std::vector<Hash>>& expected_blob_versioned_hashes);
     void do_sanity_checks(const BlockHeader& header, TotalDifficulty parent_td);
     std::tuple<bool, Hash> has_valid_ancestor(const Hash& block_hash);
+
+    size_t active_chain_validations_{0};
```
JacekGlen, line 59: Is it meant to be thread-safe? We could change it to `std::atomic` if so. Or implement some sort of lock/semaphore.

JacekGlen (approved):
Looks good!

Some minor comments, but can be ignored

silkworm/sync/sync_pos.hpp
canepat: No, `PoSSync` is not meant to be thread-safe by design because it is always called by one thread: the [unique `boost::asio::io_context`](https://github.com/erigontech/silkworm/blob/master/silkworm/sync/sync.cpp#L44) running the internal RPC daemon dedicated to the Engine API end-point.

Such endpoint has always one client only (i.e. the external CL client), so it seems fair to keep the design as simple as possible: no multi-threading, just one `boost::asio::io_context` doing I/O and dispatching the incoming requests to the execution server.

PR #1936, opened 2024-03-25
rpcdaemon: compression on separate thread

lupin012:


silkworm/rpc/http/connection.hpp
```diff
@@ -91,7 +92,7 @@ class Connection : public StreamWriter {
 
     static std::string get_date_time();
 
-    Task<void> compress(boost::asio::thread_pool& workers, const std::string& clear_data, std::string& compressed_data);
+    Task<bool> compress(boost::asio::thread_pool& workers, const std::string& clear_data, std::string& compressed_data);
```
canepat, line 95: The first parameter `workers` is unnecessary: you can use directly member variable `workers_` inside `compress` implementation

canepat (changes requested):

silkworm/rpc/http/connection.cpp
```diff
@@ -290,12 +294,12 @@ Task<void> Connection::do_write(const std::string& content, boost::beast::http::
...
             std::string compressed_content;
-            try {
-                compress(content, compressed_content);
-            } catch (const std::exception& e) {
-                SILK_ERROR << "Connection::do_write cannot compress exception: " << e.what();
-                throw;
+
+            auto success = co_await compress(workers_, content, compressed_content);
+            if (!success) {
+                throw std::runtime_error("gzip compression failed");
```
canepat, line 300: Better something like:
```c++
const auto error{co_await compress(content, compressed_content)};
if (error) {
    throw boost::system::system_error{error};
}
```
where `compress` returns `boost::asio::error_code`

silkworm/rpc/http/connection.cpp
```diff
@@ -428,4 +425,31 @@ std::string Connection::get_date_time() {
     return ss.str();
 }
 
+Task<bool> Connection::compress(
```
canepat, line 428: This should return `boost::asio::error_code` (see suggestion above)

silkworm/rpc/http/connection.cpp
```diff
@@ -428,4 +425,31 @@ std::string Connection::get_date_time() {
...
+    boost::asio::thread_pool& workers,
+    const std::string& clear_data,
+    std::string& compressed_data) {
+    auto this_executor = co_await boost::asio::this_coro::executor;
+    boost::iostreams::filtering_ostream out;
+    const auto ret_success = co_await boost::asio::async_compose<decltype(boost::asio::use_awaitable), void(bool)>(
```
canepat, line 434: This should return `boost::asio::error_code` (see suggestion above)

silkworm/rpc/http/connection.cpp
```diff
@@ -428,4 +425,31 @@ std::string Connection::get_date_time() {
...
+        [&](auto& self) {
+            boost::asio::post(workers, [&, self = std::move(self)]() mutable {
+                bool success = true;
+                try {
+                    out.push(boost::iostreams::gzip_compressor());
+                    out.push(boost::iostreams::back_inserter(compressed_data));
+                    boost::iostreams::copy(boost::make_iterator_range(clear_data), out);
+                } catch (const std::exception& e) {
```
canepat, line 442: This should catch also `boost::system::system_error` and extract error code to return back to caller using `self.complete`

silkworm/rpc/http/connection.hpp
```diff
@@ -91,7 +92,7 @@ class Connection : public StreamWriter {
 
     static std::string get_date_time();
 
-    void compress(const std::string& clear_data, std::string& compressed_data);
+    Task<bool> compress(boost::asio::thread_pool& workers, const std::string& clear_data, std::string& compressed_data);
```
canepat, line 95: The first parameter `workers` is unnecessary: you can use directly member variable `workers_` inside `compress` implementation

canepat (changes requested):

canepat (approved):

PR #1943, opened 2024-03-27
infra: setup signal handling with SA_ONSTACK flag

### LINKED ISSUE #1940: C API execute blocks: non-Go code set up signal handler without SA_ONSTACK flag
If you launch Silkworm C API block execution on Linux from Go code and then hit CTRL-C, you may sometimes encounter the following unhandled exception:
```
signal 2 received but handler not on signal stack
fatal error: non-Go code set up signal handler without SA_ONSTACK flag
```
This is because the Go runtime expects its signal handlers to be called on their dedicated signal stack, as stated in Go [documentation](https://pkg.go.dev/os/signal#hdr-Go_programs_that_use_cgo_or_SWIG):

> If the non-Go code installs any signal handlers, it must use the SA_ONSTACK flag with sigaction. Failing to do so is likely to cause the program to crash if the signal is received. Go programs routinely run with a limited stack, and therefore set up an alternate signal stack.

canepat:
This PR changes the implementation of `SignalHandler` to relying on `sigaction` system call on Linux and macOS platforms, so that we can use `SA_ONSTACK` flag on registered signal handlers.

Fixes #1940

### DIFF
--- a/silkworm/infra/concurrency/signal_handler.cpp
+++ b/silkworm/infra/concurrency/signal_handler.cpp
@@ -106,12 +106,30 @@ std::atomic_int SignalHandler::sig_code_{0};
 std::atomic_bool SignalHandler::signalled_{false};
 std::function<void(int)> SignalHandler::custom_handler_;
 bool SignalHandler::silent_{false};
-std::map<int, void (*)(int)> previous_signal_handlers;
+using SignalHandlerFunc = void (*)(int);
+std::map<int, SignalHandlerFunc> previous_signal_handlers;
+
+static SignalHandlerFunc register_signal_action(const int signal_number, SignalHandlerFunc handler_func) {
+#ifdef _WIN32
+    return signal(signal_number, handler_func);
+#else
+    struct sigaction sa{};
+    sa.sa_handler = handler_func;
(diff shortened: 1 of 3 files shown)


silkworm/infra/concurrency/signal_handler.cpp
```diff
@@ -106,12 +106,31 @@ std::atomic_int SignalHandler::sig_code_{0};
 std::atomic_bool SignalHandler::signalled_{false};
 std::function<void(int)> SignalHandler::custom_handler_;
 bool SignalHandler::silent_{false};
-std::map<int, void (*)(int)> previous_signal_handlers;
+
+using SignalHandlerFunc = void (*)(int);
+std::map<int, SignalHandlerFunc> previous_signal_handlers;
+
+static SignalHandlerFunc register_signal_action(const int signal_number, SignalHandlerFunc handler_func) {
```
JacekGlen, line 113:
```suggestion
static SignalHandlerFunc register_signal_action(const int sig_code, SignalHandlerFunc handler_func) {
```
so it matches naming in the other parts of the class

JacekGlen (approved):
Tiny change only. Looks good!

PR #1949, opened 2024-03-29
core, rpcdaemon: add check on NoPreMergeConfig and terminal total difficulty

lupin012:


canepat:
Let's try to avoid this kind of defensive programming and see if we can instead identify where and why pre-merge rule set can be null in the first place

canepat:
> Let's try to avoid this kind of defensive programming and see if we can instead identify where and why pre-merge rule set can be null in the first place

In order to trigger a `nullptr` dereference in `MergeRuleSet` the following conditions must be true:

1. `ChainConfig` instance has `NoPreMergeConfig` rule set config as a result of parsing any JSON chain config containing an [unknown rule set](https://github.com/erigontech/silkworm/blob/master/silkworm/core/chain/config.cpp#L122)
2. the same `ChainConfig` instance has non-null `terminal_total_difficulty` as a result of parsing the same JSON

As a consequence of both 1. and 2. being met, [`pre_merge_rule_set` function](https://github.com/erigontech/silkworm/blob/master/silkworm/core/protocol/rule_set.cpp#L267) returns a `nullptr`, which is then received at construction time by the `MergeRuleSet` instance created in [rule_set_factory](https://github.com/erigontech/silkworm/blob/master/silkworm/core/protocol/rule_set.cpp#L278).

However, here the point is that an implicit invariant of `ChainConfig` is violated. Theoretically, the conditions 1. and 2. cannot be true at the same time because the first one means _the chain has no pre-merge rule set because the merge happens at genesis_ and the second one means _the chain has non-zero terminal total difficulty identifying where the merge point is_ (clearly not at genesis).

Hence, a proper solution here implies making the `ChainConfig` invariant explicit by:

- enforcing the invariant when building a new `ChainConfig` instance in `ChainConfig::from_json` by returning `std::nullopt` (i.e. invalid configuration) if this is not the case
- put an assertion in [rule_set_factory](https://github.com/erigontech/silkworm/blob/master/silkworm/core/protocol/rule_set.cpp#L277) to check the invariant immediately after the pre-merge rule set creation

canepat:
@yperbasis this PR looks good to me at least for preventing null pointer dereference when `pre_merge_rule_set_` is `nullptr` in `MergeRuleSet` (e.g. for Holesky config when checked against empty block), so I'm going to merge it right now. Please review it sooner or later and suggest any improvement.

canepat (approved):

PR #1952, opened 2024-03-31
ropcdaemon: get receipt on worker thread

lupin012:


lupin012:
After performance misure these changes has less performance

PR #1968, opened 2024-04-15
capi: export RpcDaemon configuration settings

canepat:
This PR exports our Silkworm RPC Daemon configuration settings in the C API interface. The following options can be customised (all of them are opt-in, default values provide a meaningful configuration):
- number of I/O execution contexts to use in concurrency mode
- configuration options for interface log of Execution JSON-RPC end-point
- host address for Execution JSON-RPC end-point
- listening port number for Execution JSON-RPC end-point
- API namespace specification (comma-separated list of API namespaces)
- number of threads in worker pool (for long-run tasks)
- list of CORS domains
- path to the JWT file in UTF-8
- flag indicating if JSON-RPC strict compatibility w/ Erigon is supported
- flag indicating if WebSocket support is enabled
- flag indicating if compression of WebSocket messages is supported
- flag indicating if compression of HTTP messages is supported

**Extras**

This PR also adds the logging verbosity level as Silkworm instance global setting. This is necessary to make Silkworm emit logs with the same verbosity as the C API client code, if needed.

### DIFF
--- a/silkworm/capi/common.cpp
+++ b/silkworm/capi/common.cpp
@@ -18,9 +18,49 @@
 
 #include <cstring>
 
-std::filesystem::path make_path(const char data_dir_path[SILKWORM_PATH_SIZE]) {
+namespace log = silkworm::log;
+
+//! Build Silkworm log level from its C representation
+static log::Level make_log_level(const SilkwormLogLevel c_log_level) {
+    log::Level verbosity{};
+    switch (c_log_level) {
+        case SilkwormLogLevel::NONE:
+            verbosity = log::Level::kNone;
+            break;
+        case SilkwormLogLevel::CRITICAL:
+            verbosity = log::Level::kCritical;
+            break;
+        case SilkwormLogLevel::ERROR:
+            verbosity = log::Level::kError;
+            break;
+        case SilkwormLogLevel::WARNING:
+            verbosity = log::Level::kWarning;
+            break;
+        case SilkwormLogLevel::INFO:
+            verbosity = log::Level::kInfo;
+            break;
+        case SilkwormLogLevel::DEBUG:
+            verbosity = log::Level::kDebug;
+            break;
+        case SilkwormLogLevel::TRACE:
+            verbosity = log::Level::kTrace;
+            break;
+    }
+    return verbosity;
+}
+
+std::filesystem::path parse_path(const char data_dir_path[SILKWORM_PATH_SIZE]) {
     // Treat as char8_t so that filesystem::path assumes UTF-8 encoding of the input path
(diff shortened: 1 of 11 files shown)


silkworm/capi/common.cpp
```diff
@@ -0,0 +1,26 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+#include "common.hpp"
+
+#include <cstring>
+
+std::filesystem::path make_path(const char data_dir_path[SILKWORM_PATH_SIZE]) {
```
JacekGlen, line 21: The name `make` suggests that it creates path on the disk
```suggestion
std::filesystem::path cast_path(const char data_dir_path[SILKWORM_PATH_SIZE]) {
```

silkworm/capi/silkworm.h
```diff
@@ -98,7 +98,23 @@ struct SilkwormChainSnapshot {
 #define SILKWORM_PATH_SIZE 260
 #define SILKWORM_GIT_VERSION_SIZE 32
 
+//! Silkworm library logging level
+enum SilkwormLogLevel : uint8_t {
```
JacekGlen, line 102: Is it the right place for the log level? I'd move it to capi/common or even to the top level.

silkworm/capi/silkworm.cpp
```diff
@@ -44,24 +44,59 @@
...
 #include <silkworm/infra/concurrency/thread_pool.hpp>
 
+#include "common.hpp"
 #include "instance.hpp"
 
 using namespace std::chrono_literals;
 using namespace silkworm;
 
+//! Create Silkworm log level from its C representation
+static log::Level make_log_level(const SilkwormLogLevel c_log_level) {
```
JacekGlen, line 58: Same as below, can be moved to more common place

silkworm/capi/common.cpp
canepat: You're right, thank you. I'm going to rename it `parse_path` to follow the naming of other functions we already have here, e.g. [parse_cors_domains](https://github.com/erigontech/silkworm/blob/capi_rpcdaemon_settings/silkworm/capi/rpcdaemon.cpp#L60) [parse_peer_urls](https://github.com/erigontech/silkworm/blob/capi_rpcdaemon_settings/silkworm/capi/sentry.cpp#L40)

silkworm/capi/silkworm.h
canepat: I understand your doubt, but it needs to stay here because the rule is:

**C API client code must include just "silkworm.h"**

PR #1970, opened 2024-04-16
snapshots refactoring

battlmonstr:
Previously the Snapshot subclasses were responsible for everything:
* snapshot serialization formats
* related indexes management
* data access layer

Most of the logic was copy-pasted 3 times with minor variations.

Now the Snapshot logic is reorganized to decouple storage engine from the Ethereum domain objects.
The copy-paste is reduced.
The same design structure can be applied to support more Snapshot/Index types in the future.

The Repository data model is changed to contain a continuous sequence of full snapshot bundles (see (5) below).

### Changes:

1. SnapshotWordDeserializer represents the snapshot word format. Multiple deserializers can be used for the same .seg (e.g. to get words as raw RLP).

2. Snapshot represents a .seg file read-only view and provides iteration over entries with a custom SnapshotWordDeserializer.

Example to open a snapshot:

	Snapshot snapshot{path};
	snapshot.reopen_segment();

3. Index is decoupled from Snapshot. It is a wrapper of RecSplitIndex.

Example to open an index:

	Index index{path.index_file()};
	index.reopen_index();

4. Data access layer is decoupled from Snapshot and consists of snapshot readers and query objects.

A snapshot reader is a type-safe iterator over Snapshot words. A reader is bound with a particular SnapshotWordDeserializer. Readers don't depend on indexes. They assume that the snapshot was open for reading.

Example to iterate over headers:

    for (const BlockHeader& header : HeaderSnapshotReader{snapshot}) { ... }

Each query is a separate struct. Queries usually depend on indexes. In this case they assume that the index was open for reading.

Common queries are implemented in basic_queries.hpp: FindByIdQuery, FindByHashQuery, RangeFromIdQuery. They are then bound with a particular reader to produce results:

	struct HeaderFindByHashQuery : public FindByHashQuery<HeaderSnapshotReader> {}

Example to find a header by hash:

 	std::optional<BlockHeader> header_opt = HeaderFindByHashQuery{snapshot, index}.exec(hash);

Some queries have custom logic, for example: BodyTxsAmountQuery.

5. Since the indexes are decoupled, they have to be stored separately in the Repository. In order to cope with complexity I've decided to simplify Repository to a different model. Previously the repository had snapshots organized by type, then by paths, and inside each snapshot it contained related indexes. It is changed to a sequence of full bundles organized by block_from sequentially and continuously. A full bundle must have all snapshot and index types for a given block range. It means that Repository doesn't handle "partial bundles" anymore, and is more strict to not having gaps. SnapshotBundle is like a database schema:

```cpp
struct SnapshotBundle {
    Snapshot header_snapshot;
    Index idx_header_hash;
    
    Snapshot body_snapshot;
    Index idx_body_number;
    
    Snapshot txn_snapshot;
    Index idx_txn_hash;
    Index idx_txn_hash_2_block;
}
```


battlmonstr:
Making read_senders=true always has no tangible effect on performance.
This test on a 40 Gb transactions file shows it (CMAKE_BUILD_TYPE=RelWithDebInfo):

```
TEST_CASE("tt") {
    Snapshot snapshot{*SnapshotPath::parse("/Volumes/WD4K/mainnet/snapshots/v1-017500-018000-transactions.seg")};
    snapshot.reopen_segment();
    TransactionSnapshotReader reader{snapshot};
    for ([[maybe_unused]] auto& t : reader) {}
}
```

with tx.set_sender() in decode_word_into_tx():

1. 	run 1: 3:10
2. 	run 2: 3:06
3. 	run 3: 3:10

without tx.set_sender()/senders_data in decode_word_into_tx():

1. 	run 1: 3:06
2. 	run 2: 3:05
3. 	run 3: 3:11

battlmonstr:
@canepat I was able to execute pre-downloaded snapshots from scratch to the last block I have in snapshots (18.8M - 1) locally on macOS. It took about 3 days. Now it is continuing on stage 6 (`[6/12 HashState]`).

canepat:
Great work!

canepat (approved):

PR #1989, opened 2024-04-26
interfaces: update to commit b57f057

canepat:
This PR updates the internal gRPC interfaces to https://github.com/ledgerwatch/interfaces/commit/b57f05746087760985ae755dd1431a35377a2363, which is the one used in Erigon at https://github.com/ledgerwatch/erigon/commit/b9ebb6ca990d8cc36a38f6c1cf33af55a5934d6e. Specifically:

- applies the minimal required refactoring to accommodate the changes in interface spec. Given that this update contains at least one big change (i.e. removal of _internal Engine-like_ RPCs from `ethbackend`), the code changes touch modules modified to use `execution` interface: `rpcdaemon`, `sync`
- moreover, the same interface change has implication for our unit test facilities because we need to add a mock for `execution` interface, which in turn makes necessary to introduce `test_util` sub-module also in `rpcdaemon` to avoid reinserting an unnecessary dependency. This results in changes to `capi`, `db`, `infra` modules, which however are mostly just different include paths.

I've tried to limit the changes as much as possible, but the result is huge anyway.

*Extras*

Document the procedure to update gRPC interfaces into our Contribution Guide


silkworm/node/stagedsync/remote_client.cpp
```diff
@@ -29,7 +29,7 @@ namespace silkworm::execution {
 using namespace std::chrono;
 using namespace boost::asio;
 
-static void serialize_header(const BlockHeader& bh, ::execution::Header* header) {
+/*static void serialize_header(const BlockHeader& bh, ::execution::Header* header) {
```
battlmonstr, line 32: I assume that this needs extra work outside of this PR scope, therefore it is commented and not deleted?

Maybe add a TODO comment like in some other cases?

silkworm/rpc/commands/engine_api_test.cpp
```diff
@@ -198,8 +202,9 @@ TEST_CASE("handle_engine_get_payload_v1 succeeds if request is expected payload"
     ClientContextPool cp{1};
     cp.start();
     std::unique_ptr<ethdb::Database> database;
+    std::unique_ptr<ethbackend::BackEnd> backend;
     // Initialise components
-    EngineRpcApi_ForTest rpc(database.get(), backend.get());
+    EngineRpcApi_ForTest rpc(database.get(), engine.get(), backend.get());
```
battlmonstr, line 207: aren't those NULL here? (and in some other cases in this file)

consider moving those mocks to be created and owned by the EngineRpcApi_ForTest - this would save some repetition in this file

silkworm/rpc/ethbackend/remote_backend_test.cpp
```diff
@@ -245,7 +245,7 @@ TEST_CASE_METHOD(EthBackendTest, "BackEnd::node_info", "[silkworm][rpc][ethbacke
     }
 }
 
-TEST_CASE_METHOD(EthBackendTest, "BackEnd::engine_get_payload", "[silkworm][rpc][ethbackend][backend]") {
+/*TEST_CASE_METHOD(EthBackendTest, "BackEnd::engine_get_payload", "[silkworm][rpc][ethbackend][backend]") {
```
battlmonstr, line 248: add a TODO or remove

battlmonstr (approved):

silkworm/rpc/commands/engine_api_test.cpp
canepat: I'm currently doing this suggested change and even more refactoring to these tests right now

PR #1991, opened 2024-04-28
rpcdaemon: fix retry logic in eth_estimateGas

lupin012:


silkworm/rpc/commands/eth_api.cpp
```diff
@@ -910,7 +910,7 @@ Task<void> EthereumRpcApi::handle_eth_estimate_gas(const nlohmann::json& request
         reply = make_json_error(request, 100, error_msg);
         co_return;
     }
-    const auto call = params[0].get<Call>();
+    auto call = params[0].get<Call>();
```
canepat, line 913: I would prefer to keep this `const` as it is currently. I've seen why you're changing it below, but I think you can achieve the same goal with a different change. More on it in the comments below.

silkworm/rpc/core/estimate_gas_oracle.cpp
```diff
@@ -69,6 +69,12 @@ Task<intx::uint256> EstimateGasOracle::estimate_gas(const Call& call, const silk
                       << ", allowance " << allowance;
             hi = uint64_t(allowance);
         }
+    } else {
+        const auto base_fee_per_gas = block.header.base_fee_per_gas;
+        if (base_fee_per_gas) {
+            std::cout << "set gas_price to: " << intx::to_string(*base_fee_per_gas);
+            call.set_gas_price(*base_fee_per_gas);
+        }
```
canepat, line 77: Here I would just overwrite `gas_price` value with `block.header.base_fee_per_gas` (you can use `value_or` for this) and then always pass `gas_price` as overridden gas price into `call.to_transaction`.

silkworm/rpc/types/call.hpp
```diff
@@ -45,6 +45,10 @@ struct Call {
     std::optional<uint64_t> nonce;
     AccessList access_list;
 
+    void set_gas_price(const intx::uint256& new_gas_price) {
+        gas_price = new_gas_price;
+    }
+
```
canepat, line 51: I would remove this and change the `to_transaction` signature to become:
```
[[nodiscard]] silkworm::Transaction to_transaction(std::optional<intx::uint256> override_gas_price) const
```
so that you can pass a value to override the `Call` gas price just when using it as a factory for a `Transaction` object.

The same reasoning also applies to `access_list`, I suggest doing the same change for that in a separate PR.

canepat (changes requested):

silkworm/rpc/types/call.hpp
lupin012: ok

silkworm/rpc/core/estimate_gas_oracle.cpp
lupin012: ok

silkworm/rpc/commands/eth_api.cpp
lupin012: ok

canepat (approved):

PR #2004, opened 2024-05-07
execution: direct client + gRPC client and server

canepat:
This PR defines the following abstractions wrapped around the internal [`execution`](https://github.com/ledgerwatch/interfaces/blob/master/execution/execution.proto) interface:

- gRPC client and server
- direct (i.e. in-process, gRPC-free) client implementation

This PR adds the client/server bindings for such interface, but does not yet uses them. Another PR will follow to put these bindings in play refactoring the current implementation in `stagedsync` package and its usage in `sync` module.

*Extras*
Some more bits have been added or touched to accommodate the needed changes:

- convenience function `concurrency::co_spawn_and_await` to spawn a coroutine using `co_spawn_sw` and awaiting its completion asynchronously
- new type alias `TaskExecutor`
- optional `stack_size` parameter in `rpc::Server::async_run`
- support for missing `safe_block_hash` in `stagedsync::Fork` and `stagedsync::ExtendingFork`

### DIFF
--- a/silkworm/infra/concurrency/co_spawn_sw.hpp
+++ b/silkworm/infra/concurrency/co_spawn_sw.hpp
@@ -369,4 +369,12 @@ inline BOOST_ASIO_INITFN_AUTO_RESULT_TYPE(
                          std::forward<CompletionToken>(token));
 }
 
+template <typename Executor, typename F>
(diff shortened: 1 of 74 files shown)


JacekGlen (approved):
Approving with some minor comments. This will be fixed in subsequent PRs. 
Otherwise this change should not break any functionality.

PR #2013, opened 2024-05-10
node, sync: use new Execution API client/server

canepat:
This PR changes the `node` and `sync` modules to use the client/server abstractions built around the [`execution`](https://github.com/ledgerwatch/interfaces/blob/master/execution/execution.proto) interface. Specifically:

- create Execution API DirectClient and Server (disabled by config for now) in `node` module facade
- migrate usage of Execution API Client in `sync` module
- bridge Execution API DirectClient from `node` to `sync` facade in `silkworm` main

*Extras*
- extract sample block generation as common fixture in unit tests
- improve readability by adding some type alias
- handle some more error conditions in PoS sync

This PR depends on #2004 

This PR does *not* remove the old in-process-only implementation yet, another PR will follow


JacekGlen (approved):
Looks good!

PR #2019, opened 2024-05-10
rpcdaemon: fix eth_call without sender

### LINKED ISSUE #2017: rpcdaemon: `eth_call` returns error when (optional) sender parameter is absent
When trying to query a smart contract, such as like this:

```json
{
	"id": 19,
	"jsonrpc": "2.0",
	"method": "eth_call",
	"params": [
		{
			"data": "0xfeaf968c",
			"to": "0xffe5b13cb108a99edf87926faf09d5008186dd91"
		},
		"latest"
	]
}
```

Erigon replies with an error saying the sender cannot be recovered:
```json
{
	"jsonrpc": "2.0",
	"id": 19,
	"error": {
		"code": -32000,
		"message": "malformed transaction: cannot recover sender"
	}
}
```

Note that by default, if a sender isn't given it's assumed to be the zero address.

lupin012:
If sender is not specified in `Call` parameter of `eth_call`, zero address should be used.

Fixes #2017

### DIFF
--- a/silkworm/rpc/core/evm_executor.cpp
+++ b/silkworm/rpc/core/evm_executor.cpp
@@ -238,9 +238,6 @@ ExecutionResult EVMExecutor::call(
         evm.add_tracer(*tracer);
     }
 
-    if (!txn.sender()) {
-        return {std::nullopt, txn.gas_limit, Bytes{}, "malformed transaction: cannot recover sender"};
-    }
     ibs_state_.access_account(*txn.sender());
 
     const evmc_revision rev{evm.revision()};
--- a/silkworm/rpc/core/evm_executor.cpp
+++ b/silkworm/rpc/core/evm_executor.cpp
@@ -238,6 +238,9 @@ ExecutionResult EVMExecutor::call(
         evm.add_tracer(*tracer);
     }
 
+    if (!txn.sender()) {
+        return {std::nullopt, txn.gas_limit, Bytes{}, "malformed transaction: cannot recover sender"};
+    }
     ibs_state_.access_account(*txn.sender());
 
     const evmc_revision rev{evm.revision()};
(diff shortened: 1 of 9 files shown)


silkworm/rpc/core/evm_executor.cpp
```diff
@@ -238,9 +238,6 @@ ExecutionResult EVMExecutor::call(
         evm.add_tracer(*tracer);
     }
 
-    if (!txn.sender()) {
-        return {std::nullopt, txn.gas_limit, Bytes{}, "malformed transaction: cannot recover sender"};
-    }
```
canepat, line 243: I would prefer not to remove this check because at the `EVMExecutor` API level it still makes a lot of sense.

silkworm/rpc/types/call.hpp
```diff
@@ -51,6 +51,9 @@ struct Call {
         silkworm::Transaction txn{};
         if (from) {
             txn.set_sender(*from);
+        } else {
+            constexpr auto zero_address = 0x0000000000000000000000000000000000000000_address;
+            txn.set_sender(zero_address);
         }
```
canepat, line 57: Now you can and should simplify this by removing the `if-else` clause:
```
txn.set_sender(from ? *from : evmc::address{});
```

canepat (changes requested):

silkworm/rpc/types/call.hpp
lupin012: ok

silkworm/rpc/core/evm_executor.cpp
lupin012: ok

canepat (approved):

PR #2031, opened 2024-05-20
core, rpcdaemon: gas bailout in EVM

Sixtysixter:
Added gas_bailout member to EVM for call simulation

### DIFF
--- a/silkworm/core/execution/evm.cpp
+++ b/silkworm/core/execution/evm.cpp
@@ -59,11 +59,12 @@ class DelegatingTracer : public evmone::Tracer {
     IntraBlockState& intra_block_state_;
 };
 
-EVM::EVM(const Block& block, IntraBlockState& state, const ChainConfig& config) noexcept
+EVM::EVM(const Block& block, IntraBlockState& state, const ChainConfig& config, bool gas_bailout) noexcept
     : beneficiary{block.header.beneficiary},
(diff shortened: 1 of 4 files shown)


silkworm/core/execution/evm.cpp
```diff
@@ -203,7 +204,7 @@ evmc::Result EVM::call(const evmc_message& message) noexcept {
     evmc::Result res{EVMC_SUCCESS, message.gas};
 
     const auto value{intx::be::load<intx::uint256>(message.value)};
-    if (message.kind != EVMC_DELEGATECALL && state_.get_balance(message.sender) < value) {
+    if (!gas_bailout_ && message.kind != EVMC_DELEGATECALL && state_.get_balance(message.sender) < value) {
         res.status_code = EVMC_INSUFFICIENT_BALANCE;
         return res;
     }
```
canepat, line 210: I would move the condition on `gas_bailout` inside the existing condition to emphasise its role by also adding a comment.

Moreover, I suppose we need the same additional check on gas bailout also in `EVM::create`.

canepat (changes requested):

silkworm/core/execution/evm.cpp
Sixtysixter: Left in this way as discussed

PR #2053, opened 2024-05-30
rpcdaemon: fix result as error in some ots API

lupin012:


silkworm/rpc/core/evm_executor.cpp
```diff
@@ -178,6 +178,13 @@ uint64_t EVMExecutor::refund_gas(const EVM& evm, const silkworm::Transaction& tx
     return gas_left;
 }
 
+void EVMExecutor::run_transactions_n(const silkworm::Block& block, const uint64_t n_tx, const Tracers& tracers, bool refund, bool gas_bailout) {
```
canepat, line 181: I suggest renaming `run_transactions_n` -> `call_first_n` and `n_tx` -> `n`

canepat (changes requested):

canepat (approved):

PR #2056, opened 2024-05-30
snapshots: refactor SnapshotRepository

battlmonstr:
Refactor data access layer (DAL) logic and schema definition out of SnapshotRepository.

* schema is defined in snapshot_bundle.hpp and snapshot_type.hpp, indexes are defined in SnapshotBundleFactory
* find_segment is exposed with a snapshot type parameter
* view_xx_segments is replaced by iteration on bundles in a forward or reverse order (view_bundles_reverse)
* find_block_number is replaced by a new query object - TransactionBlockNumByTxnHashRepoQuery
  that runs TransactionBlockNumByTxnHashQuery on a view of bundles and returns the first match
* replace for_each_xx with nested for loops
* refactor queries to take SnapshotAndIndex

DataModel::read_transactions_from_snapshot: remove an extra reserve

### DIFF
--- a/cmd/dev/snapshots.cpp
+++ b/cmd/dev/snapshots.cpp
@@ -408,11 +408,11 @@ void lookup_header_by_hash(const SnapSettings& settings) {
     SnapshotRepository snapshot_repository{settings};
     snapshot_repository.reopen_folder();
     for (const SnapshotBundle& bundle : snapshot_repository.view_bundles_reverse()) {
-        auto snapshot = bundle.snapshot_and_index(SnapshotType::headers);
-        const auto header = HeaderFindByHashQuery{snapshot.snapshot, snapshot.index}.exec(*hash);
+        auto snapshot_and_index = bundle.snapshot_and_index(SnapshotType::headers);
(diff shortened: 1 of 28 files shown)


cmd/dev/snapshots.cpp
```diff
@@ -227,44 +228,48 @@ void decode_segment(const SnapSettings& settings, int repetitions) {
     SILK_INFO << "Decode snapshot elapsed: " << duration_as<std::chrono::milliseconds>(elapsed) << " msec";
 }
 
+static std::unique_ptr<SnapshotBundleFactory> bundle_factory() {
+    return std::make_unique<silkworm::db::SnapshotBundleFactoryImpl>();
+}
```
canepat, line 233: I suggest moving this factory function into `snapshot_bundle_factory_impl.hpp`, so that we can use it everywhere and have a one-liner creation for `SnapshotRepository` as below:
```
SnapshotRepository snapshot_repo{settings, bundle_factory()};
```

canepat (approved):

Issue #2064, opened 2024-06-03
Execution OOM crash in 16 Gb at 2-5M


battlmonstr:
Running silkworm commit https://github.com/erigontech/silkworm/commit/b520fbac6c931f81ca2933f5fdf07cb2814f3155 with pre-downloaded snapshots on a 16 Gb (about 13.5 Gb free) Debian or Ubuntu VM with options:

		--prune=htrc
		--snapshots.no_downloader # snapshots are pre-downloaded
		--sentry.remote.addr=127.0.0.1:9091 # to disable sentry

Results in a forced crash by OOM killer:

> kernel: Out of memory: Killed process 45258 (silkworm) total-vm:3843295636kB, anon-rss:15170316kB, file-rss:3068kB, shmem-rss:0kB, UID:1000 pgtables:71252kB oom_score_adj:0

This happened 5 out of 5 tries after 33 min in execution stage at around 4.5M blocks (e.g. at blocks: 4479463, 4483121, 4477025)

Lowering the execution batch size helps sometimes using this option:

		--batchsize=128MB

but it still crashes sometimes (4 out of 5 tries, at blocks: 2431423, 2433164, 2426881, 2433253).

Possible solutions (from easy to hard):
1. update [README](https://github.com/ledgerwatch/erigon?tab=readme-ov-file#system-requirements) to advice 32 Gb recommended, and 16 Gb minimal with `--batchsize=128MB` prescription.
2. update the default value dynamically based on available RAM
3. investigate what causes the crash and why (a leak? a spike?) and propose solutions
4. replace `--batchsize` with something more meaningful to the user (e.g. total max RAM for execution)
5. preallocate the required memory for execution on startup

battlmonstr:
Findings:
* `--batchsize=128MB` is not taken into account as it should in standalone silkworm. The execution stage uses a heuristic formula based on `block.header.gas_used` which results in a bad RAM estimation. The CAPI uses a different estimation method (`current_batch_state_size()`). It is mentioned in https://github.com/erigontech/silkworm/issues/2078
* Given `--batchsize=128MB` the execution stage actually eats at least 1.3 Gb (including 1 Gb by `Buffer::accounts_` and 220 Mb by `Buffer::storage_`). It then crashes because it needs even more RAM to continue execution. The crash correlates with `Buffer::accounts_` growth (`rehash_and_grow_if_necessary()`) from 1.9 Gb to 3.8 Gb.
[...]

PR #2081, opened 2024-06-06
execution: fix OOM crash due to db::Buffer::accounts realloc (#2064)

battlmonstr:
The flat_hash_map has a policy (in `rehash_and_grow_if_necessary()`) to grow 2x when its size reaches 25/32 (78%) of capacity.

Previously the batch execution logic was adding things to db::Buffer, and then testing if its size becomes bigger than batch_size. This lead to significant overshooting, because at the growth threshold it will double the memory size and make it batch_size * 2 (even more, because there's also unused 22% of capacity).

In standalone silkworm execution stage the batch_size estimation was based on a gas-based heuristic formula, which lead to significant underestimation and an OOM crash when trying to allocate 4 Gb of RAM at once.

This changes the logic from checking batch_size aposteriori to apriori. If there's a possibility to grow the buffer significantly and overshoot the limit, we shouldn't take the risk, and commit the batch.

Additional changes:
* execution stage: flush history after each block as in CAPI
* capi: persist the executed work before quitting on invalid block
* remove unused db::Buffer::current_batch_history_size()
* ExecutionProcessor: split flush_state from execute_block
* block_buffer.terminate_and_release_all using gsl::finally

### DIFF
--- a/silkworm/core/execution/processor.hpp
+++ b/silkworm/core/execution/processor.hpp
@@ -41,10 +41,13 @@ class ExecutionProcessor {
      */
     void execute_transaction(const Transaction& txn, Receipt& receipt) noexcept;
 
-    //! \brief Execute the block and write the result to the DB.
+    //! \brief Execute the block.
     //! \remarks Warning: This method does not verify state root; pre-Byzantium receipt root isn't validated either.
     //! \pre RuleSet's validate_block_header & pre_validate_block_body must return kOk.
-    [[nodiscard]] ValidationResult execute_and_write_block(std::vector<Receipt>& receipts) noexcept;
+    [[nodiscard]] ValidationResult execute_block(std::vector<Receipt>& receipts) noexcept;
+
+    //! \brief Flush IntraBlockState into cumulative State.
+    void flush_state();
 
     uint64_t available_gas() const noexcept;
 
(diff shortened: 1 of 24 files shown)


silkworm/core/execution/processor.hpp
```diff
@@ -41,10 +41,13 @@ class ExecutionProcessor {
...
 
-    //! \brief Execute the block and write the result to the DB.
+    //! \brief Execute the block.
     //! \remarks Warning: This method does not verify state root; pre-Byzantium receipt root isn't validated either.
     //! \pre RuleSet's validate_block_header & pre_validate_block_body must return kOk.
-    [[nodiscard]] ValidationResult execute_and_write_block(std::vector<Receipt>& receipts) noexcept;
+    [[nodiscard]] ValidationResult execute_block(std::vector<Receipt>& receipts) noexcept;
+
+    //! \brief Flush IntraBlockState into cumulative State.
+    void flush_state();
```
canepat, line 50: We should change this to:
```
[[nodiscard]] FlushingResult flush_state() noexcept;
```
in order to:

1. make `ExecutionProcessor` exception-free again
2. do not rely on `db::Buffer::MemoryLimitError` exception to handle an expected condition i.e. batch full

canepat (approved):

PR #2082, opened 2024-06-06
rpcdaemon: fix ots_getContractCreator

### LINKED ISSUE #2061: rpcdaemon: `ots_getContractCreator` always returns `null`
On my dev network, all invocations of `ots_getContractCreator` return `null` under Silkworm. When running Erigon without Silkworm it returns the hash and creator.

See https://github.com/otterscan/otterscan/blob/develop/docs/custom-jsonrpc.md#ots_getcontractcreator for reference.

lupin012:
Fixes #2061

### DIFF
--- a/silkworm/rpc/core/evm_executor.hpp
+++ b/silkworm/rpc/core/evm_executor.hpp
@@ -47,7 +47,8 @@ enum PreCheckErrorCode {
     kInternalError,
     kIntrinsicGasTooLow,
     kMaxFeePerBlobGasTooLowError,
-    kTipHigherThanFeeCap
+    kTipHigherThanFeeCap,
+    kIsNotLondon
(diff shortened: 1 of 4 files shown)


silkworm/rpc/core/evm_executor.hpp
```diff
@@ -47,7 +47,8 @@ enum PreCheckErrorCode {
     kInternalError,
     kIntrinsicGasTooLow,
     kMaxFeePerBlobGasTooLowError,
-    kTipHigherThanFeeCap
+    kTipHigherThanFeeCap,
+    kIsNotLondon
```
battlmonstr, line 51: should this list be sorted?

battlmonstr (approved):

PR #2131, opened 2024-06-20
sentry: absl::crc32c instead of stbrumme_crc32

yperbasis:
reduce the number of external dependencies


yperbasis:
Damn, apparently abseil has a [different flavour of CRC](https://stackoverflow.com/questions/26429360/crc32-vs-crc32c)

PR #2157, opened 2024-07-02
Fix build with clang 16

yperbasis:
See https://github.com/ethereum/evmc/pull/712.

I also had to revert https://github.com/ethereum/evmone/pull/517 in order to still be able to compile with clang 15.


chfast:
> I also had to revert [ethereum/evmone#517](https://github.com/ethereum/evmone/pull/517) in order to still be able to compile with clang 15.

Sorry about that. Do you want to align supported compilers?

yperbasis:
> > I also had to revert [ethereum/evmone#517](https://github.com/ethereum/evmone/pull/517) in order to still be able to compile with clang 15.
> 
> Sorry about that. Do you want to align supported compilers?

It's not a big deal. We can probably bump our minimum clang version to 16 at some point later.

canepat (approved):

PR #2163, opened 2024-07-04
rpcdaemon: fix clang tidy after PR 1939

canepat:


silkworm/rpc/core/evm_trace_test.cpp
```diff
@@ -47,11 +47,18 @@ using testing::_;
...
-static Bytes kZeroKey{*silkworm::from_hex("0000000000000000")};
-static Bytes kZeroHeader{*silkworm::from_hex("bf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a")};
-
-static Bytes kConfigKey{kZeroHeader};
-static Bytes kConfigValue{string_view_to_byte_view(kSepoliaConfig.to_json().dump())};
+static const Bytes kZeroKey{*silkworm::from_hex("0000000000000000")};
+static const Bytes kZeroHeader{*silkworm::from_hex("bf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a")};
+
+static const Bytes kConfigKey{kZeroHeader};
+static const Bytes kConfigValue{*from_hex(
```
yperbasis, line 54: Maybe just silence the warning? `kSepoliaConfig.to_json()` is more readably than the binary dump.

canepat: > Maybe just silence the warning? `kSepoliaConfig.to_json()` is more readably than the binary dump.

Sure. I was just a bit worried that this could be a case where "static initialisation order fiasco" can really happen, but I guess we'll figure it out.

BTW, this entire test module should definitely be refactored: it relies on many hard-coded binary dumps (included the one we're discussing), which will become unnecessary if we chose a better cut point for mocking (e.g. state access instead of cursor primitives).

PR #2172, opened 2024-07-10
rpcdaemon: fix debug/trace API family gas cost step 2

Sixtysixter:
Fix for gas cost in debug_trace API family


silkworm/rpc/core/evm_debug.hpp
```diff
@@ -75,6 +75,7 @@ struct CallFixes {
     int64_t stipend{0};
     int16_t code_cost{0};
     int64_t gas_cost{0};
+    bool precompiled{false};
 };
```
canepat, line 79: `CallFixes` is a little bit weird, I suggest renaming it as `CallGasRepair`. Also, `repair_call_gas_accounting` is probably better than `evaluate_call_fixes`.

You can do it in next PR

canepat (approved):

PR #2178, opened 2024-07-13
rpcdaemon: if gas bailout subtract only if there is enough balance

lupin012:


silkworm/core/execution/evm.cpp
```diff
@@ -149,7 +150,9 @@ evmc::Result EVM::create(const evmc_message& message) noexcept {
         state_.set_nonce(contract_addr, 1);
     }
 
-    state_.subtract_from_balance(message.sender, value);
+    if (have >= value) {
```
canepat, line 153: I would change this condition to:
```c++
if (!gas_bailout_ || have >= value)
```
in order to be more explicit and maybe also exploit short-circuiting.

silkworm/core/execution/evm.cpp
```diff
@@ -217,7 +221,9 @@ evmc::Result EVM::call(const evmc_message& message) noexcept {
             // https://github.com/ethereum/go-ethereum/blob/v1.9.25/core/vm/evm.go#L391
             state_.touch(message.recipient);
         } else {
-            state_.subtract_from_balance(message.sender, value);
+            if (have >= value) {
```
canepat, line 224: Same as above

canepat (changes requested):

silkworm/core/execution/evm.cpp
lupin012: ok

canepat (approved):

PR #2184, opened 2024-07-18
rpcdaemon: renamed fix call gas data structure

Sixtysixter:
Refactored data structure to fix call gas

### DIFF
--- a/silkworm/rpc/core/evm_debug.hpp
+++ b/silkworm/rpc/core/evm_debug.hpp
@@ -98,11 +98,11 @@ class DebugTracer : public EvmTracer {
 
   private:
     void write_log(const DebugLog& log);
-    void evaluate_call_fixes(unsigned char opcode,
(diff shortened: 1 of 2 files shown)


silkworm/rpc/core/evm_debug.hpp
```diff
@@ -111,7 +111,7 @@ class DebugTracer : public EvmTracer {
     const char* const* opcode_names_ = nullptr;
     const evmc_instruction_metrics* metrics_ = nullptr;
     std::stack<std::int64_t> start_gas_;
-    std::unique_ptr<CallFixes> call_fixes_;
+    std::unique_ptr<FixCallGasInfo> fix_call_gas_info_;
```
canepat, line 114: It's probably clearer and potentially also faster using `std::optional` instead of `std::unique_ptr` here.
You can do in next PR

canepat (approved):

PR #2220, opened 2024-08-09
db: snapshots repository thread safety

battlmonstr:
depends on https://github.com/erigontech/silkworm/pull/2215

Protect view_bundles and find_segment from threading issues without blocking too much.
The method is similar to the one used for e3 snapshots.

Bundles list is now a shared pointer and follows "copy on write" logic. If multiple readers want to read snapshots, they copy the source shared pointer (increasing a reference count). Modifying the bundles list (e.g. by add_snapshot_bundle) makes a copy. The old list and its iterators will be still usable by the readers from their shared pointers. When all the readers finish, the old list will be destroyed.

view_bundles() returns a BundlesView that holds the bundles list shared pointer for the iteration duration. This ensures that on bundles list modification the iterated list is not destroyed prematurely.

find_segment() additionally returns a bundle shared pointer. A reader must hold it while it operates on a snapshot/index. This ensures that on bundles list modification that snapshot is not closed prematurely.

### DIFF
--- a/silkworm/db/snapshots/repository.hpp
+++ b/silkworm/db/snapshots/repository.hpp
@@ -24,6 +24,7 @@
 #include <optional>
 #include <ranges>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include <silkworm/core/common/base.hpp>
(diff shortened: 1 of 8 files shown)


silkworm/db/snapshots/repository.hpp
```diff
@@ -73,13 +74,39 @@ class SnapshotRepository {
     [[nodiscard]] std::vector<std::shared_ptr<IndexBuilder>> missing_indexes() const;
     void remove_stale_indexes() const;
 
-    auto view_bundles() const { return make_map_values_view(bundles_); }
-    auto view_bundles_reverse() const { return std::ranges::reverse_view(view_bundles()); }
+    using TBundles = std::map<BlockNum, std::shared_ptr<SnapshotBundle>>;
```
canepat, line 77: We should rename `TBundles` as `Bundles` here because it's not a template typename like e.g. `TBaseView`. I know it's a minor thing, but I casually started reading from the cpp file and couldn't find what I expected to be a template parameter.

canepat (approved):

PR #2225, opened 2024-08-12
core: use evmone's hash functions precompiles

chfast:
Replace the implementations of BLAKE2bf, RIPEMD160 and SHA256 with ones from evmone. These implementations in evmone were delivered from Silkworm.

The SHA256 is also used outside of EVM for network communication.
Also, there are 2 other SHA256 implementations: in libsecp256k1 and blst.


chfast:
Let me try to also do SHA256.

yperbasis (approved):

PR #2227, opened 2024-08-12
db: integrate snapshot freezer

battlmonstr:
Run freezer data migration as a part of node tasks.
Freezer cleanup requires an RWTxn.
For that reason, schedule it in the stage loop via a helper TriggersStage.

### DIFF
--- a/silkworm/node/stagedsync/stages/stage_triggers.cpp
+++ b/silkworm/node/stagedsync/stages/stage_triggers.cpp
@@ -1,0 +1,59 @@
+/*
+   Copyright 2024 The Silkworm Authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


silkworm/node/stagedsync/stages/stage_triggers.cpp
```diff
@@ -0,0 +1,59 @@
...
+    io_context_.restart();
+    io_context_.run();
+
+    return Stage::Result::kSuccess;
+}
+
+Task<void> TriggersStage::schedule(std::function<Task<void>(db::RWTxn&)> task) {
+    auto task_caller = [this, t = std::move(task)]() -> Task<void> {
+        db::RWTxn* tx = this->current_tx_;
+        assert(tx);
```
canepat, line 48: I would use `SILKWORM_ASSERT` instead of `assert` here

canepat (approved):

silkworm/node/stagedsync/stages/stage_triggers.cpp
battlmonstr: I see the value in SILKWORM_ASSERT, maybe we should update to it everywhere?

battlmonstr: This could be one of the style proposals.

PR #2228, opened 2024-08-12
infra: enhance error message with grpc callback when port is in use

ertosns:
fix to #1398

### DIFF
--- a/cmd/test/CMakeLists.txt
+++ b/cmd/test/CMakeLists.txt
@@ -16,6 +16,7 @@
 
 find_package(Catch2 REQUIRED)
 find_package(magic_enum REQUIRED)
+find_package(gRPC REQUIRED)
 
 # Silkworm Core Tests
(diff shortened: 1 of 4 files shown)


canepat:
Thank you for your contribution!

Unfortunately, this PR is not ready to be considered for fixing #1398 due to the following reasons:
- the issue explicitly asks for improving the error message in the gRPC server, your change is printing a console message
- the unit test is not testing anything, it just registers the gRPC callback w/o triggering any invocation
- the way you're plugging the unit test in is unnecessarily changing the build procedure, instead it's sufficient to follow the existing naming conventions for other tests
- the linter complains about broken conventions

canepat:
@ertosns are you willing to work on this PR in the upcoming weeks?

ertosns:
Thanks for your feedback! @canepat 

no problem i can fix that.

canepat:
I'm going to close this PR next week if there is no activity

PR #2245, opened 2024-08-16
core: use blst from evmone

chfast:
Reuse evmone's configuration for blst dependency.


chfast:
This was an illustration and test to show that evmone's integration of blst can be propagated to silkworm. The difference is that evmone uses CMake ExternalProject to get the sources instead of git submodule. Do you think this is fine for you?

third_party/blst/CMakeLists.txt
```diff
@@ -1,55 +1 @@
...
-add_library(blst STATIC IMPORTED GLOBAL)
-add_dependencies(blst blst_build)
-# cmake-format: off
-set_target_properties(
-  blst
-  PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${BLST_DIR}/bindings
-             IMPORTED_LOCATION ${BLST_DIR}/${BLST_LIB}
-)
-# cmake-format: on
+include(../evmone/evmone/cmake/blst.cmake)
```
chfast, line 1: This is to minimize number of changes in Silkworm, but we can also remove this file and include evmone/cmake/blst.cmake directly.

PR #2251, opened 2024-08-20
core: decode account from E3 encoded storage

canepat:


silkworm/core/types/account.cpp
```diff
@@ -148,6 +148,55 @@ tl::expected<Account, DecodingError> Account::from_encoded_storage(ByteView enco
...
+            if (encoded_payload.length() == pos && i < 3) {
+                return tl::unexpected{DecodingError::kUnexpectedLength};
+            }
+            continue;
+        }
+        if (encoded_payload.length() < pos + len) {
+            return tl::unexpected{DecodingError::kInputTooShort};
+        }
+        const auto encoded_value{encoded_payload.substr(pos, len)};
+        switch (i) {
```
battlmonstr, line 169: this loop with a switch is a smart hack to structure the code
it took me a while to figure it out

a more basic way would be to extract this preambule into a helper function or a local lambda,
and then call it before each decoding attempt

battlmonstr (approved):

silkworm/core/types/account.cpp
canepat: I did it this way for similarity with the code already present in `Account::from_encoded_storage`

PR #2262, opened 2024-08-23
rpcdaemon: add STOP opcode as last opcode in specific success cases

lupin012:
Synthesise OP_STOP as last opcode in `trace` and `debug` API if execution is successful and last opcode executed is NOT in [OP_SELFDESTRUCT, OP_RETURN, OP_STOP]

### DIFF
--- a/silkworm/rpc/core/evm_trace.cpp
+++ b/silkworm/rpc/core/evm_trace.cpp
@@ -576,6 +576,7 @@ std::string to_string(intx::uint256 value) {
 }
 
 void VmTraceTracer::on_execution_start(evmc_revision rev, const evmc_message& msg, evmone::bytes_view code) noexcept {
+    last_opcode_ = std::nullopt;
     if (opcode_names_ == nullptr) {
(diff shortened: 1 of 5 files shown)


silkworm/rpc/core/evm_trace.cpp
```diff
@@ -778,6 +780,23 @@ void VmTraceTracer::on_execution_end(const evmc_result& result, const silkworm::
             }
             break;
     }
+
+    /* EVM WA: EVMONE add OP_STOP at the end of tx if not present but doesn't notify to the tracer. Add sw to add STOP to the op list */
+    if (result.status_code == EVMC_SUCCESS && last_opcode_ && last_opcode_ != OP_SELFDESTRUCT && last_opcode_ != OP_RETURN && last_opcode_ != OP_STOP) {
```
canepat, line 785: I suggest extracting this condition in a dedicated function not to repeat it twice. Next PR is fine

canepat (approved):

PR #2263, opened 2024-08-23
db: web seed client for downloading snapshots

canepat:


silkworm/db/snapshots/bittorrent/web_seed_client.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <vector>
+
+#include <boost/asio/io_context.hpp>
+#include <boost/url/url.hpp>
+// Disable warnings raised during compilation of libtorrent
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wc++11-compat"
+#pragma GCC diagnostic ignored "-Wshadow"
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+#include <libtorrent/torrent_info.hpp>
```
battlmonstr, line 32: is there a way to hide these inside third_party/libtorrent/CMakeLists.txt ?

silkworm/db/snapshots/bittorrent/web_seed_client.hpp
```diff
@@ -0,0 +1,82 @@
...
+    return lhs->name() < rhs->name();
+};
+using TorrentInfoPtrList = std::set<TorrentInfoPtr, decltype(torrent_info_compare)>;
+
+class WebSeedClient {
+  public:
+    WebSeedClient(std::vector<std::string>&& url_seeds, const PreverifiedList& preverified);
+    WebSeedClient(std::unique_ptr<WebSession>&& web_session,
+                  std::vector<std::string>&& url_seeds,
+                  const PreverifiedList& preverified);
```
battlmonstr, line 56:
```suggestion
    WebSeedClient(std::vector<std::string> url_seeds, const PreverifiedList& preverified);
    WebSeedClient(std::unique_ptr<WebSession> web_session,
                  std::vector<std::string> url_seeds,
                  const PreverifiedList& preverified);
```

I suggest to pass by value by default instead of move-only ref (unless we absolutely have to).
This is a convention I use in my code, and I'd like it to be a common codebase convention. I'll add it to the proposal doc for further discussion.

silkworm/db/snapshots/bittorrent/web_seed_client.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    [[nodiscard]] Task<TorrentInfoPtrList> discover_torrents(bool fail_fast = false);
```
battlmonstr, line 58:
```suggestion
    Task<TorrentInfoPtrList> discover_torrents(bool fail_fast = false);
```

I've disabled `modernize-use-nodiscard` a while ago to stop `[[nodiscard]]` proliferation. I suggest to just remove it here, or be consistent and add it everywhere in this file. I'll add a point to the code style proposal to consider removing it everywhere.

silkworm/db/snapshots/bittorrent/web_session.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+namespace silkworm::snapshots::bittorrent {
+
+namespace net = boost::asio;
+namespace ssl = net::ssl;
+namespace urls = boost::urls;
+
+//! Custom HTTP headers to include in any request to web servers hosted by Cloudflare
+const std::map<std::string_view, std::string_view> kCloudflareHeaders{
+    {"lsjdjwcush6jbnjj3jnjscoscisoc5s", "I%OSJDNFKE783DDHHJD873EFSIVNI7384R78SSJBJBCCJBC32JABBJCBJK45"},
```
battlmonstr, line 42: where does this value come from?

silkworm/db/snapshots/bittorrent/web_session.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // These objects perform our I/O
+    net::ip::tcp::resolver resolver{co_await net::this_coro::executor};
+    beast::ssl_stream<beast::tcp_stream> ssl_stream{co_await net::this_coro::executor, ssl_ctx};
+    beast::tcp_stream& tcp_stream = beast::get_lowest_layer(ssl_stream);
+
+    // Set SNI Hostname (many hosts need this to handshake successfully)
+    const std::string host{web_url.host()};
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
```
battlmonstr, line 66: why is this needed?

silkworm/db/snapshots/bittorrent/web_session.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <silkworm/infra/concurrency/task.hpp>
+
+#include <boost/beast/http.hpp>
+#include <boost/url/url.hpp>
+
+namespace silkworm::snapshots::bittorrent {
+
+namespace beast = boost::beast;
+namespace http = beast::http;
+namespace urls = boost::urls;
```
battlmonstr, line 32: normally we don't alias namespaces in .hpp files, could this be moved to .cpp?

silkworm/db/snapshots/bittorrent/web_session.hpp
```diff
@@ -0,0 +1,58 @@
...
+    //! \param web_url the URL address of the web server
+    //! \param target the relative path of the requested file
+    [[nodiscard]] virtual Task<StringResponse> https_get(const urls::url& web_url, std::string_view target_file) const;
+
+  protected:
+    using EmptyRequest = http::request<http::empty_body>;
+    static void include_cloudflare_headers(EmptyRequest& request);
+
+    //! The HTTP protocol version to use
+    static constexpr int kHttpVersion{11};
```
battlmonstr, line 53: does BEAST and/or cloudflare support HTTP 2?

battlmonstr (changes requested):

silkworm/db/snapshots/bittorrent/web_seed_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+Task<void> WebSeedClient::build_list_of_torrents(bool fail_fast) {
+    using namespace concurrency::awaitable_wait_for_one;
+
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-capturing-lambda-coroutines)
+    auto build_list_of_torrents_factory = [this, fail_fast](size_t index) -> Task<void> {
+        try {
+            const auto& provider_url = url_seeds_[index];
+            co_await build_list_of_torrents(provider_url);
+        } catch (const std::exception&) {
+            if (fail_fast) throw;
```
battlmonstr, line 86: if fail_fast = false, we shouldn't ignore the exception

silkworm/db/snapshots/bittorrent/web_seed_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+        }
+    };
+
+    // Parallelize async build of the list of torrent files for each provider
+    auto group_task = concurrency::generate_parallel_group_task(url_seeds_.size(), build_list_of_torrents_factory);
+    try {
+        co_await (std::move(group_task) || concurrency::timeout(kParallelManifestDownloadTimeout));
+    } catch (const concurrency::TimeoutExpiredError&) {
+        throw;
+    }
```
battlmonstr, line 96: this is redundant

silkworm/db/snapshots/bittorrent/web_seed_client.cpp
```diff
@@ -0,0 +1,201 @@
...
+            torrent_infos.emplace(std::move(torrent_info));
+        }
+    };
+
+    // Parallelize async download and validate of torrent files for each provider
+    auto group_task = concurrency::generate_parallel_group_task(torrent_files.size(), download_and_validate_factory);
+    try {
+        co_await (std::move(group_task) || concurrency::timeout(kParallelTorrentDownloadTimeout));
+    } catch (const concurrency::TimeoutExpiredError&) {
+        throw;
```
battlmonstr, line 153: this is redundant

silkworm/db/snapshots/bittorrent/web_session.hpp
```diff
@@ -0,0 +1,58 @@
...
+    //! \brief Constructor optionally accepting a server certificate to append to root certificates
+    //! \remarks the custom server certificate is used only in tests
+    explicit WebSession(std::optional<std::string> server_certificate = {});
+    virtual ~WebSession() = default;
+
+    using StringResponse = http::response<http::string_body>;
+
+    //! \brief Asynch send a HTTPS GET request for \p target file to \p web_url and receive the response
+    //! \param web_url the URL address of the web server
+    //! \param target the relative path of the requested file
```
battlmonstr, line 45:
```suggestion
    //! \param target_file the relative path of the requested file
```

silkworm/db/snapshots/bittorrent/web_session.hpp
```diff
@@ -0,0 +1,58 @@
...
+    [[nodiscard]] virtual Task<StringResponse> https_get(const urls::url& web_url, std::string_view target_file) const;
+
+  protected:
+    using EmptyRequest = http::request<http::empty_body>;
+    static void include_cloudflare_headers(EmptyRequest& request);
```
battlmonstr, line 50: I'd pass extra kCloudflareHeaders as a parameter to https_get instead of adding it regardless.

silkworm/db/snapshots/bittorrent/web_session.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // Send the HTTP request to the remote host
+    tcp_stream.expires_after(kHttpTimeoutSecs);
+    const auto written_bytes = co_await http::async_write(ssl_stream, req, net::use_awaitable);
+    SILK_TRACE << "WebSeedClient::http_session HTTP request written_bytes: " << written_bytes;
+
+    // This buffer is used for reading
+    beast::flat_buffer data;
+
+    // Declare a container to hold the response
+    http::response<http::string_body> rsp;
```
battlmonstr, line 102: what's the relation between these? how come the data in flat_buffer survives after this returns?

silkworm/db/snapshots/bittorrent/web_session.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    // Receive the HTTP response
+    const auto read_bytes = co_await http::async_read(ssl_stream, data, rsp, net::use_awaitable);
```
battlmonstr, line 105: how big are the downloaded files?

battlmonstr (changes requested):

silkworm/db/snapshots/bittorrent/web_seed_client.hpp
battlmonstr: Task<T> class is already marked with `[[nodiscard]]`, which applies to all code by default

silkworm/db/snapshots/bittorrent/root_certificates.hpp
```diff
@@ -0,0 +1,3956 @@
...
+// See also: https://github.com/boostorg/beast/blob/develop/example/common/root_certificates.hpp
+
+#include <optional>
+#include <string>
+
+#include <boost/asio/ssl.hpp>
+
+namespace silkworm::snapshots::bittorrent {
+
+// Including root certificates is theoretically a bad practice for performing
```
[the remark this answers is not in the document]
battlmonstr: Either generate it using something like embed.cpp, or remove unused certs.

silkworm/db/snapshots/bittorrent/web_session.hpp
canepat: No, it doesn't: https://github.com/boostorg/beast/issues/1302

silkworm/db/snapshots/bittorrent/web_session.cpp
battlmonstr: rsp -> response

cmd/dev/snapshots.cpp
```diff
@@ -382,6 +435,21 @@ void download(const bittorrent::BitTorrentSettings& settings) {
...
+
+    bittorrent::TorrentInfoPtrList web_seed_torrents;
+    if (!settings.url_seed.empty()) {
+        web_seed_torrents = download_web_seed(settings);
+
+        if (settings.download_web_seed_torrents) {
+            download_bittorrent(settings, web_seed_torrents);
+        }
+    } else {
+        download_bittorrent(settings, web_seed_torrents);
```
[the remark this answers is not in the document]
canepat: If you don't specify any web seed URL, it will download the target files specified in input file containing a list of magnet links. Your doubt is absolutely legit, the code is not clear enough: I'm going to change it and make it more readable.

silkworm/db/snapshots/bittorrent/root_certificates.hpp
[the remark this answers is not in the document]
canepat: Sure, I was a bit lazy and just copied the root certificates from some Boost Beast example (see comment in code), but I do agree it's a far from ideal setup. On the other hand, doing it properly can be quite overkilled for our usage here, so as discussed I'm going to reduce the list of hard-coded certificates to the bare minimum by removing the unused ones.

silkworm/db/snapshots/bittorrent/web_seed_client.hpp
canepat: > I've disabled `modernize-use-nodiscard` a while ago to stop `[[nodiscard]]` proliferation. I suggest to just remove it here, or be consistent and add it everywhere in this file. I'll add a point to the code style proposal to consider removing it everywhere.

I would suggest to remove it _almost_ everywhere: it's valuable using `[[nodiscard]]` sporadically in cases where it signals some potential misbehaviour at the call site (e.g. error conditions as return values, discarded result after heavy computation...)

silkworm/db/snapshots/bittorrent/web_session.cpp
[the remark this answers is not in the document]
canepat: Good catch!

silkworm/db/snapshots/bittorrent/web_session.cpp
canepat: It comes from Erigon [here](https://github.com/erigontech/erigon/blob/main/erigon-lib/downloader/downloader.go#L154)

silkworm/db/snapshots/bittorrent/web_session.cpp
canepat: It's probably a leftover, I will remove it

silkworm/db/snapshots/bittorrent/web_session.cpp
canepat: > what's the relation between these? how come the data in flat_buffer survives after this returns?

https://www.boost.org/doc/libs/1_86_0/libs/beast/doc/html/beast/ref/boost__beast__http__async_read/overload2.html

silkworm/db/snapshots/bittorrent/web_session.cpp
canepat: It's needed because of the `SSL_set_tlsext_host_name` signature, see failure [here](https://app.circleci.com/pipelines/github/erigontech/silkworm/12776/workflows/cbfb0a35-3e31-4af4-b77a-8d535e848b66/jobs/56511)

silkworm/db/snapshots/bittorrent/web_seed_client.hpp
canepat: I would prefer to address this issue in another PR, because the same directives to disable warnings are present also on other pre-existing header files, e.g. [here](https://github.com/erigontech/silkworm/blob/master/silkworm/db/snapshots/bittorrent/client.hpp#L34)

battlmonstr (approved):

PR #2265, opened 2024-08-26
third-party: update erigon-snapshot to da5d0f5

bzawisto:
Here's the list of updated snapshots for mainnet, bor and sepolia. In addition there's an extra function that filters out unsupported snapshot files (introduced in erigon 3). One of the tests has been simplified since the ranges have changed at some point (from 500k to 100k) and later down to even 10k (hence computing snapshots count have become unreliable).


silkworm/db/snapshots/config.cpp
```diff
@@ -60,4 +63,16 @@ BlockNum Config::compute_max_block() {
     return max_block > 0 ? max_block - 1 : 0;
 }
 
+void Config::remove_unsupported_snapshots() {
+    constexpr std::array UNSUPPORTED_SNAPSHOT_NAME_TOKENS = {
```
canepat, line 67: In terms of code style we follow Google C++ Style Guidelines, where the naming convention for constants is [this](https://google.github.io/styleguide/cppguide.html#Constant_Names). Hence, this one should read:
```
kUnsupportedSnapshotNameTokens
```

canepat (approved):

PR #2268, opened 2024-08-27
Add snapshot setting to keep exported blocks in mdbx

bzawisto:
As title says.

### DIFF
--- a/cmd/common/snapshot_options.cpp
+++ b/cmd/common/snapshot_options.cpp
@@ -26,7 +26,7 @@ void add_snapshot_options(CLI::App& cli, snapshots::SnapshotSettings& snapshot_s
         ->description("If set, the snapshot downloader is disabled and just already present local snapshots are used")
         ->capture_default_str();
     cli.add_flag("--snapshots.keepblocks", snapshot_settings.keepblocks)
-        ->description("Flag indicating if exported blocks into snapshots should be still kept in mbdbx")
+        ->description("Flag indicating if exported blocks into snapshots should be still kept in mdbx")
         ->capture_default_str();
     cli.add_option("--snapshots.repository.path", snapshot_settings.repository_dir)
         ->description("Filesystem path where snapshots will be stored")
--- a/cmd/common/snapshot_options.cpp
+++ b/cmd/common/snapshot_options.cpp
@@ -25,7 +25,7 @@ void add_snapshot_options(CLI::App& cli, snapshots::SnapshotSettings& snapshot_s
     cli.add_flag("--snapshots.no_downloader", snapshot_settings.no_downloader)
         ->description("If set, the snapshot downloader is disabled and just already present local snapshots are used")
         ->capture_default_str();
-    cli.add_flag("--snapshots.keepblocks", snapshot_settings.keepblocks)
+    cli.add_flag("--snapshots.keepblocks", snapshot_settings.keep_blocks)
         ->description("If set, the blocks exported from mdbx to snapshots are kept in mdbx")
         ->capture_default_str();
     cli.add_option("--snapshots.repository.path", snapshot_settings.repository_dir)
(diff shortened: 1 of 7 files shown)


cmd/common/snapshot_options.cpp
```diff
@@ -25,6 +25,9 @@ void add_snapshot_options(CLI::App& cli, snapshots::SnapshotSettings& snapshot_s
     cli.add_flag("--snapshots.no_downloader", snapshot_settings.no_downloader)
         ->description("If set, the snapshot downloader is disabled and just already present local snapshots are used")
         ->capture_default_str();
+    cli.add_flag("--snapshots.keepblocks", snapshot_settings.keepblocks)
+        ->description("Flag indicating if exported blocks into snapshots should be still kept in mbdbx")
```
battlmonstr:
```suggestion
        ->description("If set, the blocks exported from mdbx to snapshots are kept in mdbx")
```

cmd/dev/db_toolbox.cpp
```diff
@@ -2397,6 +2397,8 @@ int main(int argc, char* argv[]) {
     // Freeze command
     auto cmd_freeze = app_main.add_subcommand("freeze", "Migrate data to snapshots");
 
+    auto cmd_freeze_keep_blocks_opt = cmd_freeze->add_flag("--snap.keepblocks", "Keeps exported blocks in mdbx");
```
battlmonstr, line 2400:
```suggestion
    auto cmd_freeze_keep_blocks_opt = cmd_freeze->add_flag("--snap.keepblocks", "If set, the blocks exported from mdbx to snapshots are kept in mdbx");
```

cmd/dev/db_toolbox.cpp
```diff
@@ -2505,7 +2507,7 @@ int main(int argc, char* argv[]) {
         } else if (*cmd_reset_to_download) {
             do_reset_to_download(src_config, static_cast<bool>(*cmd_reset_to_download_keep_senders_opt));
         } else if (*cmd_freeze) {
-            do_freeze(src_config, data_dir);
+            do_freeze(src_config, data_dir, cmd_freeze_keep_blocks_opt);
```
battlmonstr, line 2510:
```suggestion
            do_freeze(src_config, data_dir, static_cast<bool>(*cmd_freeze_keep_blocks_opt));
```

silkworm/db/freezer.cpp
```diff
@@ -156,6 +156,11 @@ Task<void> Freezer::cleanup() {
     if (range.first >= range.second) co_return;
     log::Debug(name()) << "cleanup [" << range.first << ", " << range.second << ")";
 
+    if (keep_blocks_) {
+        log::Debug(name()) << " skipping cleanup [" << range.first << ", " << range.second << ")";
```
battlmonstr, line 160:
```suggestion
        log::Debug(name()) << "skipping cleanup";
```
The range is already printed above. The initial space is not needed, everything that is pushed to the logger is right-aligned as a separate column.

silkworm/db/snapshots/settings.hpp
```diff
@@ -28,6 +28,7 @@ struct SnapshotSettings {
     bool enabled{true};                                                        // Flag indicating if snapshots are enabled
     bool no_downloader{false};                                                 // Flag indicating if snapshots download is disabled
     bittorrent::BitTorrentSettings bittorrent_settings;                        // The Bittorrent protocol settings
+    bool keepblocks{false};                                                    // Flag indicating if exported blocks should be kept in mdbx
```
battlmonstr, line 31:
```suggestion
    bool keep_blocks{false};                                                    // Flag indicating if exported blocks should be kept in mdbx
```

We have to keep the flag names compatible with erigon. Unfortunately they don't follow a certain convention when it comes to word separation. Nevertheless, inside silkworm we should do it appropriately.

battlmonstr (changes requested):

battlmonstr (approved):
👍

PR #2270, opened 2024-08-27
db: snapshot merger cleanup

battlmonstr:
Delete merged snapshot files.

### DIFF
--- a/silkworm/db/snapshot_merger.cpp
+++ b/silkworm/db/snapshot_merger.cpp
@@ -17,6 +17,9 @@
 #include "snapshot_merger.hpp"
 
 #include <algorithm>
+#include <filesystem>
+#include <memory>
+#include <vector>
 
 #include <silkworm/infra/common/filesystem.hpp>
 #include <silkworm/infra/common/log.hpp>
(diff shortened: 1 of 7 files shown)


silkworm/db/snapshot_merger.cpp
```diff
@@ -84,6 +87,14 @@ struct RawSnapshotWordDeserializer : public SnapshotWordDeserializer {
     }
 };
 
+static std::vector<std::shared_ptr<SnapshotBundle>> bundles_in_range(BlockNumRange range, SnapshotRepository& repository) {
+    std::vector<std::shared_ptr<SnapshotBundle>> bundles;
+    for (auto& bundle : repository.view_bundles())
+        if (range.contains(bundle->block_from()) && range.contains(bundle->block_to() - 1))
+            bundles.push_back(bundle);
+    return bundles;
```
canepat, line 95:
```suggestion
    std::vector<std::shared_ptr<SnapshotBundle>> bundles;
    for (auto& bundle : repository.view_bundles()) {
        if (range.contains(bundle->block_from()) && range.contains(bundle->block_to() - 1)) {
            bundles.push_back(bundle);
        }
    }
    return bundles;
```
Even if Google C++ Style Guide allows [not to use curly braces for one statement](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching), here we have a loop with one condition inside, so I believe it's better to add braces, which makes the code also more consistent with the rest (see below in `schedule_bundle_cleanup`).

silkworm/db/snapshots/snapshot_bundle.hpp
```diff
@@ -31,7 +31,7 @@
 
 namespace silkworm::snapshots {
 
-struct SnapshotBundle {
+struct SnapshotBundleStruct {
```
canepat, line 34: `SnapshotBundleStruct` -> `SnapshotBundleData`

canepat (changes requested):

canepat (approved):

Issue #2230, opened 2024-08-12
fuzzy testing: workflow broken after PR 2217


canepat:
https://github.com/erigontech/silkworm/actions/runs/10335502673

PR #2276, opened 2024-08-29
fuzzer: fix CLang 16

JacekGlen:
- Update the `compiler_install.sh` script
- Update the job definition to ensure that the build config is always cleaned up
- Increase memory available to each fuzzer job


cmake/setup/compiler_install.sh
```diff
@@ -17,15 +17,31 @@ function install_gcc {
...
-	sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++ 100
-	sudo update-alternatives --install /usr/bin/llvm-cov llvm-cov /usr/bin/llvm-cov-$CLANG_VERSION 100
-	sudo update-alternatives --install /usr/bin/llvm-profdata llvm-profdata /usr/bin/llvm-profdata-$CLANG_VERSION 100
+	
+	if [[ "$CLANG_VERSION" -ge 16 ]]
+	then
+		# Clang 16 and later versions have a different package repository
+		# Get the latest installation script and execute
+ 		sudo rm -f llvm.sh
+ 		sudo wget https://apt.llvm.org/llvm.sh
```
battlmonstr, line 27: this shell script should be committed into third_party (as we do for wasmer) and reviewed

cmake/setup/compiler_install.sh
```diff
@@ -17,15 +17,31 @@ function install_gcc {
...
+		sudo chmod u+x llvm.sh
+		sudo ./llvm.sh $CLANG_VERSION
+		sudo apt-get install -y libc++-$CLANG_VERSION-dev libc++abi-$CLANG_VERSION-dev
+		sudo ln -sfv /usr/bin/clang-$CLANG_VERSION /usr/bin/clang
+		sudo ln -sfv /usr/bin/clang++-$CLANG_VERSION /usr/bin/clang++
+		sudo update-alternatives --install /usr/bin/cc cc /usr/bin/clang 100
+		sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++ 100
```
battlmonstr, line 34: this section is almost identical to the lines below, could we refactor it to not copy paste?

in particular, we need llvm-cov and llvm-profdata for some CI jobs

battlmonstr (changes requested):
@JacekGlen this new shell script seems to support the old versions as well. Could we use it for everything?

cmake/setup/compiler_install.sh
JacekGlen: Done

cmake/setup/compiler_install.sh
JacekGlen: Done

JacekGlen:
> @JacekGlen this new shell script seems to support the old versions as well. Could we use it for everything?

I think the new approach combines the best of two worlds: it is backwards compatible and uses llvm shell script for all versions.

The reason I didn't use the llvm script for everything is that their apt repository do not all the versions available for all the distributions. For example they only have CLang >= 17 for Ubuntu 24, see: https://apt.llvm.org/noble/dists/

battlmonstr (approved):

JacekGlen:
Fixes #2230

PR #2284, opened 2024-08-30
core: implement bor_transfer

yperbasis:
See https://github.com/maticnetwork/bor/blob/v1.3.7/core/bor_fee_log.go & https://github.com/erigontech/erigon/pull/10810

### DIFF
--- a/silkworm/core/execution/evm.hpp
+++ b/silkworm/core/execution/evm.hpp
@@ -23,6 +23,7 @@
 #include <evmone/baseline.hpp>
 #include <evmone/execution_state.hpp>
 #include <evmone/vm.hpp>
+#include <gsl/pointers>
 #include <intx/intx.hpp>
 
 #include <silkworm/core/chain/config.hpp>
@@ -72,6 +73,18 @@ using EvmTracers = std::vector<std::reference_wrapper<EvmTracer>>;
 
 using AnalysisCache = LruCache<evmc::bytes32, std::shared_ptr<evmone::baseline::CodeAnalysis>>;
 
+using TransferFunc = void(IntraBlockState& state, const evmc::address& sender, const evmc::address& recipient,
+                          const intx::uint256& amount, bool bailout);
+
+// See consensus.Transfer in Erigon
+inline void standard_transfer(IntraBlockState& state, const evmc::address& sender, const evmc::address& recipient,
+                              const intx::uint256& amount, bool bailout) {
+    if (!bailout) {
+        state.subtract_from_balance(sender, amount);
+    }
+    state.add_to_balance(recipient, amount);
+}
+
 class EVM {
   public:
     // Not copyable nor movable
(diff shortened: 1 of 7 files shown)


silkworm/core/execution/evm.hpp
```diff
@@ -72,6 +73,19 @@ using EvmTracers = std::vector<std::reference_wrapper<EvmTracer>>;
 
 using AnalysisCache = LruCache<evmc::bytes32, std::shared_ptr<evmone::baseline::CodeAnalysis>>;
 
+using TransferFunc = void(IntraBlockState& state, const evmc::address& sender, const evmc::address& recipient,
+                          const intx::uint256& amount, bool bailout);
+
+// See consensus.Transfer in Erigon
+inline void standard_transfer(IntraBlockState& state, const evmc::address& sender, const evmc::address& recipient,
+                              const intx::uint256& amount, bool bailout) {
+    // TODO(yperbasis) why is the bailout condition different from Erigon?
```
canepat, line 82: Probably it appears to be different because in Erigon the bailout mechanism is handled all in [one module](https://github.com/erigontech/erigon/blob/main/core/state_transition.go), whilst in Silkworm it's unfortunately done in two different modules:
- in `rpc::EVMExecutor` for the [gas](https://github.com/erigontech/silkworm/blob/master/silkworm/rpc/core/evm_executor.cpp#L300)
- here in `EMV` for the value transferred

It is worth it to amend this, so I'm going open an issue on this topic.

Moreover, there's another very related topic that's worth improving as well: `rpc::EVMExecutor` captures a (simulated) EVM execution not so differently wrt `ExecutionProcessor` and exists for historical reasons (i.e. it comes from standalone Silkrpc project), but I've been having this idea for a while of getting rid of it by moving its additional functionality into `ExecutionProcessor`. I will open an issue also for this.

silkworm/core/protocol/bor_rule_set.cpp
```diff
@@ -141,37 +141,70 @@ evmc::address BorRuleSet::get_beneficiary(const BlockHeader& header) {
...
+            0xe6497e3ee548a3372136af2fcb0696db31fc6cf20260707645068bd3fe97f3c4_bytes32};
+        intx::uint256 sender_initial_balance{state.get_balance(sender)};
+        intx::uint256 recipient_initial_balance{state.get_balance(recipient)};
+        // TODO(yperbasis) why is the bailout condition different from that of Erigon?
+        if (!bailout || sender_initial_balance >= amount) {
+            state.subtract_from_balance(sender, amount);
+        }
+        state.add_to_balance(recipient, amount);
+        intx::uint256 output1{state.get_balance(sender)};
+        intx::uint256 output2{state.get_balance(recipient)};
```
canepat, line 189: Here it could be better renaming:

`output1` -> `sender_final_balance`
`output2` -> `recipient_final_balance`

canepat (approved):

PR #2291, opened 2024-09-02
db: fix completed count when the same snapshot is downloaded twice

canepat:
This PR adjusts the completed count during snapshot sync because the same snapshot file can be downloaded more than once if its content changes, hence it can be notified multiple times in the completed signal of `BitTorrentClient`.

### DIFF
--- a/silkworm/db/snapshot_sync.cpp
+++ b/silkworm/db/snapshot_sync.cpp
@@ -41,6 +41,16 @@ static constexpr std::chrono::seconds kCheckCompletionInterval{1};
 
 using namespace silkworm::snapshots;
 
+//! \warning Hash provider for std::filesystem::path necessary to avoid the following error in Clang + LLVM 16
+//! \verbatim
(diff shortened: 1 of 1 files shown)


silkworm/db/snapshot_sync.cpp
```diff
@@ -135,6 +145,14 @@ bool SnapshotSync::download_snapshots(const std::vector<std::string>& snapshot_f
 
     std::latch download_done{num_snapshots};
     auto log_completed = [&](const std::filesystem::path& snapshot_file) {
+        // The same snapshot segment may be downloaded multiple times in case of content change over time and
+        // hence notified for completion multiple times. We need to count each snapshot segment just once here
+        static std::unordered_set<std::filesystem::path, PathHasher> snapshot_set;
```
battlmonstr, line 150: I'd move this before the lambda as a local variable in order to make SnapshotSync::download_snapshots callable multiple times

battlmonstr (approved):

PR #2292, opened 2024-09-02
cmd: support version update in schema command of snapshots tool

canepat:
This PR adds `force_update` option to `schema` subcommand of `snapshots` command-line utility to update the database schema to the current Silkworm code version. This is useful when upgrading the code between compatible versions or if at some point embedded versioning was wrong (i.e. not correctly updated to match Erigon's one).

### DIFF
--- a/cmd/dev/db_toolbox.cpp
+++ b/cmd/dev/db_toolbox.cpp
@@ -698,18 +698,21 @@ void do_freelist(db::EnvConfig& config, bool detail) {
     env.close(config.shared);
 }
 
-void do_schema(db::EnvConfig& config) {
+void do_schema(db::EnvConfig& config, bool force_update) {
     auto env{silkworm::db::open_env(config)};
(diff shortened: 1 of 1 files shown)


cmd/dev/db_toolbox.cpp
```diff
@@ -2442,7 +2449,7 @@ int main(int argc, char* argv[]) {
         } else if (*cmd_freelist) {
             do_freelist(src_config, static_cast<bool>(*freelist_detail_opt));
         } else if (*cmd_schema) {
-            do_schema(src_config);
+            do_schema(src_config, cmd_schema_force_version_update_opt->as<bool>());
```
battlmonstr, line 2452:
```suggestion
            do_schema(src_config, static_cast<bool>(*cmd_schema_force_version_update_opt));
```

consistent with the style in this file

battlmonstr (approved):

PR #2307, opened 2024-09-05
Use the same block executor for both CAPI and StagedSync

bzawisto:


silkworm/node/execution/block/block_executor.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <silkworm/core/execution/processor.hpp>
+#include <silkworm/infra/common/log.hpp>
+#include <silkworm/node/execution/block/block_executor.hpp>
+
+namespace silkworm::execution::block {
+
+using namespace std::chrono_literals;
+
+//! Generate log arguments for execution progress at specified block
+static log::Args log_args_for_exec_progress(BlockExecutor::ExecutionProgress& progress, uint64_t current_block) {
```
canepat, line 29: This logging format is specifically crafted for Erigon++ to match Erigon logs and must be kept into the C API implementation. The stage pipeline has its own timer-based logging mechanism that must remain the same because it's shared among all the stages.

silkworm/node/execution/block/block_executor.cpp
```diff
@@ -0,0 +1,117 @@
...
+    progress_.processed_gas += block.header.gas_used;
+
+    const auto now{std::chrono::steady_clock::now()};
+    if (log_time_ <= now) {
+        progress_.batch_progress_perc = static_cast<float>(state_buffer.current_batch_state_size()) / static_cast<float>(max_batch_size_);
+        progress_.end_time = now;
+        log::Info{"[4/12 Execution] Executed blocks",  // NOLINT(*-unused-raii)
+                  log_args_for_exec_progress(progress_, block.header.number)};
+        log_time_ = now + 20s;
+    }
```
canepat, line 113: This logging format is specifically crafted for Erigon++ to match Erigon logs and must be kept into the C API implementation. The stage pipeline has its own timer-based logging mechanism that must remain the same because it's shared among all the stages.

canepat (changes requested):

canepat (approved):

PR #2308, opened 2024-09-05
capi: Fix batch size control 

JacekGlen:
The problem with the current approach is that it catches `db::Buffer::MemoryLimitError` too late, when the block has already been popped from the `block_buffer`. Then in the next iteration we would get the next block and the assert `SILKWORM_ASSERT(block->header.number == block_number)` would fail. For this to work we need to re-insert the failed block in the exception handler. 

I think the better approach is to control the batch size with the `state_buffer.current_batch_state_size() <= max_batch_size` check. Additionally, using exceptions for control flow is generally discouraged. 

I changed the unit test to cover for this scenario. Also I have tested the change by synchronizing the full Sepolia chain data.


battlmonstr (changes requested):
This was done like you suggest before, but was changed in order to [fix OOM crashes on low RAM VMs](https://github.com/erigontech/silkworm/pull/2081). The current_batch_state_size check doesn't guarantee that the next iteration execution overflows memory. When it fails to execute due to a potential overflow, it produces an error. The condition for that happens multiple layers down, so I've decided to use an exception and not result/error codes.

I think it is likely my fault that the logic is broken. I agree that on MemoryLimitError it should push the block back to the block_buffer, because the block was not executed, and needs to be executed again (but block_buffer.pop_back won't return it).

For silkworm_execute_blocks_ephemeral this is not a problem, because prefetched_blocks.pop_front is done only for successful executions.

Alternatively consider to align silkworm_execute_blocks_perpetual logic to do the same and pop only on success.

JacekGlen:
> This was done like you suggest before, but was changed in order to [fix OOM crashes on low RAM VMs](https://github.com/erigontech/silkworm/pull/2081). The current_batch_state_size check doesn't guarantee that the next iteration execution overflows memory. When it fails to execute due to a potential overflow, it produces an error. The condition for that happens multiple layers inside, so I've decided to use an exception and not result/error codes.
> 
> I think it is likely my fault that the logic is broken here. I agree that on MemoryLimitError it should reinsert the block back to the block_buffer, because the block was not executed, and needs to be executed again (for block_buffer.pop_back to return it again).
> 
> For silkworm_execute_blocks_ephemeral this is not a problem, because prefetched_blocks.pop_front is done only for successful executions.
> 
> Alternatively consider to align silkworm_execute_blocks_perpetual logic to do the same and pop only on success.


I am split on this, so let's pick @canepat brain. To summarize:

The current logic is:
1. Execute blocks until MemoryLimit exception
2. Re-insert the last block
3. Commit to the db

I see two problems with this: we use exceptions to control program flow and some blocks must be executed twice. Even if we decide to pop the block only on success, we still face these two issues.

The original approach was:
1. Execute blocks until the current batch size exceeds MaxBatchSize
2. Commit to the db

The problem is that we find out if the block exceeds the limit only **after** the execution. In some cases, this can cause out-of-memory exceptions. There is no easy way to detect if a block exceeds the limit before the execution. 

Some potential fixes:
- Go deeper into the execution and split it into execute and flush.  So we can potentially commit to the db before flushing to the buffer
- Using the original approach, lower the MaxBatchSize by some arbitrary number (e.g. 10%, or 1k) so executing a block will never raise and exception (but batch sizes will be different to Erigon)
- ?

battlmonstr (approved):
:pray:

battlmonstr:
silkworm_capi_test hangs on CI though

JacekGlen:
> silkworm_capi_test hangs on CI though

@battlmonstr issue fixed, please review again

battlmonstr (approved):
:+1:

PR #2325, opened 2024-09-08
rpcdaemon: gas cost from EVMONE for debug family API

Sixtysixter:
Relevant changes:
- use customised tracing API from our evmone [fork](https://github.com/erigontech/evmone/tree/gas_cost_in_tracers)
- get the _internal_ gas cost from EVMONE for debug_trace* API


canepat (approved):
Even if the code is not clean enough yet, I'm really keen on merging this PR because it introduces the improved evmone tracing API.
@Sixtysixter as discussed, you will do a follow-up PR to clean up the code

Issue #2304, opened 2024-09-04
Deduplicate code in rpc::EVMExecutor and ExecutionProcessor


canepat:
For historical reasons, the simulated block and transaction execution necessary to implement many relevant JSON-RPC API endpoints (e.g. `eth_call`, `eth_estimateGas`, `trace_call`, `trace_block`...) is captured in [`rpc::EVMExecutor`](https://github.com/erigontech/silkworm/blob/master/silkworm/rpc/core/evm_executor.hpp#L89) class, which unfortunately  contains quite a lot of duplicated code wrt [`ExecutionProcessor`](https://github.com/erigontech/silkworm/blob/master/silkworm/core/execution/processor.hpp#L31) from `core` package.

We should investigate if it's feasible to merge these two abstraction into a single one.

See also https://github.com/erigontech/silkworm/pull/2284#discussion_r1742187707

PR #2354, opened 2024-09-16
core, rpcdaemon: move gas+funds bailout mechanism into evm

bzawisto:
Gas deduction and gas+funds validation are now part of evm (core).


canepat (approved):
LGTM

It would be probably wise to move also the gas refunding happening at the end of `EVMExecutor::call`, given that it is related to the gas bailout as well. You can do it in another PR, maybe after addressing #2304

PR #2362, opened 2024-09-18
infra: remove context pool wait modes

battlmonstr:
* remove wait_mode option because the default kBlocking mode (based on io_context.run() internal cond var) is good enough
* refactor context pools to make sure they are not empty from construction
* context pools have a constant size, make add_context protected
* rename num_contexts to size


silkworm/infra/grpc/server/server.hpp
```diff
@@ -161,6 +157,8 @@ class Server {
     //! The server configuration options.
     ServerSettings settings_;
 
+    grpc::ServerBuilder builder_;
```
canepat, line 160: AFAIK this is kind of an anti-pattern in usage of gRPC API: `::grpc::ServerBuilder` is a factory class, which should be used to build and start the server, then thrown away (as it was before in `build_and_start`) instead of keeping it alive for the entire server lifecycle. Moreover, this change makes the public `build_and_start` method no more idempotent, so the `Server` API becomes more dangerous.

I understand this change is driven by the change in `ServerContexPool`, where `add_context` is now private. TBH the only reason why such method was public before (i.e. adding contexts to the pool after creation) was exactly to implement `build_and_start` as it was, using the `::grpc::ServerBuilder` factory on-the-fly.

In my experience, the gRPC server setup and lifecycle handling is quite tricky, so I would prefer to avoid this specific change (the overall PR is fine, though).

canepat (changes requested):

silkworm/infra/grpc/server/server.hpp
battlmonstr: This is done, see the last commit.

canepat (approved):

PR #2368, opened 2024-09-19
db: move db setup from silkworm.cpp into Node and db

battlmonstr:
chain_data_init (old run_db_checklist):

* refactored to accept only required settings
* refactored to not modify node_settings, and return ChainConfig
* moved from cmd/common to db
* called from Node, then chaindata_env is created inside Node.

Future ideas: node_settings.chain_config should be separated from NodeSettings. NodeSettings could be split into 2 groups: DbSettings and ExecSettings.


canepat (approved):
Nitpicks:
- `run_db_checklist` -> `run_checklist`
- `DbChecklistSettings` -> `ChecklistSettings`
- `db_checklist.*` -> `checklist.*`

PR #2376, opened 2024-09-21
rpcdaemon: Fix cost case emvc error

lupin012:


lupin012:
wrong branch start point

PR #2381, opened 2024-09-29
integration: use shell scripts & enable workflow for E3

lupin012:


.github/workflows/rpc-integration-tests.yml
```diff
@@ -56,11 +60,13 @@ jobs:
...
         run: |
-          echo "Silkworm RpcDaemon starting..."
-          ./rpcdaemon --datadir $ERIGON_DATA_DIR --api admin,debug,eth,parity,erigon,trace,web3,txpool,ots,net --log.verbosity 1 --erigon_compatibility --jwt ./jwt.hex --skip_protocol_check --ws &
-          RPC_DAEMON_PID=$!
+          if [ ${{ matrix.backend }} == 'Erigon2' ]; then
+             ${{runner.workspace}}/silkworm/.github/workflows/e2_run_daemon.sh $ERIGON_DATA_DIR ./jwt.hex &
+          else
+             ${{runner.workspace}}/silkworm/.github/workflows/e3_run_daemon.sh $ERIGON_DATA_DIR ./jwt.hex &
+          fi
+          RPC_DAEMON_PID=$?
```
mriccobene, line 68: The directive $? captures the last command output, here we want the last command PID (directive $!).
Is not clear to me why you changed it.
Intact the workflows fails to stop rpcdaemon.

mriccobene (changes requested):

.github/workflows/rpc-integration-tests.yml
lupin012: The intention was:
- the script runs rpcdaemon in background. Capture the pid e return to the caller as exit code
- here the caller read the exit code of the script (the PID of rpcdaemon) and store to the RPC_DAEMON_PID variable
But this solution doesn't work.

If I use $! get the PID of the script and when kill it. The script dead but not the daemon.

Now I have modify again:
- here use $!  and kill the script
- the script launch daemon in background and wait its completion(wait). In case of signal from the caller script kill the daemon

lupin012:
This PR is substituted by PR #2400

PR #2385, opened 2024-09-30
docs: code style RFC amendments

battlmonstr:
Code style RFC amendments according to decisions made on the related workshop session.

### DIFF
--- a/docs/code_style.md
+++ b/docs/code_style.md
@@ -27,6 +27,7 @@ This is a list of project-specific guidelines that take precedence over the rule
 1. `using namespace foo` is allowed inside .cpp files, but not inside headers.
 1. User-defined literals are allowed.
 1. `template <Concept T>` syntax is allowed.
+1. Use `std::size_t` with `std::` prefix instead of `size_t`.  
 
 ### Libraries
 
@@ -139,18 +140,26 @@ An example where it is useful: if a function has a side effect, and returns no r
 
 Default to move by value. Use `T&&` parameters with caution.
 
-Legitimate uses are move-constructors. In some other cases it could be seen as a premature optimization. When copy-prevention is required, consider making a type move-only.
+Legitimate uses are move-constructors. In some other cases it could be seen as a premature optimization.
+
+When copy-prevention is required, consider making a type move-only. If T is an aggregate type (e.g. a basic struct), deleting the copy member functions defeats the aggregate semantics and forces to provide a custom constructor. In this case, if copy-prevention is critical for performance, it might be easier to use `T&&` than providing a custom constructor. 
 
 ### P10 constants definition keywords
+
+For class-level constants in .hpp files or constants in .cpp files:
 
 Use `static constexpr` if possible (inline is implicit):
 
 	static constexpr uint64_t kMinDifficulty{0x20000};
 
-Otherwise use `static const` if `inline` is not required by the compiler.
+Otherwise use `static const`.
 
+For global constants in .hpp files:
+
+Use `static constexpr` if possible (inline is implicit),  
 Otherwise use `inline static const`.
 
+See also: [Constants: Safe Idioms](https://abseil.io/tips/140)
 
 ### P11 CLion code inspections
 
--- a/docs/code_style.md
+++ b/docs/code_style.md
@@ -1,0 +1,226 @@
+# Silkworm code style
+
+The codebase respects:
+1. [C++ Core Guidelines][cpp-core-guidelines]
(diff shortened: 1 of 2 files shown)


docs/code_style.md
```diff
@@ -0,0 +1,226 @@
...
+
+### P9 move-only reference parameters
+
+Default to move by value. Use `T&&` parameters with caution.
+
+Legitimate uses are move-constructors. In some other cases it could be seen as a premature optimization. When copy-prevention is required, consider making a type move-only.
+
+### P10 constants definition keywords
+
+Use `static constexpr` if possible (inline is implicit):
```
yperbasis, line 146: Whether to use `inline` or not depends on whether the constant is in a header or in a .cpp file: https://abseil.io/tips/140

docs/code_style.md
```diff
@@ -0,0 +1,226 @@
...
+
+We've had a "modernize-use-nodiscard" clang-tidy policy, which led to the proliferation of `[[nodiscard]]` throughout the codebase. The policy was disabled, but a lot of usages were redundant. For example, usage on Task is redundant, because the type itself is already marked with `[[nodiscard]]` within the library.
+
+An example where it is useful: if a function has a side effect, and returns no result or an error (e.g. `bool ok = f()`).
+
+### P9 move-only reference parameters
+
+Default to move by value. Use `T&&` parameters with caution.
+
+Legitimate uses are move-constructors. In some other cases it could be seen as a premature optimization. When copy-prevention is required, consider making a type move-only.
```
canepat, line 142: IMO another legitimate usage is when `T` is an aggregate type. In fact, in such case the approach of forcing move by deleting the copy functions (i.e. copy-constructor and copy-assignment) defeats the aggregate semantics in the first place by asking you to provide a custom constructor.

docs/code_style.md
battlmonstr: updated formulation

docs/code_style.md
battlmonstr: updated formulation

docs/code_style.md
yperbasis: The doc says "inline is implicit", but it's not implicit for constants.

docs/code_style.md
```diff
@@ -0,0 +1,235 @@
...
+
+static constexpr size_t kThreadNameFixedSize = 11;
+
+Bad:
+
+static constexpr auto kThreadNameFixedSize = 11;
+
+### P18a optional for an empty std::function
+
+Sometimes we use `std::function` as a factory or a callback object. If it is optional prefer using `optional<function<T(U)>>` and `nullopt` where null is expected instead of using an empty `std::function` (with a `nullptr` inside).
```
yperbasis, line 225: Inconsistent with the recommendation for `unique_ptr` below.

yperbasis: To my mind it should be the opposite. `std::function` already supports the nil state by design, so I don't see a need to complicate things and use `optional<function<T(U)>>`.

docs/code_style.md
battlmonstr: @yperbasis are you saying that inline is not implicit for constexpr?

There's an inspection warning at least for class-level constants:

Screenshot 2024-10-01 at 15 00 50

docs/code_style.md
battlmonstr: Yes, it is inconsistent. It was a team decision on the meeting, after we've considered the alternative you suggest. The rationale for this difference is practical: for `std::function` the majority of usages (>95%) implies nonnull, and we want to be explicit about a few null cases. For `unique_ptr` it was not a clear cut (a lot of nullable cases), and we didn't want to introduce any extra noise, even though a distinction might be useful (as in gsl::not_null).

docs/code_style.md
yperbasis: It might be explicit for class members, but not for global constants. See "An Example Bug" in https://abseil.io/tips/140

battlmonstr: @yperbasis thanks for correction. I've understood the bug when I read, but I didn't realize it is solved by adding `inline`, because I misunderstood how `inline` works. I thought that it simply copy-pastes the constants everywhere (thus having even more copies than "static linkage"), while in fact it keeps only one definition (and memory address), and implies extern linkage. I've asked GPT, and it got it completely wrong (saying that inline is implied since C++17, which is false).

I've updated the text to my best knowledge, please check.

yperbasis (approved):

canepat (approved):

PR #2392, opened 2024-10-01
code style: use size_t without std:: prefix

battlmonstr:


yperbasis (commented):
This is contrary to Google's recommendation to omit `std::` for integer types: https://google.github.io/styleguide/cppguide.html#Integer_Types

battlmonstr:
> This is contrary to Google's recommendation to omit `std::` for integer types: https://google.github.io/styleguide/cppguide.html#Integer_Types

@yperbasis I've added a vote for you here: https://discord.com/channels/1133875232453693480/1133875438511476837/1290304675777609730

for me it doesn't matter which way we decide as long as we are consistent

yperbasis (approved):

PR #2396, opened 2024-10-02
seg: PatternAggregator kMaxPatternsSoftLimit

battlmonstr:
Increase the max number of patterns to keep in reserve while building the sorted list incrementally in order_by_score_and_limit.

Having more than kMaxPatterns patterns in consideration produces 10% better compression ratio while still keeping RAM usage under control.

see https://github.com/erigontech/erigon/pull/11672

### DIFF
--- a/silkworm/db/snapshots/seg/compressor/pattern_aggregator.cpp
+++ b/silkworm/db/snapshots/seg/compressor/pattern_aggregator.cpp
@@ -48,7 +48,7 @@ class PatternAggregatorImpl {
   public:
     explicit PatternAggregatorImpl(const filesystem::path& etl_work_path)
         : etl_work_path_(etl_work_path),
-          collector_(make_unique<db::etl::Collector>(etl_work_path, db::etl::kOptimalBufferSize / 2)) {}
+          collector_(make_unique<db::etl::Collector>(etl_work_path, db::etl::kOptimalBufferSize / 4)) {}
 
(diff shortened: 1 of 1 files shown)


canepat (approved):

silkworm/db/snapshots/seg/compressor/pattern_aggregator.cpp
```diff
@@ -57,6 +57,15 @@ class PatternAggregatorImpl {
...
 
   private:
+    /**
+     * The max number of patterns to keep in reserve
+     * while building the sorted list incrementally in order_by_score_and_limit.
+     * Having more than kMaxPatterns patterns in consideration
+     * produces 10% better compression ratio while still keeping RAM usage under control.
+     * see DictReducerSoftLimit in erigon
+     */
+    static constexpr size_t kMaxPatternsSoftLimit = 1'000'000;
```
JacekGlen, line 67: Can we have kMaxPatternsSoftLimit value referring to kMaxPatterns? E.g:
```
static constexpr size_t kMaxPatternsSoftLimit = kMaxPatterns * 16;
```

battlmonstr: this constant has to be identical to erigon's in order to produce bit-identical .seg files (having the same torrent hash to be seeded together with erigon 3 instances)

PR #2413, opened 2024-10-09
infra: log macros with a module name

battlmonstr:


silkworm/infra/common/log.hpp
```diff
@@ -137,15 +137,23 @@ using Message = LogBuffer<Level::kNone>;
...
-#define SILK_LOG SILK_LOGBUFFER(silkworm::log::Level::kNone)
+        silkworm::log::LogBuffer<level_>(__VA_ARGS__)
+
+#define SILK_TRACE2(...) SILK_LOGBUFFER(silkworm::log::Level::kTrace, __VA_ARGS__)
+#define SILK_DEBUG2(...) SILK_LOGBUFFER(silkworm::log::Level::kDebug, __VA_ARGS__)
+#define SILK_INFO2(...) SILK_LOGBUFFER(silkworm::log::Level::kInfo, __VA_ARGS__)
+#define SILK_WARN2(...) SILK_LOGBUFFER(silkworm::log::Level::kWarning, __VA_ARGS__)
+#define SILK_ERROR2(...) SILK_LOGBUFFER(silkworm::log::Level::kError, __VA_ARGS__)
+#define SILK_CRIT2(...) SILK_LOGBUFFER(silkworm::log::Level::kCritical, __VA_ARGS__)
+#define SILK_LOG2(...) SILK_LOGBUFFER(silkworm::log::Level::kNone, __VA_ARGS__)
```
canepat, line 151: I understand your intent is keeping the names as short as possible, nevertheless I would prefer them to be slightly more evocative. Something like:
```
SILK_TRACE_M
SILK_DEBUG_M
...
```
where `M` stands for Module or
```
SILK_TRACE_VA
SILK_DEBUG_VA
...
```
where `VA` stands for VarArgs

battlmonstr: X_M is OK. I hope that in the future we convert to X_M almost everywhere, and rename X_M back to X without suffix (for a few cases without arguments we could use `X_M()` with no params). Another idea could be to try using a constexpr function instead of macros.

canepat (approved):

PR #2423, opened 2024-10-15
snapshots: improve verbose dump in count subcommands

canepat:


cmd/dev/snapshots.cpp
```diff
@@ -252,11 +260,26 @@ void parse_command_line(int argc, char* argv[], CLI::App& app, SnapshotToolboxSe
...
+auto as_milliseconds(const std::chrono::duration<R, P>& elapsed) {
+    return duration_as<std::chrono::milliseconds>(elapsed);
+}
+
+//! Convert the given duration into seconds
+//! \param elapsed the duration to convert
+template <typename R, typename P>
+auto as_seconds(const std::chrono::duration<R, P>& elapsed) {
+    return duration_as<std::chrono::seconds>(elapsed);
+}
```
battlmonstr, line 281: After watching a [talk on chrono](https://www.youtube.com/watch?v=P32hvk8b13M) I believe that this is not a proper usage of duration_cast. A better approach is to create a `string format_duration(duration)`. I suggest to format it to seconds with ms fractional part e.g.: `12.455 sec`

battlmonstr (approved):

PR #2429, opened 2024-10-17
db: set operations on paginated Data API iterators

canepat:
Add support for intersection/union set operations on E3 Data API paginated iterators and change the paginated iterator API exposing `has_next`/`next` stream-like methods, notably:

```c++
PaginatedUint64& paginated = ...;
auto it = co_await paginated.begin();
while (const auto value = co_await it.next()) {
    // use *value;
}
```

Furthermore, conceptualise and improve `PaginatedIterator` abstraction and increase test coverage.

*Extra*
- cleanup `RemoteTransactionTest`


battlmonstr (approved):
:+1:

Issue #2318, opened 2024-09-06
upgrade to Conan 2


battlmonstr:
```
WARN: *** Conan 1 is legacy and on a deprecation path ***
WARN: *** Please upgrade to Conan 2 ***
```

chfast:
I will wait for this on evmone side.

PR #2432, opened 2024-10-19
cmake: upgrade to Conan 2 (#2318)

battlmonstr:
extras:
* refactor compiler_install.sh
* reduce verbosity of cmake configuration output

"light" workflow jobs build time of conan (cmake configuration command):
```
gcc (linux-gcc-12-release) - no conan cache
- conan 1 - 2,6 min
- conan 2 - 2,5 min
clang (linux-clang-16-release) - no conan cache
- conan 1 - 3 min
- conan 2 - 13 min

gcc (linux-gcc-12-release) - with conan cache
- conan 1 - 0,3 min
- conan 2 - 0,23 min
clang (linux-clang-16-release) - with conan cache
- conan 1 - 0,25  min
- conan 2 - 0,22  min
```
conan 2 has no clang binaries for Linux, so the first build on clang has to build everything from source. That takes 10 min longer compared to conan 1 where we build only 2-3 libs from source. It is cached and then the build time is about equal.


battlmonstr:
superseded by https://github.com/erigontech/silkworm/pull/2607

PR #2435, opened 2024-10-21
third_party: update erigon-interfaces to commit cd6dd2549

lupin012:


silkworm/rpc/ethbackend/backend.hpp
```diff
@@ -45,6 +45,7 @@ class BackEnd {
     virtual Task<BlockNum> get_block_number_from_txn_hash(const HashAsSpan& hash) = 0;
     virtual Task<BlockNum> get_block_number_from_hash(const HashAsSpan& hash) = 0;
     virtual Task<evmc::bytes32> get_block_hash_from_block_number(uint64_t number) = 0;
+    virtual Task<std::string> canonical_body_for_storage(uint64_t number) = 0;
```
canepat, line 48: 1. `uint64_t number` -> `BlockNum block_number`
2. `Task<std::string>` -> `Task<Bytes>`

silkworm/rpc/ethbackend/remote_backend.cpp
```diff
@@ -202,6 +202,17 @@ Task<evmc::bytes32> RemoteBackEnd::get_block_hash_from_block_number(uint64_t num
     co_return hash;
 }
 
+Task<std::string> RemoteBackEnd::canonical_body_for_storage(uint64_t number) {
+    const auto start_time = clock_time::now();
+    UnaryRpc<&::remote::ETHBACKEND::StubInterface::AsyncCanonicalBodyForStorage> canonical_body_for_storage_rpc{*stub_, grpc_context_};
+    ::remote::CanonicalBodyForStorageRequest request;
+    request.set_blocknumber(number);
+    const auto reply = co_await canonical_body_for_storage_rpc.finish_on(executor_, request);
+    SILK_TRACE << "RemoteBackEnd::canonical_body_for_storage bn="
```
canepat, line 211: Add missing `<< block_number` at the end of this line

silkworm/rpc/ethbackend/remote_backend.cpp
```diff
@@ -202,6 +202,17 @@ Task<evmc::bytes32> RemoteBackEnd::get_block_hash_from_block_number(uint64_t num
...
+               << " t=" << clock_time::since(start_time);
+    co_return reply.body();
```
canepat, line 213: After changing return type to `Task<Bytes>`, you should also check which data the `body` string is carrying on and handle the decoding here (as opposed to putting such burden on each call site).

canepat (changes requested):

silkworm/rpc/ethbackend/backend.hpp
lupin012: ok

silkworm/rpc/ethbackend/remote_backend.cpp
lupin012: ok

silkworm/rpc/ethbackend/remote_backend.cpp
lupin012: I don't understand if the return types remains Bytes or it is necessary calls decode_stored_block_body() and so the returns types becomes Task<BlockBodyForStorage>

canepat: You're right, it's better to change the return type to `BlockBodyForStorage`

canepat (approved):

PR #2439, opened 2024-10-22
db: reuse ROTxn reference in LocalState

canepat:


canepat:
This is not going to work due to MDBX transaction lifetime issues when using `LocalState` in worker threads.

PR #2454, opened 2024-10-25
rpcdaemon: fix mismatch in erigon_getHeaderBy[Hash|Number]

lupin012:


canepat (approved):

silkworm/rpc/commands/eth_api.cpp
```diff
@@ -957,7 +957,7 @@ Task<void> EthereumRpcApi::handle_eth_get_balance(const nlohmann::json& request,
         reply = make_json_content(request, "0x" + (account ? intx::hex(account->balance) : "0"));
     } catch (const std::exception& e) {
         SILK_ERROR << "exception: " << e.what() << " processing request: " << request.dump();
-        reply = make_json_error(request, kInternalError, e.what());
+        reply = make_json_content(request, "0x0");
```
canepat, line 960: This should be investigated more: it's far from ideal to return just zero balance if any exception happens on the node

PR #2476, opened 2024-11-04
Implement eip 7685 wiring with three types of request

bzawisto:
This PR introduces the implementation of four Ethereum Improvement Proposals (EIPs) including:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7685.md
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6110.md
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7002.md
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7251.md

Please not that these EIPs are subject to change in the future and the implementation might be not final yet.
Worth noting is that EEST are not included in this PR and will be shipped in upcoming requests.

Current implementation is up to date with Erigon.

### DIFF
--- a/.github/workflows/rpc-integration-tests.yml
+++ b/.github/workflows/rpc-integration-tests.yml
@@ -35,7 +35,7 @@ jobs:
           rm -rf ${{runner.workspace}}/rpc-tests
           git -c advice.detachedHead=false clone --depth 1 --branch v1.14.0 https://github.com/erigontech/rpc-tests ${{runner.workspace}}/rpc-tests
           cd ${{runner.workspace}}/rpc-tests
-          pip3 install -r requirements.txt
+          pip3 install -r requirements.txt --break-system-packages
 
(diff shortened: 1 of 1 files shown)


silkworm/execution/grpc/common/block.cpp
```diff
@@ -59,6 +59,9 @@ void header_from_proto(const ::execution::Header& proto_header, BlockHeader& hea
     if (proto_header.has_excess_blob_gas()) {
         header.excess_blob_gas = proto_header.excess_blob_gas();
     }
+    if (proto_header.has_requests_hash()) {
+        header.receipts_root = rpc::bytes32_from_h256(proto_header.requests_hash());
```
yperbasis, line 63: Should be requests_hash

canepat (changes requested):

PR #2478, opened 2024-11-05
rpcdaemon: fix canonical body for storage provider not found key

lupin012:
THIS PR fixes problem on ots_hasCode() and eth_estimateGas()

### DIFF
--- a/silkworm/db/kv/txn_num.cpp
+++ b/silkworm/db/kv/txn_num.cpp
@@ -127,9 +127,11 @@ Task<std::optional<BlockNum>> block_num_from_tx_num(kv::api::Transaction& tx,
     const auto block_num = co_await async_binary_search(last_block_num + 1, [&](size_t i) -> Task<bool> {
         const auto max_tx_num = co_await last_tx_num_for_block(max_tx_num_cursor, i, provider);
         if (!max_tx_num) {
-            auto first_key = co_await max_tx_num_cursor->first();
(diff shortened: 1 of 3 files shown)


silkworm/db/kv/txn_num.cpp
```diff
@@ -35,12 +35,19 @@ static Task<TxNum> last_tx_num_for_block(const std::shared_ptr<kv::api::Cursor>&
...
         if (!block_body_data) {
-            throw std::invalid_argument("Bad block_num  " + std::to_string(block_number));
+            key_value = co_await max_tx_num_cursor->last();
+            if (key_value.value.empty()) {
+                co_return 0;
+            }
+            if (key_value.value.size() != sizeof(TxNum)) {
+                throw std::length_error("Bad TxNum value size " + std::to_string(key_value.value.size()) + " in db");
+            }
+            co_return endian::load_big_u64(key_value.value.data());
```
canepat, line 50: Returning the max transaction number associated to the last block when the target block cannot be found neither in db nor in snapshots is OK for `min_tx_num` and `max_tx_num`, but not for `block_num_from_tx_num`.
I would suggest to change the return type to `Task<std::optional<TxNum>>` and use `std::nullopt` here, relying on `last` call outside when needed.

canepat (changes requested):

canepat (approved):

PR #2480, opened 2024-11-05
Use evmone APIv2 - initial integration

chfast:
Integrate evmone's higher level API for transaction execution.

The new API moves the level of granularity of EVM execution from internal calls to whole transaction. This hides a lot of EVM details from the user (Silkworm), e.g. transient storage, internal calls logic, worm/cold accesses, state revert journal.

### evmone API

The API itself is "work in progress", but the goal is to improve it on both sides going into the future. The current best indicator how well it is defined are the additional evmone files added to the compilation of Silkworm. We also use the side branch with some shortcoming on top of the evmone's main branch. The plan is to integrate these changes to evmone over time. Therefore, the length of the branch is good indicator how much more is left to be done.

- [ ] https://github.com/ethereum/evmone/pull/1069
- [ ] https://github.com/ethereum/evmone/issues/1070

### Performance

I did a performance test of historical block execution up to block 15M on Mainnet. This was used to motivate changes in evmone by eliminating obvious overhead (e.g. transaction should not be validated in evmone, evmone should not compute log's bloom filters). The final result was the sync time was the same (within 1%).

### DIFF
--- a/cmd/state-transition/state_transition.cpp
+++ b/cmd/state-transition/state_transition.cpp
@@ -273,7 +273,6 @@ void cleanup_error_block(Block& block, ExecutionProcessor& processor, const evmc
         processor.evm().state().access_account(block.header.beneficiary);
     }
     processor.evm().state().add_to_balance(block.header.beneficiary, 0);
-    processor.evm().state().finalize_transaction(rev);
     processor.evm().state().write_to_db(block.header.number);
 }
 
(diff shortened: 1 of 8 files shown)


canepat:
Some unit tests are failing, namely:

- `silkworm_core_test`
  - [ ] Create contract, destruct and then recreate
  - [ ] Destruct and recreate
  - [ ] Smart contract creation w/ insufficient gas
  - [ ] State changes for creation+destruction of smart contract
  - [ ] Tracing smart contract with storage

- `silkworm_node_test`
  - [ ] Sync Stages :: Execution and CallTraceIndex

PR #2482, opened 2024-11-05
bor.proto codegen target fix

battlmonstr:


chfast (approved):
Fixed the ninja build issue.

PR #2489, opened 2024-11-06
Pectra EIP 7702: set eoa account code for one transaction

bzawisto:
Last EIP from Pectra. It's not fully implemented and, depending on the outcome of evm v2 API, the remaining bit is either integrating with new API or, more complicated, implementing required logic directly in silkworm.


silkworm/core/types/transaction_test.cpp
```diff
@@ -234,4 +234,88 @@ TEST_CASE("Recover sender 2") {
...
+        "f20194327d773bdc6c27cd28a533e81074372dc33a8afd884ef63dce09c5e56c8088cb702ac89cff7"
+        "65f88d26fe11c3d471949f20194f61ffc773a97207c8124c29526a59e6fa0b34a52880e563a787da9"
+        "52ab808884f2a19b171abfb2882d473907f3ada086f20194c1d608bb39e078a99086e7564e89a7625"
+        "ed86dca88e8a0ab45821912e88088df6c3d43080350518895a828c35680a0278088e2487fd89ca40b"
+        "3488689accdbeb8d4d2e");
+
+    Transaction decoded;
+    ByteView view{encoded_rlp};
+    const auto status = rlp::decode_transaction(view, decoded, rlp::Eip2718Wrapping::kBoth,
+                                                rlp::Leftover::kAllow);
```
yperbasis, line 270: better use `Leftover::kProhibit`

silkworm/core/protocol/validation.cpp
```diff
@@ -187,6 +199,12 @@ ValidationResult validate_call_precheck(const Transaction& txn, const EVM& evm)
         }
     }
 
+    if (evm.revision() >= EVMC_PRAGUE) {
+        if (txn.type == TransactionType::kSetCode && !txn.to) {
```
yperbasis, line 203: Hmm. I'm not sure what's the difference between `validate_call_precheck` & `pre_validate_transaction` is (looks like we should eliminate code duplication between the two methods), but it looks suspicious to me that EIP-7702 transactions are validated differently between the two.

canepat: `pre_validate_transaction` and `validate_call_precheck` are slightly different because the checks done also by Erigon before call simulations (e.g. `eth_call`) are somewhat more relaxed.
Anyway, validation of EIP-7702 transactions seems to be equivalent except for the missing check on Prague revision in `pre_validate_transaction`, which should be added.

I agree we should eliminate duplication between the two methods (see #2522), doing it in separate PR is fine or even better.

yperbasis (approved):

PR #2509, opened 2024-11-14
snapshots: more state schema configurations

battlmonstr:
* .kvei and .bt schema (only schema for now, TODOs added for opening files)
* configure KVSegment compression kind: uncompressed by default, Storage & Commitment .kv has keys compressed, Code .kv has values compressed (.ef is always uncompressed)
* configure Segment compression: history is uncompressed by default, compressed for Code .v
* disable Commitment history

### DIFF
--- a/silkworm/db/datastore/snapshots/snapshot_repository.hpp
+++ b/silkworm/db/datastore/snapshots/snapshot_repository.hpp
@@ -107,7 +107,7 @@ class SnapshotRepository {
     }
 
     std::pair<std::optional<SegmentAndIndex>, std::shared_ptr<SnapshotBundle>> find_segment(
-        std::pair<datastore::EntityName, datastore::EntityName> names,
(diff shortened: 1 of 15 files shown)


silkworm/db/datastore/snapshots/snapshot_repository.hpp
```diff
@@ -107,7 +107,7 @@ class SnapshotRepository {
     }
 
     std::pair<std::optional<SegmentAndIndex>, std::shared_ptr<SnapshotBundle>> find_segment(
-        std::pair<datastore::EntityName, datastore::EntityName> names,
+        std::array<datastore::EntityName, 3> names,
```
canepat, line 110: It would be better to insert a `using` declaration for `std::array<datastore::EntityName, 3>`, as it is used in 4 different places

canepat (approved):

PR #2514, opened 2024-11-15
rpcdaemon: debug_storageRangeAt for e3

lupin012:
On CI it is necessary that erigon is at commit cf3eddc63c5bf2fdd4b28e9a563f174c1d07d243
This PR is connected to https://github.com/erigontech/rpc-tests/pull/291

### DIFF
--- a/silkworm/rpc/core/storage_walker.cpp
+++ b/silkworm/rpc/core/storage_walker.cpp
@@ -212,7 +212,7 @@ Task<void> StorageWalker::storage_range_at(
         const auto key = value->first.substr(20);
         auto hash = hash_of(key);
         const auto sec_key = ByteView{hash.bytes};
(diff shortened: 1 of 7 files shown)


silkworm/rpc/core/storage_walker.cpp
```diff
@@ -177,46 +177,43 @@ Task<void> StorageWalker::walk_of_storages(
...
 
+static void increment_key(Bytes& array) {
+    for (auto& it : std::ranges::reverse_view(array)) {
+        if (it < 0xFF) {
+            ++it;
+            break;
+        }
+        it = 0x00;
+    }
+}
```
canepat, line 188: This exact function is present also in `parity_api.cpp`, we should move it to `rpc/common` folder. You can do this in next PR

canepat (approved):

PR #2528, opened 2024-11-22
ci: fix clang tidy after PRs 2518 & 2476

canepat:


silkworm/infra/common/environment.cpp
```diff
@@ -84,4 +84,10 @@ void Environment::set_pre_verified_hashes_disabled() {
     environment["DISABLE_PRE_VERIFIED_HASHES"] = "1";
 }
 
+std::string Environment::get(const std::string& var_name) {
+    auto environment = boost::this_process::environment();
```
battlmonstr, line 88: what does boost give us here compared to getenv/setenv?

silkworm/db/datastore/snapshots/btree/btree_index.cpp
```diff
@@ -76,8 +76,8 @@ std::optional<BTreeIndex::Cursor> BTreeIndex::seek(ByteView seek_key, const KVSe
     if (key.compare(seek_key) >= 0) {
         return BTreeIndex::Cursor{
             this,
-            std::move(key),
-            std::move(value),
+            key,
+            value,
```
battlmonstr, line 80: the move is intentional here

please update the Cursor constructor to take Bytes instead (and move into Cursor::key_/Cursor::value_)

battlmonstr (changes requested):

silkworm/infra/common/environment.cpp
canepat: > what does boost give us here compared to getenv/setenv? (just curious, no need to change the code)

At least no warning about `std::getenv` being thread-unsafe 😃

PR #2546, opened 2024-11-29
rpcdaemon: Move blocks api on chain storage

lupin012:


lupin012:
substitued by PR #2550

PR #2569, opened 2024-12-08
rpcdaemon: use transaction-level state in debug API

lupin012:
This PR introduces transaction-level state for the following API:
 - `debug_trace_transaction`
 - `debug_traceCall`
 - `debug_traceCallMany`

### DIFF
--- a/silkworm/rpc/core/evm_debug.cpp
+++ b/silkworm/rpc/core/evm_debug.cpp
@@ -378,11 +378,9 @@ Task<void> DebugExecutor::trace_call(json::Stream& stream, const BlockNumOrHash&
     const auto& block = block_with_hash->block;
     const auto block_num = block.header.number;
 
-    auto transaction_index = static_cast<std::int32_t>(block_with_hash->block.transactions.size());
-
     stream.write_field("result");
     stream.open_object();
-    co_await execute(stream, storage, block_num, block, transaction, transaction_index);
+    co_await execute(stream, storage, block_num, block, transaction, /* index */0);
     stream.close_object();
 
     co_return;
@@ -473,6 +471,7 @@ Task<void> DebugExecutor::execute(json::Stream& stream, const ChainStorage& stor
     co_return;
 }
 
+// used by unit-test
 Task<void> DebugExecutor::execute(json::Stream& stream, const ChainStorage& storage, const silkworm::Block& block, const Call& call) {
     rpc::Transaction transaction{call.to_transaction()};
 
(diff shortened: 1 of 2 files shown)


silkworm/rpc/core/evm_debug.cpp
```diff
@@ -378,9 +378,11 @@ Task<void> DebugExecutor::trace_call(json::Stream& stream, const BlockNumOrHash&
     const auto& block = block_with_hash->block;
     const auto block_num = block.header.number;
 
+    auto transaction_index = static_cast<std::int32_t>(block_with_hash->block.transactions.size());
```
canepat, line 381: Use `gsl::narrow<int32_t>` instead of `static_cast<std::int32_t>`

silkworm/rpc/core/evm_debug.cpp
```diff
@@ -473,7 +475,10 @@ Task<void> DebugExecutor::execute(json::Stream& stream, const ChainStorage& stor
 
 Task<void> DebugExecutor::execute(json::Stream& stream, const ChainStorage& storage, const silkworm::Block& block, const Call& call) {
     rpc::Transaction transaction{call.to_transaction()};
-    co_await execute(stream, storage, block.header.number, block, transaction, -1);
+
+    auto transaction_index = static_cast<std::int32_t>(block.transactions.size());
```
canepat, line 479: Same as above

silkworm/rpc/core/evm_debug_test.cpp
```diff
@@ -173,6 +173,9 @@ TEST_CASE_METHOD(DebugExecutorTest, "DebugExecutor::execute call 1") {
             .WillOnce(InvokeWithoutArgs([]() -> Task<std::optional<evmc::bytes32>> {
                 co_return kZeroHeaderHash;
             }));
+        EXPECT_CALL(transaction, first_txn_num_in_block(_)).WillOnce(Invoke([]() -> Task<TxnId> {
```
canepat, line 176: Specify the number instead of using `_`, if possible

silkworm/rpc/core/evm_debug.cpp
```diff
@@ -543,15 +549,16 @@ Task<void> DebugExecutor::execute(
 
     const auto chain_config = co_await storage.read_chain_config();
     auto current_executor = co_await boost::asio::this_coro::executor;
+
+    // We must do the execution at the state after the txn identified by transaction_with_block param in the same block
+    // at the state of the block identified by the given block_num, i.e. at the start of the block (block_num)
+    const auto first_txn_num_in_block = co_await tx_.first_txn_num_in_block(block.header.number + 1);
```
canepat, line 555: In recent PRs targeting trace/debug API end-points, there are code portions like this one regarding the transaction-level state access which keep returning again and again. When finished, we should strive for some refactoring, at least give it a try.

Nothing to do here specifically, just a general comment 😄

canepat (changes requested):

silkworm/rpc/core/evm_debug.cpp
lupin012: ok

silkworm/rpc/core/evm_debug.cpp
lupin012: ok

PR #2580, opened 2024-12-10
node: restore account code hash in HashState unwind

canepat:
We need to restore the code hash into the account during the HashState unwind procedure when we encounter an account having non-zero incarnation and empty code hash.

*Extras*
- improve error logs
- avoid some CLion warnings


battlmonstr (approved):
👼

PR #2589, opened 2024-12-13
evm: add flag to execute in evmone APIv2

chfast:
Add a flag to ExecutionProcesser to indicate the transaction execution should happen in evmone APIv2 only. This flag is off by default and setting it to on has no effect for now.

### DIFF
--- a/cmd/dev/check_changes.cpp
+++ b/cmd/dev/check_changes.cpp
@@ -131,7 +131,7 @@ int main(int argc, char* argv[]) {
             db::Buffer buffer{txn, std::make_unique<db::BufferFullDataModel>(access_layer)};
             buffer.set_historical_block(block_num);
 
-            ExecutionProcessor processor{block, *rule_set, buffer, *chain_config};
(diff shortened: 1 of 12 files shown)


chfast:
I know you are using style `/*evm1_v2=*/false`. Should I apply it or is it fine considering this is temporary?

chfast:
We can also set to `true` by default so enabling evmone APIv2 later will mean removing the `, false`.

canepat:
> I know you are using style `/*evm1_v2=*/false`. Should I apply it or is it fine considering this is temporary?

It is fine as you did considering this flag is temporary.

canepat:
> We can also set to `true` by default so enabling evmone APIv2 later will mean removing the `, false`.

IMO it's OK not to have a default value for the flag: this way makes it self-evident in each usage of `ExecutionProcessor` if we're using APIv1 or APIv2 and then we will remove this flag in any case.

canepat (approved):

PR #2601, opened 2024-12-14
rpcdaemon: Use state on tx on bundles

lupin012:


lupin012:
This PR is substituted by #2602

PR #2605, opened 2024-12-16
node: fix task caller lifecycle in Triggers schedule

### LINKED ISSUE #2604: silkworm: assertion failed in Triggers stage
Running standalone Silkworm on mainnet, an assertion failure occurred in `TriggersStage` during the first cycle after restart:

[sw_tip_triggers_stage_assertion_failed.log](https://github.com/user-attachments/files/18142785/sw_tip_triggers_stage_assertion_failed.log)

but it was not always reproducible at startup. After adding some log traces and repeating the execution, an unexpected termination of Silkworm occurred with the logs shown below:

[sw_tip_triggers_stage_unexpected_exit_startup_cycle.log](https://github.com/user-attachments/files/18142787/sw_tip_triggers_stage_unexpected_exit_startup_cycle.log)

canepat:
Fixes #2604 

The problem was hidden behind the lifecycle of the trigger task caller: its coroutine frame could be destroyed before being spawned.

### DIFF
--- a/silkworm/node/stagedsync/stages/stage_triggers.cpp
+++ b/silkworm/node/stagedsync/stages/stage_triggers.cpp
@@ -40,13 +40,13 @@ Stage::Result TriggersStage::forward(db::RWTxn& tx) {
 }
 
 Task<void> TriggersStage::schedule(std::function<void(db::RWTxn&)> callback) {
-    auto task_caller = [this, trigger = std::move(callback)]() -> Task<void> {
-        db::RWTxn* tx = this->current_tx_;
+    auto task_caller = [](auto* self, auto trigger) -> Task<void> {
+        db::RWTxn* tx = self->current_tx_;
         SILKWORM_ASSERT(tx);
         trigger(*tx);
         co_return;
     };
-    co_return co_await concurrency::spawn_task(ioc_, task_caller());
+    return concurrency::spawn_task(ioc_, task_caller(this, std::move(callback)));
 }
 
 bool TriggersStage::stop() {
--- a/silkworm/node/stagedsync/stages/stage_triggers.cpp
+++ b/silkworm/node/stagedsync/stages/stage_triggers.cpp
@@ -40,13 +40,13 @@ Stage::Result TriggersStage::forward(db::RWTxn& tx) {
 }
 
 Task<void> TriggersStage::schedule(std::function<void(db::RWTxn&)> callback) {
-    auto task_caller = [this, c = std::move(callback)]() -> Task<void> {
+    auto task_caller = [this, trigger = std::move(callback)]() -> Task<void> {
         db::RWTxn* tx = this->current_tx_;
         SILKWORM_ASSERT(tx);
-        c(*tx);
+        trigger(*tx);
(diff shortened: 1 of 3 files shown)


silkworm/node/stagedsync/stages/stage_triggers.cpp
```diff
@@ -40,13 +40,13 @@ Stage::Result TriggersStage::forward(db::RWTxn& tx) {
...
-    auto task_caller = [this, c = std::move(callback)]() -> Task<void> {
+    auto task_caller = [this, trigger = std::move(callback)]() -> Task<void> {
         db::RWTxn* tx = this->current_tx_;
         SILKWORM_ASSERT(tx);
-        c(*tx);
+        trigger(*tx);
         co_return;
     };
-    return concurrency::spawn_task(ioc_, task_caller());
+    co_return co_await concurrency::spawn_task(ioc_, task_caller());
```
battlmonstr, line 49: This is an interesting side effect. I wonder why this change makes any difference?

The call to schedule is already awaited [here on the call site](https://github.com/erigontech/silkworm/blob/master/silkworm/db/freezer.cpp#L164)

An additional await here delays the destruction of some locals, right? Which ones? `task_caller` is not relevant after calling it. `trigger` is moved into the task, so not destroyed unless the spawned task is destroyed. The spawned task is moved inside co_spawn. I hope that co_spawn is not destroying the passed task prematurely, but maybe it does? Is it possible to verify with a unit test?

If it really does, maybe we modify spawn_task to await for co_spawn inside spawn_task so that this mistake is not possible in other places (we use spawn_task without awaiting in a bunch of places) and in the future code?

I'd like to fill the gap in my understanding here. I know it is not the high priority, so let's maybe create an issue about researching and generalizing this and make it an exercise for the future reader?

battlmonstr (approved):

silkworm/node/stagedsync/stages/stage_triggers.cpp
battlmonstr: This could potentially be a reason of some sentry discovery crashes, because SerialNodeDb follows this pattern.

battlmonstr:
@canepat 
https://github.com/chriskohlhoff/asio/issues/1491
https://devblogs.microsoft.com/oldnewthing/20211103-00/?p=105870
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rcoro-capture

silkworm/node/stagedsync/stages/stage_triggers.cpp
battlmonstr: let's try to move the callback from the capture list to the lambda parameter list and see if it works without co_await (see links below)

canepat: Yep, we must move both `this` and the callback because all the capture list is affected (indeed, the issue was happening on `this->current_tx_`).
Moreover, it seems to be reproducible systematically in unit tests, so I'm going to add one test and fix by moving from capture list to the lambda parameters.

PR #2634, opened 2025-01-03
Implement Pectra Eip-7623

bzawisto:
See: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7623.md
This has become a part of the devnet-5.

### DIFF
--- a/silkworm/core/protocol/validation.cpp
+++ b/silkworm/core/protocol/validation.cpp
@@ -167,9 +167,6 @@ ValidationResult pre_validate_common_base(const Transaction& txn, evmc_revision 
         return ValidationResult::kIntrinsicGas;
     }
 
-    if (revision >= EVMC_PRAGUE) {
-    }
-
     if (intx::count_significant_bytes(txn.maximum_gas_cost()) > 32) {
         return ValidationResult::kInsufficientFunds;
     }


silkworm/core/protocol/intrinsic_gas.cpp
```diff
@@ -30,6 +30,26 @@ intx::uint128 intrinsic_gas(const UnsignedTransaction& txn, const evmc_revision
...
+
+        // EIP-3860: Limit and meter initcode
+        if (contract_creation && rev >= EVMC_SHANGHAI) {
+            gas += num_words(data_len) * fee::kInitCodeWordCost;
+        }
+
+        if (rev >= EVMC_PRAGUE) {
+            // EIP-7623: Increase calldata cost
+            const intx::uint128 floor_cost = fee::kGTransaction + (zero_bytes + non_zero_bytes * 4) * fee::kTotalCostFloorPerToken;
+            gas = std::max(gas, floor_cost);
```
chfast, line 49: This is not exactly correct. The floor cost is not a part of intrinsic cost. However, the transaction gas limit must be big enough to cover the floor cost just in case.

After the transaction execution you need another modification: if the gas used by the transaction is lower than the floor cost, bump it to the floor cost.

chfast (commented):
The pectra-devnet-5 tests has a bug where they apply the final floor cost adjustment before applying gas refunds while it should be done before.

PR #2641, opened 2025-01-08
rpcdaemon: fix and improve perf in trace_filter

Sixtysixter:
trace_filter performance for long range of not existent block numbers

### DIFF
--- a/silkworm/rpc/core/evm_trace.cpp
+++ b/silkworm/rpc/core/evm_trace.cpp
@@ -1759,12 +1759,8 @@ Task<void> TraceCallExecutor::trace_filter(const TraceFilter& trace_filter, cons
     filter.count = trace_filter.count;
 
     auto block_num = trace_filter.from_block.number();
-    std::shared_ptr<BlockWithHash> block_with_hash;
-    try {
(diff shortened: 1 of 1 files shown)


silkworm/rpc/core/evm_trace.cpp
```diff
@@ -1765,26 +1766,34 @@ Task<void> TraceCallExecutor::trace_filter(const TraceFilter& trace_filter, cons
     } catch (const std::invalid_argument& iv) {
         block_with_hash = nullptr;
     }
+    rpc::BlockReader block_reader{chain_storage_, tx_};
+    auto latest_block_number = co_await block_reader.get_latest_block_num();
```
canepat, line 1770: The `get_latest_block_num` call seems redundant to me because the latest block is a moving target and you don't need it for this API implementation: you can just iterate on the provided block range one-by-one and stop at the first block not found.
I'd also suggest to double-check the not-found condition for `core:: read_block_by_block_num_or_hash` and `core::read_block_by_number`.

canepat (changes requested):

silkworm/rpc/core/evm_trace.cpp
Sixtysixter: OK

canepat (approved):

PR #2648, opened 2025-01-12
db, node, sentry, sync: more exception handlers

canepat:


silkworm/db/snapshot_sync.cpp
```diff
@@ -96,7 +96,15 @@ Task<void> SnapshotSync::run() {
         co_return;
     }
 
-    co_await (setup_and_run() && client_.async_run("bit-torrent"));
+    try {
+        co_await (setup_and_run() && client_.async_run("bit-torrent"));
+    } catch (const boost::system::system_error& ex) {
+        SILK_WARN_M("SnapshotSync") << "SnapshotSync::run ex=" << ex.what();
+        if (ex.code() == boost::system::errc::operation_canceled) {
+            SILK_WARN_M("SnapshotSync") << "SnapshotSync::run operation_canceled";
```
battlmonstr, line 104:
```suggestion
        SILK_ERROR_M("SnapshotSync") << "SnapshotSync::run ex=" << ex.what();
        if (ex.code() == boost::system::errc::operation_canceled) {
            SILK_DEBUG_M("SnapshotSync") << "SnapshotSync::run operation_canceled";
```
operation_canceled could be a normal expected graceful exit, and error is an unrecoverable error in this case

P.S. if we implement https://github.com/erigontech/silkworm/issues/1527 this extra code could be removed

battlmonstr (changes requested):

silkworm/db/snapshot_sync.cpp
battlmonstr: @canepat can't you enable DEBUG level while debugging?

canepat: > @canepat can't you enable DEBUG level while debugging?

not in this case, because as you can see the exception handlers are in different modules (db, node, sentry, sync) and that would enable too many logs

battlmonstr (approved):

PR #2652, opened 2025-01-13
evm: drop bailout option from validate_call_funds()

chfast:
The bailout option to validate_call_funds() makes it no-op function. Drop this option from the function so that the transaction validation code is not aware of the bailout semantic. Instead, ignore the validation result in the single place where it matters.

### DIFF
--- a/silkworm/rpc/core/evm_executor.cpp
+++ b/silkworm/rpc/core/evm_executor.cpp
@@ -240,8 +240,8 @@ ExecutionResult EVMExecutor::call(
 
     const auto owned_funds = execution_processor_.intra_block_state().get_balance(*txn.sender());
 
-    if (const auto result = protocol::validate_call_funds(txn, evm, owned_funds, bailout);
-        result != ValidationResult::kOk) {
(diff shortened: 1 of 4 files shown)


yperbasis (approved):

silkworm/rpc/core/evm_executor.cpp
```diff
@@ -240,8 +240,8 @@ ExecutionResult EVMExecutor::call(
 
     const auto owned_funds = execution_processor_.intra_block_state().get_balance(*txn.sender());
 
-    if (const auto result = protocol::validate_call_funds(txn, evm, owned_funds, bailout);
-        result != ValidationResult::kOk) {
+    if (const auto result = protocol::validate_call_funds(txn, evm, owned_funds);
+        !bailout && result != ValidationResult::kOk) {
```
chfast, line 244: The application is a big awkward because we call the function even if we don't need the result. But considering the bailout is not enabled often this somehow guarantees the `validate_call_funds` still work correctly without crashes or assert failures.

PR #2656, opened 2025-01-15
datastore: DomainGetLatestQuery for snapshots and datastore

battlmonstr:
DomainGetLatestQuery was implemented only for kvdb before.

This implements DomainGetLatestQuery for snapshots
and combines both queries on the datastore.

The example usages are updated in LocalState.

DomainGetLatestQuery uses BloomFilter existence index. Previously BloomFilter only accepted uint64_t as a lookup key. Now it is able to lookup using any ByteView key much like RecSplit/AccessorIndex. Hashing this key requires having a salt from "salt-state.txt". This file is read by SnapshotRepository::reopen_folder() and passed along to the BloomFilter constructor. If a SnapshotRepository is externally managed (via CAPI), the index_salt is expected to be passed on construction via CAPI.

Additional changes:
* disable domain accessor index by default - it is unused in Erigon 3 ATM
* move murmur_hash3 to snapshots/common - to reuse in BloomFilterKeyHasher
* "Snapshots" is added into snapshots codec names to distinguish from KVDB codecs, necessary codecs added

### DIFF
--- a/silkworm/db/blocks/schema_config.cpp
+++ b/silkworm/db/blocks/schema_config.cpp
@@ -22,6 +22,7 @@ namespace silkworm::db::blocks {
 
 snapshots::Schema::RepositoryDef make_blocks_repository_schema() {
     snapshots::Schema::RepositoryDef repository_schema;
+    repository_schema.index_salt_file_name("salt-blocks.txt");
     snapshots::Schema::EntityDef& schema = repository_schema.default_entity();
 
     schema.segment(kHeaderSegmentName)
@@ -55,12 +56,16 @@ std::unique_ptr<snapshots::IndexBuildersFactory> make_blocks_index_builders_fact
     return std::make_unique<BlocksIndexBuildersFactory>(make_blocks_repository_schema());
 }
 
-snapshots::SnapshotRepository make_blocks_repository(std::filesystem::path dir_path, bool open) {
+snapshots::SnapshotRepository make_blocks_repository(
+    std::filesystem::path dir_path,
+    bool open,
+    std::optional<uint32_t> index_salt) {
     return snapshots::SnapshotRepository{
         std::move(dir_path),
         open,
         make_blocks_repository_schema(),
         std::make_unique<datastore::StepToBlockNumConverter>(),
+        std::move(index_salt),
         make_blocks_index_builders_factory(),
     };
 }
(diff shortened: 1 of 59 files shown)


silkworm/db/cli/snapshots.cpp
```diff
@@ -544,19 +544,14 @@ void open_existence_index(const SnapshotSubcommandSettings& settings) {
                 ByteView nonexistent_key = {full_be + kSizeDiff, sizeof(intx::uint256) - kSizeDiff};
                 SILK_TRACE << "KV: previous_key=" << to_hex(previous_key) << " key=" << to_hex(key)
                            << " nonexistent_key=" << to_hex(nonexistent_key);
-                // Hash the nonexistent key using murmur3 and check its presence in existence filter
-                rec_split::Hash128 key_hash{};
-                snapshots::rec_split::murmur_hash3_x64_128(nonexistent_key.data(), nonexistent_key.size(), salt, &key_hash);
-                if (const bool key_found = existence_index.contains_hash(key_hash.first); key_found) {
+                if (const bool key_found = existence_index.contains(nonexistent_key); key_found) {
```
JacekGlen, line 547:
```suggestion
                if (existence_index.contains(nonexistent_key)) {
```

silkworm/execution/local_state.cpp
```diff
@@ -27,7 +28,12 @@ using namespace db::state;
 using namespace datastore;
 
 std::optional<Account> LocalState::read_account(const evmc::address& address) const noexcept {
-    AccountsDomainGetLatestQuery query{tx_, data_store_.state_db().accounts_domain()};
+    AccountsDomainGetLatestQuery query{
+        db::state::kDomainNameAccounts,
+        data_store_.chaindata,
+        tx_,
+        data_store_.state_repository,
+    };
```
JacekGlen, line 36: AccountsDomainGetLatestQuery only deals with the `db::state::kDomainNameAccounts` domain, so ideally the declaration would be:

```suggestion
    AccountsDomainGetLatestQuery query{
        data_store_.chaindata,
        tx_,
        data_store_.state_repository,
    };
```

silkworm/db/blocks/schema_config.cpp
```diff
@@ -22,6 +22,7 @@ namespace silkworm::db::blocks {
 
 snapshots::Schema::RepositoryDef make_blocks_repository_schema() {
     snapshots::Schema::RepositoryDef repository_schema;
+    repository_schema.index_salt_file_name("salt-blocks.txt");
```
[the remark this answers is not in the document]
battlmonstr: Yes, this is the default name. There are 2 salt files, one for the block repo and one for state repo. The defaults are in schema_config. For CAPI it should be passed from erigon to follow the pattern of not opening files on our end, so the file name is not relevant.

silkworm/db/cli/snapshots.cpp
battlmonstr: :+1:

silkworm/execution/local_state.cpp
battlmonstr: good idea, I'll try to hide it

PR #2668, opened 2025-01-21
evm: refactor bailout option handling

chfast:
Implement the bailout option by pure state modification before the transaction validation and execution.

### DIFF
--- a/silkworm/core/execution/processor.cpp
+++ b/silkworm/core/execution/processor.cpp
@@ -264,10 +264,6 @@ CallResult ExecutionProcessor::call(const Transaction& txn, const std::vector<st
 
     SILKWORM_ASSERT(protocol::validate_call_precheck(txn, evm_) == ValidationResult::kOk);
 
-    if (!evm().bailout) {
-        SILKWORM_ASSERT(protocol::validate_call_funds(txn, evm_, state_.get_balance(*txn.sender())) == ValidationResult::kOk);
-    }
-
     const BlockHeader& header{evm_.block().header};
     const intx::uint256 base_fee_per_gas{header.base_fee_per_gas.value_or(0)};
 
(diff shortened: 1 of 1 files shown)


JacekGlen (approved):

silkworm/core/protocol/validation.cpp
```diff
@@ -258,6 +254,17 @@ intx::uint256 compute_call_cost(const Transaction& txn, const intx::uint256& eff
     return required_funds;
 }
 
+intx::uint256 compute_call_required_funds(const Transaction& txn, const intx::uint256& effective_gas_price, const EVM& evm) {
```
chfast, line 257: I think we at least need good documentation what this function is for and why it is different from `compute_call_cost()`.

silkworm/core/execution/processor.cpp
```diff
@@ -284,10 +280,18 @@ CallResult ExecutionProcessor::call(const Transaction& txn, const std::vector<st
...
+        required_funds = protocol::compute_call_required_funds(txn, effective_gas_price, evm_);
+        // If the bailout option is on, add the required funds to the sender's balance
+        // so that after the transaction costs are deducted, the sender's balance is unchanged.
+        state_.add_to_balance(*txn.sender(), required_funds);
+    } else {
+        required_funds = protocol::compute_call_cost(txn, effective_gas_price, evm_);
     }
+
+    SILKWORM_ASSERT(protocol::validate_call_funds(txn, evm_, state_.get_balance(*txn.sender())) == ValidationResult::kOk);
+    state_.subtract_from_balance(*txn.sender(), required_funds);
```
chfast, line 294: I still don't understand why we are subtracting different values depending on the bailout on/off. This still feels quite a hack.

canepat: Yep, I hear you about the feeling. I guess the reason is `standard_transfer` function [here](https://github.com/erigontech/silkworm/blob/ci/bailout_refactor/silkworm/core/execution/evm.hpp#L83) does apply the bailout on each value transfer included the `txn.value` from `sender`. @lupin012 can you comment on this?

lupin012: Now, Yes in standard_transfer() called in EVM::create() EVM::Call() in case of NOT bailout the value is subtracked. Each txn (also internal) must subtracks its ether transfer

silkworm/core/protocol/validation.cpp
canepat: Function removed by @lupin012 in last refactoring

silkworm/core/execution/processor.cpp
canepat: > I still don't understand why we are subtracting different values depending on the bailout on/off. This still feels quite a hack.

Also this point has been superseded by the last refactoring.

canepat (approved):

PR #2687, opened 2025-01-31
evm: enable evmone APIv2 in block executor

chfast:
This include a evmone change where we make sure the tracer is invoked even for empty code.


JacekGlen (approved):

canepat:
I've converted this back to draft waiting for evmone to fix #2690

PR #2689, opened 2025-01-31
db: fix Windows build after PR 2686

canepat:


battlmonstr:
What was the problem here? Some of the code existed from before.

PR #2709, opened 2025-02-10
snapshots: improve SnapshotRepository logs

canepat:
Depends on #2708

### DIFF
--- a/silkworm/db/datastore/snapshots/snapshot_repository.cpp
+++ b/silkworm/db/datastore/snapshots/snapshot_repository.cpp
@@ -20,8 +20,6 @@
 #include <iterator>
 #include <utility>
 
-#include <magic_enum.hpp>
-
 #include <silkworm/infra/common/ensure.hpp>
 #include <silkworm/infra/common/log.hpp>
 
@@ -34,14 +32,14 @@ namespace fs = std::filesystem;
 using namespace datastore;
 
 SnapshotRepository::SnapshotRepository(
-    RepositoryKind kind,
+    datastore::EntityName name,
     std::filesystem::path dir_path,
     bool open,
     Schema::RepositoryDef schema,
     std::unique_ptr<StepToTimestampConverter> step_converter,
     std::optional<uint32_t> index_salt,
     std::unique_ptr<IndexBuildersFactory> index_builders_factory)
-    : kind_(kind),
+    : name_(std::move(name)),
       dir_path_(std::move(dir_path)),
       schema_(std::move(schema)),
       step_converter_(std::move(step_converter)),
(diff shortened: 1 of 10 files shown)


silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -32,13 +34,15 @@ namespace fs = std::filesystem;
 using namespace datastore;
 
 SnapshotRepository::SnapshotRepository(
+    RepositoryKind kind,
```
battlmonstr, line 37: let's pass an EntityName here
there are constants for each repo type: kBlocksRepositoryName, kStateRepositoryNameLatest, kStateRepositoryNameHistorical

silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -157,8 +161,9 @@ void SnapshotRepository::reopen_folder() {
     bundles_ = bundles;
     lock.unlock();
 
-    SILK_INFO << "Total reopened bundles: " << bundles_count()
-              << " max block available: " << max_block_available();
+    SILK_INFO << "Total reopened " << magic_enum::enum_name(kind_) << " bundles: " << bundles_count()
+              << (kind_ == RepositoryKind::blocks ? " max block available: " : " max timestamp available: ")
+              << (kind_ == RepositoryKind::blocks ? max_block_available() : max_timestamp_available());
```
battlmonstr, line 166: both blocks and txns are timestamps in our terminology
max_block_available and max_timestamp_available return exactly the same value
max_block_available() is a legacy method and shouldn't be used within the datastore, I ought to refactor it, but didn't fine time

I'd keep it simple for now and just print `max timestamp` here (a log reader will understand what the unit is based on the repo name)

if we want to make it prettier and identify the unit name (blocks or txns), I'd add a method to the StepToTimestampConverter interface

silkworm/db/kv/api/local_transaction.cpp
```diff
@@ -66,11 +68,16 @@ Task<std::shared_ptr<CursorDupSort>> LocalTransaction::get_cursor(const std::str
...
-    return std::make_shared<chain::LocalChainStorage>(DataModel{txn_, data_store_.blocks_repository});
+    return std::make_shared<chain::LocalChainStorage>(DataModel{tx_, data_store_.blocks_repository});
 }
 
-Task<TxnId> LocalTransaction::first_txn_num_in_block(BlockNum /*block_num*/) {
-    throw std::logic_error{"LocalTransaction::first_txn_num_in_block: not yet implemented"};
+Task<TxnId> LocalTransaction::first_txn_num_in_block(BlockNum block_num) {
+    auto canonical_body_for_storage = [this](BlockNum bn) -> Task<std::optional<Bytes>> {
+        DataModel access_layer{tx_, data_store_.blocks_repository};
+        co_return access_layer.read_raw_body_for_storage_from_snapshot(bn);
```
battlmonstr, line 77: there's this extra functionality in this PR not related to logs, did you intend to update the PR description or split a new PR?

battlmonstr (changes requested):

silkworm/db/kv/api/local_transaction.cpp
canepat: It was in a separate PR #2713 based on this one, then @JacekGlen merged the former into the latter

PR #2715, opened 2025-02-11
datastore: fix state storage format

JacekGlen:
When writing data for the domain storage, the leading zeros from the value are truncated.

Similarly, when reading data from kvdb or snapshots, the storage value is refilled with zeros to fit the length of bytes32.

### DIFF
--- a/silkworm/db/state/storage_codecs.hpp
+++ b/silkworm/db/state/storage_codecs.hpp
@@ -23,6 +23,27 @@
 
 namespace silkworm::db::state {
 
+struct Bytes32NoLeadingZerosCodec : public datastore::kvdb::Codec {
+    evmc::bytes32 value;
+    ~Bytes32NoLeadingZerosCodec() override = default;
+
+    datastore::kvdb::Slice encode() override {
+        // find first non-zero byte
+        u_int64_t offset = 0;
+        while (value.bytes[offset] == 0 && offset < sizeof(value.bytes)) {
+            offset++;
+        }
+        return {value.bytes + offset, sizeof(value.bytes) - offset};
+    }
+
+    void decode(datastore::kvdb::Slice slice) override {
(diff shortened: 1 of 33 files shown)


silkworm/db/state/storage_codecs.hpp
```diff
@@ -34,26 +34,69 @@ struct Bytes32KVDBCodec : public datastore::kvdb::Codec {
...
+            offset++;
+        }
+        return {value.bytes + offset, sizeof(value.bytes) - offset};
+    }
+
+    void decode(datastore::kvdb::Slice slice) override {
+        SILKWORM_ASSERT(slice.size() <= sizeof(value.bytes));
+        u_int64_t offset = sizeof(value.bytes) - slice.size();
+        std::memset(value.bytes, 0, offset);
+        std::memcpy(value.bytes + offset, slice.data(), slice.size());
```
canepat, line 57: I suggest to use `to_bytes32` from `core/types/evmc_bytes32.hpp` to avoid code duplication

silkworm/db/state/storage_codecs.hpp
```diff
@@ -34,26 +34,69 @@ struct Bytes32KVDBCodec : public datastore::kvdb::Codec {
...
+            offset++;
+        }
+        return ByteView{value.bytes + offset, sizeof(value.bytes) - offset};
+    }
+    void decode_word(ByteView word) override {
+        if (word.size() > sizeof(value.bytes))
+            throw std::runtime_error{"Bytes32ReducedSnapshotsCodec failed to decode"};
+        u_int64_t offset = sizeof(value.bytes) - word.size();
+        std::memset(value.bytes, 0, offset);
+        std::memcpy(value.bytes + offset, word.data(), word.size());
```
canepat, line 94: Same as above

canepat (approved):

PR #2726, opened 2025-02-13
db: speedup SnapshotRepository reopening

canepat:


silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -128,37 +140,46 @@ void SnapshotRepository::reopen_folder() {
...
+
+#ifndef NDEBUG
+    // Sanity check: no gap must exist in bundles
+    if (!bundles_->empty()) {
+        Step max_end = bundles_->cbegin()->second->step_range().start;
+        for (const auto& b : *bundles_) {
+            SILKWORM_ASSERT(b.second->step_range().start == max_end);
+            max_end = b.second->step_range().end;
         }
     }
```
battlmonstr, line 180: This is optional, but std::adjacent_find seems like a perfect algorithm here. It takes a predicate that accepts 2 arguments: entry and entry_next. Note that `entry.first` is always equal to `entry.second->step_range().start` in this map. The predicate can compare entry.second->step_range().end to entry_next.first. 

With C++ 23 [pairwise view](https://en.cppreference.com/w/cpp/ranges/adjacent_view) this code would be even simpler.

silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -246,7 +267,7 @@ SnapshotPathList SnapshotRepository::get_files(std::string_view ext) const {
     return snapshot_files;
 }
 
-std::vector<StepRange> SnapshotRepository::list_dir_file_ranges() const {
+std::set<StepRange> SnapshotRepository::list_dir_file_ranges() const {
```
battlmonstr, line 270: I suggest to refactor it to something like:
```suggestion
void SnapshotRepository::list_files(std::function<void(SnapshotPath)> consumer) const {
```
It makes the function more reusable: we can use it whether we want if we want duplicates/sorting or not

This is optional though, and can be refactored later.

battlmonstr (changes requested):

silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -128,37 +140,46 @@ void SnapshotRepository::reopen_folder() {
...
-        if (!bundles->contains(num)) {
-            SnapshotBundlePaths bundle_paths{schema_, dir_path_, range};
-            // if all bundle paths exist
-            if (std::ranges::all_of(bundle_paths.files(), [](const fs::path& p) { return fs::exists(p); })) {
-                SnapshotBundle bundle = open_bundle(range);
-                bundles->insert_or_assign(num, std::make_shared<SnapshotBundle>(std::move(bundle)));
-                // avoid gaps/overlaps
-                num = range.end;
+        SnapshotBundlePaths bundle_paths{schema_, dir_path_, range};
+        // Open iff all bundle paths exist
```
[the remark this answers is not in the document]
canepat: yep, it's meant to be "if and only if"

silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -128,37 +140,46 @@ void SnapshotRepository::reopen_folder() {
...
+                std::shared_ptr<SnapshotBundle> bundle_ptr = future_bundle_ptr.get();
+                const auto step_range = bundle_ptr->step_range();
+                bundles_->insert_or_assign(step_range.start, std::move(bundle_ptr));
+                return true;
             }
+            return false;
+        });
+    }
+
+#ifndef NDEBUG
```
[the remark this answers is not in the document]
canepat: Only `NDEBUG` macro is C++ standard, so even if unfortunately we're forced to double logic here there is not cleaner and standard way to obtain the same effect. Specifically, there is no such thing as `DEBUG` macro: MSVC has `_DEBUG`, but again not standard.

silkworm/db/datastore/snapshots/snapshot_repository.cpp
```diff
@@ -128,37 +140,46 @@ void SnapshotRepository::reopen_folder() {
...
-            if (std::ranges::all_of(bundle_paths.files(), [](const fs::path& p) { return fs::exists(p); })) {
-                SnapshotBundle bundle = open_bundle(range);
-                bundles->insert_or_assign(num, std::make_shared<SnapshotBundle>(std::move(bundle)));
-                // avoid gaps/overlaps
-                num = range.end;
+        SnapshotBundlePaths bundle_paths{schema_, dir_path_, range};
+        // Open iff all bundle paths exist
+        if (std::ranges::all_of(bundle_paths.files(), [](const fs::path& p) { return fs::exists(p); })) {
+            // Schedule each bundle opening on worker pool collecting its future result for completion handling
+            completed_future_bundles.emplace_back(worker_pool.submit([&, range]() -> std::shared_ptr<SnapshotBundle> {
```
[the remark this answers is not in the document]
canepat: I prefer to do this once the logic of missing_bundle_paths becomes a bit more complicated, it feels unnecessary right.

silkworm/db/datastore/snapshots/snapshot_repository.cpp
canepat: I'll leave it for later

silkworm/db/datastore/snapshots/snapshot_repository.cpp
canepat: I'll leave it for later

battlmonstr (approved):

PR #2740, opened 2025-02-19
rpc: get EVM opcode names from evmone

chfast:
Use evmone's instruction traits instead of EVMC to get the opcode names.

### DIFF
--- a/silkworm/rpc/common/util.cpp
+++ b/silkworm/rpc/common/util.cpp
@@ -16,6 +16,8 @@
 
 #include "util.hpp"
 
+#include <evmone/instructions_traits.hpp>
+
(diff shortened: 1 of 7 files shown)


Sixtysixter:
My suggestion is to also keep the old version of `get_opcode_name` for use in the `trace_*` APIs and to migrate only the `debug_*` APIs to the new version. This will help maintain compatibility with **ERIGON** regarding **undefined opcodes**.  

For the opcode name **0x44 (DIFFICULTY/PREVRANDAO)**, in order to maintain compatibility and while waiting for the name to be updated in **ERIGON**, a name change should be added in the new version of `get_opcode_name`.

canepat (approved):

PR #2750, opened 2025-02-24
db: Data API timestamp conversion

canepat:


lupin012 (approved):

silkworm/db/kv/api/local_timestamp.cpp
battlmonstr, line 1: I'd either call the file ts_range_from_kv.cpp, or move it to api/endpoint/timestamp.cpp and move api::Timestamp and TimestampRange from common.hpp there.

silkworm/db/kv/api/local_timestamp.cpp
```diff
@@ -0,0 +1,37 @@
...
+   limitations under the License.
+*/
+
+#include "local_timestamp.hpp"
+
+#include <silkworm/infra/common/ensure.hpp>
+
+namespace silkworm::db::kv::api {
+
+datastore::TimestampRange ts_range_from_kv(Timestamp from_ts, Timestamp to_ts, bool reverse) {
```
battlmonstr, line 23:
```suggestion
datastore::TimestampRange ts_range_from_kv(TimestampRange ts_range, bool reverse) {
```
TimestampRange is defined in endpoint/common.hpp

battlmonstr (approved):

silkworm/db/kv/api/local_timestamp.cpp
canepat: I'd like to keep it as is at least for now, because it seems more coherent with local_cursor.cpp and local_transaction.cpp.
I'm not against some refactoring to give better names (if any) for all such abstractions, though.

canepat: I'll take your first suggestion (renaming the file) and in the meantime improve the naming from `ts_range_from_kv` to `as_datastore_ts_range`, since the function is already in namespace `db::kv::api`.

PR #2760, opened 2025-02-28
db: improve performance of Decompressor::Iterator

canepat:


silkworm/db/datastore/domain_range_latest_query.hpp
```diff
@@ -58,6 +58,11 @@ struct DomainRangeLatestQuery {
     static_assert(std::same_as<Key1, Key2>);
     using Key = Key1;
 
+    using Word1 = typename TKeyDecoder2::Word;
+    using Word2 = typename TValueDecoder2::Word;
+    static_assert(std::same_as<Word1, Word2>);
+    using Word = Word1;
+
```
battlmonstr, line 65:
```suggestion
    using Word = snapshots::Decoder::Word;
```
Since snapshots::Decoder::Word = BytesOrByteView in the snapshots::Decoder interface, and TKeyDecoder2 / TValueDecoder2 implement a snapshots::DecoderConcept - they need to inherit Decoder interface, so TKeyDecoder2::Word is always BytesOrByteView. I suggest to keep it simple and use snapshots::Decoder::Word directly.

silkworm/db/datastore/snapshots/elias_fano/elias_fano_decoder.hpp
```diff
@@ -26,9 +26,15 @@ struct EliasFanoDecoder : public snapshots::Decoder {
...
     ~EliasFanoDecoder() override = default;
 
-    void decode_word(Bytes& word) override {
-        value = EliasFanoList32::from_encoded_data(word);
-        value.data_holder = std::move(word);
+    void decode_word(Word& word) override {
+        if (word.holds_bytes()) {
+            auto word_value = std::get<Bytes>(word);
+            value = EliasFanoList32::from_encoded_data(word_value);
+            value.data_holder = std::move(word_value);
```
battlmonstr, line 33: it would be nice to encapsulate this in EliasFanoList32. how about adding `from_encoded_data(Bytes)` that does exactly these 2 lines?

silkworm/db/datastore/snapshots/elias_fano/elias_fano_decoder.hpp
```diff
@@ -26,9 +26,15 @@ struct EliasFanoDecoder : public snapshots::Decoder {
 
     ~EliasFanoDecoder() override = default;
 
-    void decode_word(Bytes& word) override {
-        value = EliasFanoList32::from_encoded_data(word);
-        value.data_holder = std::move(word);
+    void decode_word(Word& word) override {
+        if (word.holds_bytes()) {
+            auto word_value = std::get<Bytes>(word);
```
battlmonstr, line 31:
```suggestion
            Bytes& word_value = std::get<Bytes>(word);
```

silkworm/db/datastore/snapshots/domain_range_latest_query.hpp
```diff
@@ -77,18 +77,24 @@ struct DomainRangeLatestQuery {
     using ResultItemKey = decltype(TKeyDecoder::value);
     using ResultItemValue = decltype(TValueDecoder::value);
     using ResultItem = std::pair<ResultItemKey, ResultItemValue>;
+    using Word1 = typename TKeyDecoder::Word;
+    using Word2 = typename TValueDecoder::Word;
+    static_assert(std::same_as<Word1, Word2>);
+    using Word = Word1;
```
battlmonstr, line 83:
```suggestion
    using Word = snapshots::Decoder::Word;
```

silkworm/db/datastore/snapshots/history_range_query.hpp
```diff
@@ -38,6 +38,7 @@ struct HistoryRangeSegmentQuery {
     using Key = decltype(TKeyDecoder::value);
     using Value = decltype(TValueDecoder::value);
     using ResultItem = std::pair<Key, Value>;
+    using Word = typename TKeyDecoder::Word;
```
battlmonstr, line 41:
```suggestion
    using Word = snapshots::Decoder::Word;
```

silkworm/db/datastore/snapshots/segment/kv_segment_test.cpp
```diff
@@ -36,8 +36,12 @@ struct CharCodec : public Codec {
         word.push_back(*byte_ptr_cast(&value));
         return word;
     }
-    void decode_word(Bytes& input_word) override {
-        value = *byte_ptr_cast(input_word.data());
+    void decode_word(Word& input_word) override {
+        auto input_word_view = input_word.byte_view();
```
battlmonstr, line 40:
```suggestion
        ByteView input_word_view = input_word.byte_view();
```
I prefer to only use `auto` for long and complex types. With an explicit type the code is easier to understand (given that the type name itself is readable).

silkworm/db/datastore/snapshots/segment/seg/decompressor.cpp
```diff
@@ -862,14 +862,15 @@ bool Decompressor::Iterator::is_next_word_compressed() const {
 Decompressor::Iterator& Decompressor::Iterator::operator++() {
     if (has_next()) {
         current_word_offset_ = word_offset_;
-        current_word_.clear();
 
         bool is_next_word_compressed = this->is_next_word_compressed();
         is_next_value_ = !is_next_value_;
         if (is_next_word_compressed) {
-            next_compressed(current_word_);
+            current_word_ = Bytes{};
```
battlmonstr, line 869:
```suggestion
            std::get<Bytes>(current_word_).clear();
```
The idea taken from the erigon code was to not reallocate current_word_ unless it grows bigger. For files that have word size limited, this can result in a better perf.

Doing .clear() instead of reassign should work for .seg and .v files, and even .kv files if compression mode = both.

If we want to keep this optimization for .kv files with only keys or values compressed, an option could be to use a custom `union { Bytes, ByteView, flag }` instead of `std::variant`, and not touch Bytes between iterations.

silkworm/db/state/account_codecs.hpp
```diff
@@ -69,9 +69,10 @@ struct AccountSnapshotsCodec : public snapshots::Codec {
         return word;
     }
 
-    void decode_word(Bytes& input_word) override {
-        if (!input_word.empty()) {
-            value = unwrap_or_throw(AccountCodec::from_encoded_storage_v3(input_word),
+    void decode_word(Word& input_word) override {
+        const auto input_word_view = input_word.byte_view();
```
battlmonstr, line 73:
```suggestion
        const ByteView input_word_view = input_word.byte_view();
```

silkworm/db/state/account_codecs_test.cpp
```diff
@@ -84,28 +84,28 @@ TEST_CASE("AccountSnapshotsCodec") {
         CHECK(codec.encode_word() == *from_hex("010201030000"));
     }
     SECTION("decode") {
-        auto word1 = *from_hex("");
+        auto word1 = BytesOrByteView{*from_hex("")};
```
battlmonstr, line 87:
```suggestion
        Word word1{*from_hex("")};
```
(same for word*N* below)

silkworm/db/state/address_codecs.hpp
```diff
@@ -45,10 +45,11 @@ struct AddressSnapshotsCodec : public snapshots::Codec {
         return ByteView{reinterpret_cast<uint8_t*>(&value.bytes), kAddressLength};
     }
 
-    void decode_word(Bytes& word) override {
-        if (word.size() < kAddressLength)
+    void decode_word(Word& word) override {
+        const auto word_view = word.byte_view();
```
battlmonstr, line 49:
```suggestion
        const ByteView word_view = word.byte_view();
```

silkworm/db/state/hash_decoder.hpp
```diff
@@ -28,10 +28,11 @@ struct HashSnapshotsDecoder : public snapshots::Decoder {
 
     ~HashSnapshotsDecoder() override = default;
 
-    void decode_word(Bytes& word) override {
-        if (word.size() < kHashLength)
+    void decode_word(Word& word) override {
+        auto word_view = word.byte_view();
```
battlmonstr, line 32:
```suggestion
        ByteView word_view = word.byte_view();
```

silkworm/db/state/receipts_domain.hpp
```diff
@@ -38,10 +38,11 @@ enum class ReceiptsDomainKey : uint8_t {
 struct ReceiptsDomainKeySnapshotsDecoder : public snapshots::Decoder {
     ReceiptsDomainKey value{};
     ~ReceiptsDomainKeySnapshotsDecoder() override = default;
-    void decode_word(Bytes& word) override {
-        if (word.empty())
+    void decode_word(Word& word) override {
+        auto word_view = word.byte_view();
```
battlmonstr, line 42:
```suggestion
        ByteView word_view = word.byte_view();
```

battlmonstr (approved):

silkworm/core/common/bytes.hpp
```diff
@@ -59,4 +60,20 @@ class ByteView : public evmc::bytes_view {
...
+
+    bool holds_bytes() const { return std::holds_alternative<Bytes>(*this); }
+
+    bool holds_byte_view() const { return std::holds_alternative<ByteView>(*this); }
+
+    BytesOrByteView substr(size_t offset) {
+        return holds_bytes() ? BytesOrByteView{std::get<Bytes>(*this).substr(offset)} : BytesOrByteView{std::get<ByteView>(*this).substr(offset)};
+    }
+
+    ByteView byte_view() const {
```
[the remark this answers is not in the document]
canepat: The interface choice here is careful: this method is explicit because I really want the calling code to show that we're _potentially_ taking a view of an underlying data hold by the variant container. The `BytesOrByteView` instance must survive the byte view lifecycle if you want to be safe in all circumstances.

Using an implicit cast would somewhat defeat this purpose, but I'm open to do the change you ask if you still think it's well suited.

silkworm/db/datastore/snapshots/elias_fano/elias_fano_decoder.hpp
canepat: I've refactored this to avoid using l-values altogether

silkworm/db/datastore/snapshots/segment/seg/decompressor.cpp
canepat: We need to check whether `current_word_` contains `Bytes` or `ByteView`, so it must be:
```
if (current_word_.holds_bytes()) {
    std::get<Bytes>(current_word_).clear();
} else {
    current_word_ = Bytes{};
}
```
I'd like to measure this change, so I'm gonna leave it for a follow-up PR.

PR #2768, opened 2025-03-04
db: rename kv Query to Request

battlmonstr:


cmd/dev/grpc_toolbox.cpp
```diff
@@ -1050,14 +1050,14 @@ int kv_get_latest() {
     const auto verbose{absl::GetFlag(FLAGS_verbose)};
 
     if (timestamp > -1) {
-        GetAsOfQuery query{
+        GetAsOfRequest query{
```
JacekGlen, line 1053:
```suggestion
        GetAsOfRequest request{
```

cmd/dev/grpc_toolbox.cpp
```diff
@@ -1050,14 +1050,14 @@ int kv_get_latest() {
...
             .table = table_name,
             .key = *key_bytes,
             .timestamp = timestamp,
         };
         return execute_temporal_kv_query(target, kv_get_as_of_query, std::move(query), verbose);
     }
-    GetLatestQuery query{
+    GetLatestRequest query{
```
JacekGlen, line 1060:
```suggestion
    GetLatestRequest request{
```

JacekGlen (approved):
That looks good.

I just have one thought: can we apply similar logic to silkworm::datastore::kvdb:: and silkworm::db::state::?

Or maybe use a different approach and apply the following naming convention:
- DomainGetLatestQuery -> DomainGetLatestQuery
- DomainPutQuery -> DomainPutCommand
- DomainDeleteQuery -> DomainDeleteCommand

I know it's a minor thing but things like `DomainPutQuery` might be confusing

battlmonstr:
@JacekGlen thanks for the review, let's discuss it further live. I'll address the found leftovers in a follow up PR.

PR #2788, opened 2025-03-11
evm: refactor "norefund" option handling

chfast:
Because `refund_gas()` method returns gas left, use this value directly and compute `gas_used` independently of the "norefund" option.


JacekGlen (approved):

chfast:
Next step is to try to move this logic outside of this function... I will try to handle this later.

PR #2797, opened 2025-03-18
docs: code style updates

battlmonstr:
Updates with decisions discussed and voted at the workshop meeting.

Andrew confirmed P25 changing to the SPDX short form Apache boilerplate (as in Linux kernel and evmone).

### DIFF
--- a/docs/code_style.md
+++ b/docs/code_style.md
@@ -255,6 +255,153 @@ Sometimes we use `std::unique_ptr/smart_ptr` to delay initialization or for opti
 
 Use a default-constructed `std::unique_ptr/smart_ptr` where null is expected.
 
+### P19 name prefix for factory functions
+
+Use `make` name/prefix for factory functions instead of `build` or `create`:
+
+	SentryPtrPair make_sentry(...)
+	std::shared_ptr<TimerImpl> make(...)
+	std::string make_thread_name(...)
+
+`build` prefix can be used for a “builder” pattern, where it might have no parameters, and potential side effects, for example `RecSplit::build()`.
+
+### P20 to_string method naming
+
+Use `to_string()` by default to convert an object to std::string (e.g. for logging purposes). Only provide `operator<<` if needed (implemented on top of `to_string()`).
+
+If a global function is required, `xxx_to_string()` name should be used.
+
+### P21 multiline parameters indentation
+
+Default to 1 parameter per line.
+
+Accepted styles:
+
+	ValidationResult execute_block(
+		const Block& block,
+		State& state,
+		const ChainConfig& chain_config);
+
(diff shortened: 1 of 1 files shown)


docs/code_style.md
```diff
@@ -255,6 +255,153 @@ Sometimes we use `std::unique_ptr/smart_ptr` to delay initialization or for opti
...
+
+	friend bool operator==(const Config& lhs, const Config& rhs) = default;
+
+Bad:
+
+	bool operator==(const Config& other) const = default;
+
+### P30 tx vs txn naming
+
+Use "txn", "txns" abbreviations for Ethereum transactions.
```
chfast, line 394: Just a comment, not complain: evmone uses `tx` for Ethereum transactions because it don't have db :). I think `tx` is uniformly used in all EVM work.

docs/code_style.md
```diff
@@ -255,6 +255,153 @@ Sometimes we use `std::unique_ptr/smart_ptr` to delay initialization or for opti
...
+
+If a global function is required, `xxx_to_string()` name should be used.
+
+### P21 multiline parameters indentation
+
+Default to 1 parameter per line.
+
+Accepted styles:
+
+	ValidationResult execute_block(
```
chfast, line 280: Does clang-format is happy with both variants?

I think you can configure clang-format to have desired formatting but this likely requires putting the line length limit.

chfast (approved):
Very good, thanks for working on this.

docs/code_style.md
battlmonstr: Yes, this is the status quo.
I wanted to be stricter here, but we ended up fixating the status quo, because nobody was sure how to reconfigure the clang-format. And in particular, people wanted to have a clear separation of the function signature from the body. E.g. in Go/Swift/Rust style it is possible to have the final `) {` on a separate line, and even have comma after the last argument.

docs/code_style.md
```diff
@@ -255,6 +255,153 @@ Sometimes we use `std::unique_ptr/smart_ptr` to delay initialization or for opti
...
+
+	TemporaryDirectory tmp_dir;
+
+Bad:
+
+	TempChainData temp_chaindata;
+
+### P25 license boilerplate
+
+Use SPDX short form like in evmone:
```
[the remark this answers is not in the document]
battlmonstr: yes, Andrew confirmed

see also:
https://chatgpt.com/share/67d9410a-7894-8001-9349-2d3a7d127e31
https://github.com/torvalds/linux/blob/76b6905c11fd3c6dc4562aefc3e8c4429fefae1e/rust/kernel/init.rs#L4

JacekGlen (approved):

PR #2810, opened 2025-03-22
snapshots: optimize access pattern in BTree index

canepat:


battlmonstr (changes requested):
I understand the nature of optimizations, but I'd like to discuss the design of this.

battlmonstr (approved):

canepat:
> I understand the nature of optimizations, but I'd like to discuss the design of this.

We discussed the changes and agreed that a slightly better solution in terms of internal design was possible, see fe8639107b5a1721a108e2b45ba1abcff4449732.

PR #2821, opened 2025-03-28
Adding block_start/block_end entrypoints, fixing saving receipts

bzawisto:
This pull request resolves an issue with saving receipts by ensuring that the correct `gas`, `blob_gas`, and `logIndex` values are used. These values must be calculated individually for each receipt, taking into account the receipts generated previously in the same block. To facilitate this, I introduced `block_start` and `block_end` entry points in Silkworm that Erigon must utilize.
For now, to keep the implementation straightforward, receipts remain in the Silkworm instance, though this will need revisiting in the future. Furthermore, computing the `receipts_root_hash` requires that all receipts be accessible within Erigon, so the idea of using a temporary MDBX store will be explored later.

### DIFF
--- a/silkworm/core/state/state.hpp
+++ b/silkworm/core/state/state.hpp
@@ -62,9 +62,12 @@ class State : public BlockState {
 
     virtual void decanonize_block(BlockNum block_num) = 0;
 
-    virtual void insert_receipts(BlockNum block_num, const std::vector<Receipt>& receipts) = 0;
+    virtual void insert_receipts([[maybe_unused]] BlockNum block_num, [[maybe_unused]] const std::vector<Receipt>& receipts){};
 
-    virtual void insert_receipt([[maybe_unused]] const Receipt& receipt, [[maybe_unused]] uint64_t blob_gas_used){};
+    // Added in Erigon3
+    virtual void insert_txn_receipts([[maybe_unused]] std::vector<Receipt>& receipts, [[maybe_unused]] const std::vector<std::pair<TxnId, Transaction>>& transactions){};
+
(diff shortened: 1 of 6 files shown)


silkworm/core/state/state.hpp
```diff
@@ -62,9 +62,12 @@ class State : public BlockState {
 
     virtual void decanonize_block(BlockNum block_num) = 0;
 
-    virtual void insert_receipts(BlockNum block_num, const std::vector<Receipt>& receipts) = 0;
+    virtual void insert_receipts([[maybe_unused]] BlockNum block_num, [[maybe_unused]] const std::vector<Receipt>& receipts){};
 
-    virtual void insert_receipt([[maybe_unused]] const Receipt& receipt, [[maybe_unused]] uint64_t blob_gas_used){};
+    // Added in Erigon3
+    virtual void insert_txn_receipts([[maybe_unused]] std::vector<Receipt>& receipts, [[maybe_unused]] const std::vector<std::pair<TxnId, Transaction>>& transactions){};
```
canepat, line 68: [[maybe_unused]] should be removed here

silkworm/core/state/state.hpp
```diff
@@ -62,9 +62,12 @@ class State : public BlockState {
...
+
+    virtual void insert_receipt([[maybe_unused]] const Receipt& receipt, [[maybe_unused]] uint64_t current_log_index, [[maybe_unused]] uint64_t blob_gas_used){};
```
canepat, line 70: Here as well, I suppose

canepat (approved):

PR #2823, opened 2025-03-30
rpcdaemon: improve eth_getTransactionReceipt using txn_id

lupin012:


canepat:
MacBookPro M1-Max 64GB RAM

_Before (master): 9fd99462_
% ./run_perf_tests.py -m 2 -p pattern/mainnet/stress_test_eth_getTransactionReceipt_14M.tar -r 5 -t 100:20 -y eth_getTransactionReceipt
Performance Test started
Test repetitions: 5 on sequence: 100:20 for pattern: pattern/mainnet/stress_test_eth_getTransactionReceipt_14M.tar
[1. 1] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=108.11ms]
[1. 2] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=56.649ms]
[1. 3] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=60.491ms]
[1. 4] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=71.865ms]
[1. 5] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=70.21ms]


_After (this branch): caf3e3d_
% ./run_perf_tests.py -m 2 -p pattern/mainnet/stress_test_eth_getTransactionReceipt_14M.tar -r 5 -t 100:20 -y eth_getTransactionReceipt
Performance Test started
Test repetitions: 5 on sequence: 100:20 for pattern: pattern/mainnet/stress_test_eth_getTransactionReceipt_14M.tar
[1. 1] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=121.946ms]
[1. 2] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=23.086ms]
[1. 3] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=24.05ms]
[1. 4] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=29.196ms]
[1. 5] daemon: executes test qps: 100 time: 20 -> [R=100.00% max=23.432ms]

silkworm/db/datastore/snapshots/rec_split/rec_split.hpp
```diff
@@ -706,10 +706,10 @@ class RecSplit {
         return lookup_by_ordinal(Ordinal{data_id - base_data_id()});
     }
 
-    std::optional<size_t> lookup_by_key(ByteView key) const {
+    std::optional<std::pair<size_t, uint64_t>> lookup_by_key(ByteView key) const {
```
battlmonstr, line 709: instead of modifying this method I suggest to call 2 other methods in a sequence from FindByHashSegmentQuery:
1. lookup_data_id_by_key - returns a data_id (Timestamp) by key
2. lookup_by_data_id - returns a value (offset in .seg) by data_id (Timestamp)

lookup_data_id_by_key needs to be exposed on the AccessorIndex (see accessor_index.hpp using declarations)

silkworm/db/datastore/snapshots/basic_queries.hpp
```diff
@@ -92,19 +93,27 @@ struct FindByHashSegmentQuery : public BasicSegmentQuery<TDecoder, segment_names
...
 
     std::optional<decltype(TDecoder::value)> exec(const Hash& hash) {
-        auto offset = this->index_.lookup_by_key(hash);
-        if (!offset) {
+        auto result = exec_with_timestamp(hash);
+        if (!result) return std::nullopt;
+        return result->first;
+    }
+
+    std::optional<std::pair<decltype(TDecoder::value), datastore::Timestamp>> exec_with_timestamp(const Hash& hash) {
```
battlmonstr, line 101:
```suggestion
    std::optional<Result> exec(const Hash& hash) {
```

I suggest to rename this method to `exec` (and remove original exec), because like that it will become accessible to combinators that execute queries over a range of bundles (such as FindMapQuery).
The convention is that each segment query has a single exec method (maybe we should switch to `operator()` to make it mandatory).

This query is used by HeaderFindByHashQuery which is used in a single method - DataModel::read_header_from_snapshot, so a little modification there will be needed.

Also I suggest to define an inner Result struct containing value and timestamp instead of using std::pair. This improves readability on the call sites. We do this in a few places, e.g. see DomainGetLatestQuery in snapshots.

silkworm/db/datastore/snapshots/rec_split/rec_split.hpp
battlmonstr: the methods I've mentioned (lookup_data_id_by_key/lookup_by_data_id) already do the job, and one is already exposed, no need to expose more or replicate anything, it will just work, trust me :)

silkworm/db/datastore/snapshots/basic_queries.hpp
canepat: I agree with the suggested change, i.e. `exec` instead of `exec_with_timestamp` but not with the switch to `operator()`, which in any case would not prevent having multiple methods through overloading.
The inner Result struct is really a plus for readability, thanks for the suggestion.

I'm gonna do this change into this PR.

